PR #122, opened 2013-03-11
AndroidLogChannel

yukoba:
I created AndroidLogChannel.
I'm glad this is added to the repository.


aleks-f:
@RangelReale : do you care to do this properly and test on Android for 1.7?

PR #278, opened 2013-08-30
Add example backtrace in exceptions

patrickjwhite:
A pull request only for you to see the code... 

I always find myself missing backtraces in c++ exceptions. I don't know if this is a totally terrible idea for some reason, but I tried this out and it works OK. The method it uses doesn't give line numbers, but you can at least find the function that the exception was thrown from.

I have only tested this on Linux, but if it makes sense could investigate other platforms.

Thoughts?


aleks-f:
@patrickjwhite : any plans to work on this?

PR #414, opened 2014-03-25
Mingw64

comio:
Hi all,

I'm just trying to compile under MinGW32-W64 compiler.

Here there are some patches. I think that 91c8f61 can be used to format strings without check the potinters size. This code can be used also in other places, removing the printf-like code.
4c87b5f must be also included because when we compile using mingw-make tool, the directory schema uses the Windows convention (C:/xxxxx) instead Cygwin (/c/....). $(CURDIR) is filled by make tool with the correct value.

ciao

luigi


aleks-f:
https://github.com/pocoproject/poco/commit/0076696d40057a76369c134234e4d4f3a5ea0204 merged, the rest passed on

Issue #532, opened 2014-08-29
FastMutex functionality differs on Windows & POSIX (and possibly other non-Windows OS)


smiley:
In [Mutex_POSIX.cpp](/pocoproject/poco/blob/b058143/Foundation/src/Mutex_POSIX.cpp#L78), `FastMutex` uses the `Mutex(bool fast)` constructor which creates a mutex that cannot be locked recursively.

However, in [Mutex_WIN32.h](/pocoproject/poco/blob/b058143/Foundation/include/Poco/Mutex_WIN32.h#L46) `FastMutexImpl` is essentially `MutexImpl`, [which uses a CriticalSection object](/pocoproject/poco/blob/b058143/Foundation/src/Mutex_WIN32.cpp#L28). On Windows, [CriticalSection objects can be locked recursively](http://msdn.microsoft.com/en-us/library/windows/desktop/ms682530.aspx).

This causes some frustration when you have code that works on Windows, but deadlocks on Linux for no clearly-visible reason. (E.g.: Class with synchronised methods, each one locking/unlocking the FastMutex, with one method using another)

PR #601, opened 2014-11-07
Issue #532 -  Changes to address `FastMutex` being non-recursive on Win32 platforms

martin-osborne:
Added c'tor to `Mutex` to specify if it should be recursive or non-recursive, this required:

* Minor changes to POSIX and VX implementations of `MutexImpl`
* Adding a boolean to Win32 and WinCE implementations of `MutexImpl` to indicate whether they should behave recursively or non-recursively, and a lock count for determining if the lock has gone recursive

Explicitly implemented Win32 `FastMutexImpl` as a simple wrapper around a Critical Section.

Updated `FastMutex` documentation to indicate it may be recursive or non-recursive.

Added tests for checking recursive behavior of `Mutex`.


aleks-f:
@obiltschnig is this good to be merged?

obiltschnig:
This will have to wait for 1.7 as it's a rather significant change.

Instead of a boolean argument, we should have an enum:

    enum MutexType
    {
        MUTEX_RECURSIVE,
        MUTEX_NONRECURSIVE
    }

and while at it, also deprecate the bool argument to the Event constructor and replace it with an enum as well.

martin-osborne:
Sorry for not responding sooner.  I will make a start updating the `Mutex` and `Event` c'tors to take an enum.

When the work is complete I may issue a new pull request so the history is clean and the changes easy to review.  But for now I would like to keep this open so activity can be tracked.

martin-osborne:
I've change the `Mutex` and `Event` c'tors to accept an enum, and removed uses of the now deprecated `Event::Event(bool)` c'tor.  The implementation is based on `FPEnvironment::RoundingMode`.

I've made sure all the code compiles on Win32 and POSIX, and the Foundation tests are passing.  (I've run other test suites as well.)

I've updated the WinCE and VX code but I have neither built nor tested those changes.  I've built the documentation and made sure that the docs look okay, especially that **Deprecated** is appearing on `Event::Event(bool)`.

I think the changes are ready for review and I look forward to your comments.

Issue #609, opened 2014-11-18
Allow a program option to hide in the usage help display


Janulka:
hidden options shall not be displayed to user on the help message invocation

This can be useful when the option name is refactored but the old usage is required to be preserved to prevent regression. However, only the new usage name is desired to be displayed to the user.

PR #610, opened 2014-11-18
Options: allow to set option as hidden

Janulka:
hidden options shall not be displayed to user on the help invocation

this can be useful when the option name is refactored but the old
usage is required to be preserved to prevent regression

Issue: Allow a program option to hide in the usage help display #609


aleks-f:
@Janulka : any plans to tidy up this pull?

PR #647, opened 2014-12-19
SNMP v1 and v2c Client implementation

RangelReale:
SNMP v1 and v2c client implementation, together with an ASN.1 encoder / decoder (in Foundation) needed by it.
Sample and test cases included, all code fully commented.

There is a raw version that returns the raw ASN.1 objects, and an user-friendly version.

Comments on the design of the classes are welcome.

Sample code:

```
		Poco::Net::SNMPClient snmpcli;

		snmpcli.snmpReply += Poco::FunctionDelegate<Poco::Net::SNMPEventArgs>(tpreply2);
		snmpcli.snmpError += Poco::FunctionDelegate<Poco::Net::SNMPEventArgs>(tperror2);

		snmpcli.get("172.18.1.164", "1.3.6.1.2.1.1.1.0");
```


aleks-f:
I have created [SNMP branch](https://github.com/pocoproject/poco/tree/SNMP) with separate ASN library for inclusion in 1.7.

@RangelReale : please check the ASN library for correctness and resubmit your pull into this branch. Note: please add and ifdef SNMP code in the Net with POCO_NET_NO_SNMP so Net can optionally be compiled without it.

You can submit projects for a VS version of your choice, I'll generate the rest once you have it in. Thanks again, it is a nice contribution.

RangelReale:
I was about to comment about your edit, Net (or NetEx) will have to depend on the ASN library if SNMP is enabled, but AFAIK there is no way to make a VC solution don't link a library depending on a define. (well there is "#import", but it is Windows only, don't know about other platforms).

aleks-f:
Optional exclusion would be only for the case we add it to Net; if you will do NetEx, then there is no need for exclusion. As for how to do it, on Windows, it isa matter of including [ASN.h](https://github.com/pocoproject/poco/blob/SNMP/ASN/include/Poco/ASN/ASN.h#L59). On POSIX, it can be ifdef'd in the Makefile. e.g.:

```c++
ifndef POCO_NET_NO_SNMP
        target_libs = PocoASN
endif

target_libs  +=  PocoFoundation CppUnit
```

RangelReale:
I submit a PR about the separate ASN library, now will see about making a NetEx library.
I should keep the include path / namespace as Poco::Net right?

aleks-f:
Yes

RangelReale:
I created a NetEx library with the ICMP and NTP components (still needs to integrate SNMP), if you want to review what was done to see if it is ok, this is the commit in my fork.

https://github.com/RangelReale/poco/commit/bfc7e3f6503da31c90b7ed303ffdbc73a0520a47

aleks-f:
Looks good to me.

aleks-f:
If you have time and interest to do it, there is also an almost-baked [routing](https://github.com/pocoproject/poco/tree/Route-develop) implementation sitting in a branch; it would fit nicely into NetEx.

aleks-f:
@RangelReale: Renamed SNMP branch to [NetEx](https://github.com/pocoproject/poco/tree/NetEx) and added asynchronous [HTTPClient](https://github.com/pocoproject/poco/commit/ecb310572fb80fa3ce43d4ea370e856933779d11). @obiltschnig: please check the HTTPClient code

aleks-f:
Closing this, will track development through [this issue](https://github.com/pocoproject/poco/issues/682).

PR #664, opened 2015-01-08
Don't automatically make copied files writable on Windows

cryptoknight:
Current Windows implementation of File makes copies of read-only files writable, which does not seem to be something that most users would want or expect.

### DIFF
--- a/Foundation/src/File_WIN32.cpp
+++ b/Foundation/src/File_WIN32.cpp
@@ -34,17 +34,17 @@ public:
 			FileImpl::handleLastErrorImpl(path);
 		}
 	}
-	
+
 	~FileHandle()
 	{
 		if (_h != INVALID_HANDLE_VALUE) CloseHandle(_h);
 	}
-	
+
 	HANDLE get() const
 	{
 		return _h;
 	}
-	
+
 private:
 	HANDLE _h;
 };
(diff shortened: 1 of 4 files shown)


aleks-f:
@obiltschnig ?

obiltschnig:
I'm pretty sure there's a reason for this behavior but I don't remember it. Probably consistency across platforms. I'm inclined to leave it as it is unless there's a compelling reason to change it.

cryptoknight:
There's no CopyFile equivalent function on *nix and the cp utility preserves permissions by default (unless a umask removes them), so I'm not sure which other platform this would be consistent with.

Reject the request if you wish and we'll just have to carry forward the patch in our local repo. We use copyTo() to implement some backup/restore functionality so we need it to not muck with file permissions in the process.

cryptoknight:
Also, it looks like the UNIX version of copyToImpl preserves permissions. The copied filed is created with the same mode as the source file.

aleks-f:
Yes, the UNIX version does copy the permissions of the original file; it seems it would make sense to have the same thing on Windows. However, your pull only changes the [File_WIN32U.cpp](https://github.com/pocoproject/poco/blob/develop/Foundation/src/File_WIN32U.cpp). There are also [File_WIN32.cpp](https://github.com/pocoproject/poco/blob/develop/Foundation/src/File_WIN32.cpp) and [File_WINCE.cpp](https://github.com/pocoproject/poco/blob/develop/Foundation/src/File_WINCE.cpp).

cryptoknight:
Will patch those as well and resubmit

obiltschnig:
Ok, looked at the change history. The change to make the copy writable has been done in May 2007 "to be consistent with other platforms". So if the other platforms now preserve the permissions, so should Windows.

aleks-f:
@cryptoknight : could you please also write a test for this so we get it bullet-proof for the future

cryptoknight:
Good to go

PR #666, opened 2015-01-08
Catch and log failures of individual Subsystems to (un)initialize

cryptoknight:
Currently if any Subsystem throws an exception while an Application is initializing or uninitializing, it will propagate to the caller and any subsequent Subsystems in the list will fail to be initialized or uninitialized. It seems like this would leave the Application in an indeterminate state and that instead we should make sure that every Subsystem init/uninit gets called regardless of whether some of them throw exceptions.

### DIFF
--- a/Util/src/Application.cpp
+++ b/Util/src/Application.cpp
@@ -94,12 +94,12 @@ Application::~Application()
 void Application::setup()
 {
 	poco_assert (_pInstance == 0);
-	
+
 	_pConfig->add(new SystemConfiguration, PRIO_SYSTEM, false, false);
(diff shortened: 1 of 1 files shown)


PureAbstract:
FWIW, I'm not convinced "log it and carry on" is appropriate if subsytem *initialization* fails.

aleks-f:
I agree. @obiltschnig ?

cryptoknight:
Yeah. I realized as I was writing this pull request up that I could have put the try/catch block in our master subclass of Subsystem and gotten the same effect without modifying Poco. If you want to reject this one, go ahead. Probably safer for existing users, anyway.

Issue #667, opened 2015-01-10
install_name not set correctly on OSX


quantumsteve:
The install name of shared libraries is currently set to the build directory instead of the install directory. Currently package managers such as macports and homebrew must fix these linking issues by either patching the config file or using install_name_tool. Fixing this directly in poco would be a much better solution. 

https://github.com/Homebrew/homebrew/pull/35716
https://trac.macports.org/browser/trunk/dports/devel/poco/Portfile#L74

PR #670, opened 2015-01-12
set install_name to POCO_PREFIX/lib

quantumsteve:
This modifies the install name of shared libraries so the path is the install directory instead of the build directory. 

https://github.com/pocoproject/poco/issues/667


obiltschnig:
Not entirely happy with that. Breaks things if you're not using the global Makefile (e.g., for me as a POCO maintainer). Logic to determine install name needs to be a bit more advanced.
1) Don't use POCO_PREFIX, as it's not defined if you're building without the global Makefile
2) In build/config/global, define POCO_LIB_INSTALLDIR to either POCO_PREFIX/lib (if POCO_PREFIX is defined), or to $(LIBPATH) (same as previous behavior).
3) Use POCO_LIB_INSTALLDIR for setting install name.

Issue #297, opened 2013-09-27
MemoryOutputStream and MemoryInputStream don't support seekp/seekg


jbitoniau:
I'd like to know the position in a memory stream (MemoryOutputStream  or MemoryInputStream) where the next read/write will occur. 
Unfortunately this doesn't seem supported by these classes as the tellp/g methods systematically return -1

Here's a test case:

```
Poco::Buffer<char> buffer( 1024 );
Poco::MemoryOutputStream ostr( buffer.begin(), buffer.size() );
ostr << 'H';
ostr << 'e';
ostr << 'l';
ostr << 'l';
ostr << 'o';
ostr << '\0';
std::streamoff np = ostr.tellp();  //  <=== This returns -1
assert( np==6 );                      //  <=== This fails!
printf("np=%d\n", np);

Poco::MemoryInputStream istr( buffer.begin(), buffer.size() );
char c;
istr >> c;  printf("c=%c\n", c );
istr >> c;  printf("c=%c\n", c );
istr >> c;  printf("c=%c\n", c );   //  <=== This returns -1
std::streamoff ng = istr.tellg();    //  <=== This fails!
assert( ng==3 );
printf("ng=%d\n", ng);
```

PR #688, opened 2015-01-25
Issue #297 - Added implementation of seekoff to BasicMemoryStreamBuf

martin-osborne:
Hi, I've been taking a look at issue #297.  In order to get `tellp`, `tellg`, `seekp` and `seekg` working it's necessary to provide an override of `seekoff`.  The C++ Standard gives in **`lib.stringbuf.virtual`** quite a detailed description of the the implementation of `seekoff`, and those details apply to `streambuf` (see **`lib.streambuf.virt.buffer`**).  In the attached code the calculation of `newoff`, the range check to ensure the new pointer position lies within the memory range, and all of the error conditions are all taken from the C++ Standard.  The only part of the implementation that is anything like my own work are the calls to `setg` and `pbump`, everything else is just translation.

There isn't anything magical about the code, it is simple pointer arithmetic, but I know it can be easy to make off-by-one errors, so I hope the fact it is mostly taken from the C++ Standard will give some reassurance, as will the tests I've added.


aleks-f:
Thanks for this contribution, @martin-osborne . I compiled and ran the test on OSX and Windows, all looks good.

obiltschnig:
@martin-osborne or @aleks-f  Could you please look into the testsuite issues (see above comment). The TeamCity build is broken for various Visual Studio versions.

aleks-f:
I guess it would have helped if I had pulled the changes on Windows before running tests, lol.

Anyway, I can't think of a non-standard library solution to the problem that would make this ambiguity go away. Here are two workarounds:
```
assert (4 == istr.tellg());
assert (static_cast<int>(istr.tellg()) == 4);
```

FWIW, someone thought this should [also happen in gcc](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49759) but the std library authors disagreed.

aleks-f:
compile errors [fixed](https://github.com/pocoproject/poco/commit/b2e209bf54d1135d891ea42b2b87b1960a8aa719)

martin-osborne:
@aleks-f - Thanks for fixing this.  I'll be sure to test 64bit VS2012 builds in future, as it would also have exposed the problem.

PR #702, opened 2015-01-29
class Cipher expressively says that it doesn't copy or move

cppfool:
In C++14 we expressively delete what we don't need, and therefore we don't need
to make copy ctor or copy assignment operators private.
And defalted definition of defalt ctor and defaulted definition of destructor works
well so we shouldn't intervene.

### DIFF
--- a/Crypto/include/Poco/Crypto/Cipher.h
+++ b/Crypto/include/Poco/Crypto/Cipher.h
@@ -99,7 +99,10 @@ public:
 		ENC_BINHEX_NO_LF = 0x82, /// BinHex-encoded output, no linefeeds
 		
 	};
-
+	
+	Cipher(Cipher&&) = delete;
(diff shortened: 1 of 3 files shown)


aleks-f:
@cppfool : This will have to wait for 2.0. Also, Travis build is failing

cppfool:
Yes, 2.0 will be better. Travis build is failing with *older* gcc, modern
gcc will pass it
as clang has done.
I want to contribute C++14 code. How should you I fit in there?
Thanks.

aleks-f:
Unless someone steps in to maintain the third parallel branch (right now, we have 1.6.1 and develop, which will become 1.7.0), you'll have to wait until we release 1.7. The we will branch off 1.7.1 and turn develop into 2.0, or something along those lines.

PR #708, opened 2015-02-04
Make regex compatible with VERSIONS like "1.6.0-all (2014-12-22)".

saschazelzer:
Without the fix, the patch number contained spaces which led to errors on Linux with CMake because the generated symlinks contained spaces.


bachp:
I agree your change fixes the issue with the version number and is more robust.
@aleks-f: please merge

@saschazelzer: However I'm wondering how you ended up with such a version string?

saschazelzer:
Thanks for merging. The VERSION file in http://pocoproject.org/releases/poco-1.6.0/poco-1.6.0-all.tar.gz contains the string `1.6.0-all (2014-12-22)`

bachp:
I didn't know that. Thanks for fixing it.

PR #721, opened 2015-02-23
Fix for BasicFIFOBuffer next()-based writing

Sevalecan:
This is the first pull request I've ever made, so let me know where I can improve.

This should fix issues with the StreamSocket implementation writing to a BasicFIFOBuffer derivative using the next() function to retreive a pointer to the first unused byte after the existing data. I couldn't think of a cleaner way other than just moving the memory to the front of the buffer otherwise weird stuff could happen with pointers, or 'iterators' if you prefer becoming invalid, so there's a chance for a small performance hit if you're not going to actually write to the end of the buffer.

I was told to make a test case, so I went and made a test suite for the BasicFIFOBuffer and included it in the commits(I hope this wasn't going too far, and hopefully I integrated it with the build properly):

```
// String length is 88 characters.
std::string text("The Quick Brown Dog Jumps Over The Lazy Fox.");
BasicFIFOBuffer<char> buffer(128);
buffer.write(text.data(), text.size());
char c_buffer[buffer.size()];
std::memset(c_buffer, 0, buffer.size());

buffer.read(c_buffer, 4);

assert(std::string(c_buffer, 4) == std::string("The "));

buffer.peek(c_buffer, buffer.used());
assert(std::string(c_buffer, buffer.used()) ==
std::string("Quick Brown Dog Jumps Over The Lazy Fox."));

memcpy(buffer.next(), "The ", 4);
buffer.advance(4);

buffer.peek(c_buffer, buffer.used());

assert(std::string("Quick Brown Dog Jumps Over The Lazy Fox.The ") == std::string(c_buffer, buffer.used()));
```

### DIFF
--- a/Foundation/include/Poco/FIFOBuffer.h
+++ b/Foundation/include/Poco/FIFOBuffer.h
@@ -352,6 +352,11 @@ public:
 	T* begin()
 		/// Returns the pointer to the beginning of the buffer.
 	{
+		if (_begin != 0)
+		{
+			std::memmove(_buffer.begin(), _buffer.begin() + _begin, _used * sizeof(T));
+			_begin = 0;
+		}
 		return _buffer.begin();
 	}
 
@@ -360,6 +365,12 @@ public:
 	{
 		if (available() == 0)
 			throw InvalidAccessException("Buffer is full.");
+
+		if (_begin != 0)
+		{
+			std::memmove(_buffer.begin(), _buffer.begin() + _begin, _used * sizeof(T));
+			_begin = 0;
+		}
 
 		return _buffer.begin() + _used;
 	}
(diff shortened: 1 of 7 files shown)


aleks-f:
Thanks for the contribution. Few comments:

What I had in mind is eliminating all _buffer.begin() calls except one - the one in FIFOBufer::begin(), which should return the "sliding" beginning of the buffer (_buffer.begin() + _begin). All other places where beginning is needed should call FIFOBuffer::begin().

Protect next() (and any other member that should be) call with mutex.

There are three FIFOBuffer tests already in [Foundation::CoreTest](https://github.com/pocoproject/poco/blob/develop/Foundation/testsuite/src/CoreTest.cpp). Please move those to your new FIFOBufferTest class.

Thanks again.

PR #730, opened 2015-03-04
Add setaffinity method to thread implementation and interface class

bas524:
Limit specified thread to run only on the specified core.

### DIFF
--- a/Foundation/CMakeLists.txt
+++ b/Foundation/CMakeLists.txt
@@ -108,7 +108,7 @@ endif (${CMAKE_CXX_COMPILER_ID} MATCHES "SunPro")
 add_definitions( -DPCRE_STATIC)
 
 # For SetAffinity
-if(UNIX)
+if(UNIX AND NOT APPLE)
     INCLUDE (CheckFunctionExists)
     INCLUDE (CheckCXXSourceCompiles)
     CHECK_FUNCTION_EXISTS(pthread_setaffinity_np HAVE_PTHREAD_SETAFFINITY_NP)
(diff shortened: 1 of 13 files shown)


aleks-f:
Thanks, can you do few more things before merge:
- add getAffinity()
- add testcase
- throw Poco::NotImplementedException instead of silently doing nothing for platforms where not implemented

bas524:
Thread affinity support for *nix, windows and mac os x 
Sorry for the indentation

bas524:
Sorry for the indentation

aleks-f:
could you please fix the indentation before I merge it? I know it may seem like a minor issue but when you have to deal with a few merges every day, it becomes quite a task to keep everything in order.
thanks

bas524:
Have you got config file for astyle, or something like astyle. ? 


Отправлено с устройства Samsung

-------- Исходное сообщение --------
От: Aleksandar Fabijanic <notifications@github.com> 
Дата: 2015.03.07  21:19  (GMT+03:00) 
Кому: pocoproject/poco <poco@noreply.github.com> 
Копия: Alexander B <bas524@ya.ru> 
Тема: Re: [poco] Add setaffinity method to thread implementation and
  interface class (#730) 

could you please fix the indentation before I merge it? I know it may seem like a minor issue but when you have to deal with a few merges every day, it becomes quite a task to keep everything in order.

thanks

aleks-f:
Unfortunately, no. We have [Style Guide](http://www.appinf.com/download/CppCodingStyleGuide.pdf) but it has not been formalized into a tool/configuration. We would not mind a contribution based on, say [LibFormat](http://clang.llvm.org/docs/LibFormat.html).

bas524:
merge it please, and i create a config for clang-format.

PR #738, opened 2015-03-07
Move other FIFOBuffer tests to new BasicFIFOBufferTest

Sevalecan:
I moved the other FIFOBuffer tests over to the new version I had created as requested in the last pull request. I can move the other Buffer Tests into their own set if no one thinks this violates the "core tests" idea that I wasn't aware of at the time of initially writing these.

Also, my original intent with these FIFOBuffer fixes was to fix the stable branch for 1.6.1. I can create a patch set and apply it to a branch based on the current 1.6.1 branch to patch the next() writing issue if there's no objection.


aleks-f:
It's ok to move the Buffer, if you wish. Back porting to 1.6.1 is fine, too. We will, however, have to regenerate the VS projects in both branches.

Sevalecan:
I'll happily do that, though I'm not sure how. I've looked a little now at the build_*.cmd files and am trying to see if I can maybe get one of them to run with the free version of Visual Studio (I tend to do my development with MinGW in Windows or GCC on Linux). I'll let you know if I can figure it out.

PR #739, opened 2015-03-07
Add WebSocket::receiveFrame() that appends to a Poco::Buffer<char>

tml1024:


tencek:
Is there a plan when this feature appears in a stable release?

PR #743, opened 2015-03-11
add thread affinity policy to threadpool

bas524:
there are OS_DEFAULT, UNIFORM_DISTRIBUTION and CUSTOM
With custom policy we can run thread on specified cpu


bas524:
I think this pull request is usefull

aleks-f:
There's no need for typedef enum; can you please change the enum to:
```c++
enum ThreadAffinityPolicy
{
	OS_DEFAULT = 0,
	UNIFORM_DISTRIBUTION,
	CUSTOM
};
```

PR #748, opened 2015-03-17
Enable pocoproject to be built as a subproject in a cmake build.

tik:
When the poco sources are added via add_subdirectory() to a CMakeLists.txt, CMAKE_SOURCE_DIR makes  cmake look for the libversion and VERSION files in the top-most source dir, not in the poco project directory.


bachp:
Good catch. I see how this might be useful.
@obiltschnig good to merge

PR #751, opened 2015-03-18
Adding ostream operator for Poco::Net::IPAddress

rklabs:
I found that there is no ostream operator for Poco::Net::IPAddress. Hence adding the same which will result in less typing :). I have modified NetworkInterfaceTest.cpp to ensure ostream operator is called. I also ran NetworkInterfaceTest after re-building the project and it passed successfully.

### DIFF
--- a/Net/include/Poco/Net/NetworkInterface.h
+++ b/Net/include/Poco/Net/NetworkInterface.h
@@ -343,6 +343,7 @@ inline bool NetworkInterface::operator == (const NetworkInterface& other) const
 
 
 Net_API std::ostream& operator<<(std::ostream& os, const Poco::Net::NetworkInterface::MACAddress& mac);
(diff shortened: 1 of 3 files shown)


obiltschnig:
1) why is this in NetworkInterface and not IPAddress
2) if we add this for IPAddress, we may as well add it for SocketAddress

rklabs:
1. Yes you are right,  this should be in IPAddress. I found ostream operator for MACAddress in NetworkInterface, hence added in same place
2. Sure, will add ostream operator for SocketAddress
3. I didn't understand. IPAddress.h is already included in NetworkInterface.h. Can you please let me know which #include is missing?

obiltschnig:
Thanks, but I've already done it.

rklabs:
Cool!

Issue #643, opened 2014-12-17
Regex: match all


DEvil0000:
i would suggest a enhancement for regex.
if you extract parts of a string with capturing groups you can only get the first or last group matches.
i would suggest a method like matchall() or support for /g option (i think this is the global option) to find all matches of a regex in a string.

example:
https://regex101.com/
php regex
regex: /abc(\d)/g
string to match: abc1xyzabc2xyzabc3
expected result:
matches=3
match1: "1"
match2: "2"
match3: "3"

martin-osborne:
@DEvil0000 

I've started to prepare a pull request for this.  However, in the example above, the matches you list are for sub-expression 1, which seems odd to me.  I'd expect the matches to be for sub-expression zero, i.e. "abc1", "abc2", and, "abc3".  In what I'm preparing to submit the first form of `matchAll` will return `["abc1", "abc2", "abc3"]`.  The second form of `matchAll` will return `[["abc1", "1"], ["abc2", "2"], ["abc3", "3"]]`, so it should give you what you need.

```
    int matchAll(const std::string& subject, std::string::size_type offset, MatchVec& matches, int options = 0) const;
        /// Matches all the non-overlapping occurances of the subject string, starting at offset, against the pattern. 
        /// Returns the position of each matched substring in matches.
        /// If no part of the subject matches the pattern, matches is empty.
        /// Throws a RegularExpressionException in case of an error.
        /// Returns the number of matches.

    int matchAll(const std::string& subject, std::string::size_type offset, std::vector<MatchVec>& matches, int options = 0) const;
        /// Matches all the non-overlapping occurances of the subject string, starting at offset, against the pattern. 
        /// For each match it returns the captured text for each subexpression in matches[i].
        /// If no part of the subject matches the pattern, matches is empty.
        /// Throws a RegularExpressionException in case of an error.
        /// Returns the number of matches.
```

martin-osborne:
I should clarify that you can achieve the results in the example you posted using a positive lookahead `(?=abc)\d`, and the first form of `matchAll`.  The second form is still required (I believe) to find overlapped matches using `(?=(xyz))`.

DEvil0000:
I was using the regex `/abc(\d)/g` in my example so the matches '1','2' and '3' are perfectly expected. It is returning the captured matches for the part in brackets (capturing group). As far as i know this is perfectly expected.
If you write it like `(abc\d)` then you will match 'abc1' and so on.
[...]

PR #774, opened 2015-04-10
Issue #643 - Added support for `matchAll`

martin-osborne:


martin-osborne:
> Hallo Martin, I cannot find declaration of GroupMap ... what and where is it ?

Hi,

It's a typedef on line 84 (or thereabouts) of `RegularExpression.h`:

`typedef std::map<int, std::string> GroupMap;`

Issue #619, opened 2014-11-24
Replace std::auto_ptr with std::unique_ptr


obiltschnig:
For 2.0, remove the few uses of std::auto_ptr and use std::unique_ptr instead.

PR #777, opened 2015-04-13
Replace std::auto_ptr with std::unique_ptr

trusch:
This should solve issue #619.

This pull requests also contains a little change to the CMakeLists.txt file. It adds the "-std=c++" switch to CMAKE_CXX_FLAGS to enable c++11 compiler support. I don't know if this is right in the toplevel CMakeLists file, please let me know if it should be done in the approriate subfiles.

### DIFF
--- a/ApacheConnector/src/ApacheConnector.cpp
+++ b/ApacheConnector/src/ApacheConnector.cpp
@@ -174,21 +174,21 @@ extern "C" int ApacheConnector_handler(request_rec *r)
 		if ((rv = ap_setup_client_block(r, REQUEST_CHUNKED_DECHUNK))) 
 			return rv; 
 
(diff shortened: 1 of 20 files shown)


rafzi:
For GCC 4.6 the parameter is named "c++0x". This should fix the failing tests.

bschramke:
You should not set compile flags directly in CMakelist. Like rafzi pointed out, different compilers handle things different.

CMake defines a bunch of flags & variables for this. So instead of "-std=c++11" use 
 
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED on)

trusch:
This is pointless due to PR  #778.

PR #778, opened 2015-04-14
C++11move/rvalue ctor support for Dynamic::Var, Buffer and std::auto_ptr cleanup

mkrivos:


trusch:
+1

This makes my own pullrequest pointless (only the auto_ptr changes)!
Am 14.04.2015 07:47 schrieb "Marian Krivoš" <notifications@github.com>:

aleks-f:
Travis build is failing, can you check it? 

Also, instead of peppering code with #if __cplusplus >= 201103L, can we have the POCO_ENABLE_CPP11 defined based on it:
```
#if __cplusplus >= 201103L
  #define POCO_ENABLE_CPP11 
#endif
```

aleks-f:
And, if we decided to drop older compilers, then I guess we shouldn't be #idef-ing C++11 code. @obiltschnig ?

mkrivos:
Yes, we must define some more policies:
1. when exactly we enable POCO_ENABLE_CPP11 macro ? By configuration paremeter? By autodetection (cmake) ? 
2. Nor gcc 4.6 nor vs2013 are not fully CPP11 compatible. We can set VS2012 CPP11 features as the base (gcc 4.6 is rather VS superset from this view). Feature that is not covered by VS we can use conditionally with POCO_ENABLE_CPP11 macro ?

Any ideas?

mkrivos:
@aleks-f 
It seems that error is from missing c++11 support (not enabled by default).
Which way we use for c++11 enablement for standard makefile build? Cmake build is easy.

aleks-f:
POCO_ENABLE_CPP11 can be enabled in Config.h automatically, based on __cplusplus__ macro version; I see you have added CMake check; we can also add the flag to configure script for posix builds. Given what we are trying to do, it should probably be renamed to POCO_ENABLE_FULL_CPP11

VS2012/gcc.4.8 should be a "point of no return" and C++11 features desired but not available there should be compiled conditionally. At this time, the only thing I am immediately interested in are variadic templates (for make_shared, not available in VS2012).

mkrivos:
@aleks-f what next?

aleks-f:
Waiting for @obiltschnig to look over this.

RangelReale:
I included this patch in my c++11 port, so I think this can be closed.

PR #784, opened 2015-04-20
Update CMake

Torndeco:
Added cmake/(c/cxx)_flag_overrides.cmake
Added Program Files(x86)/MySQL to search path for MySQL

Windows Builds
Added Option to select /MT  or  /MD builds
Librarys + dlls now get proper postfixes   mdd / md / mtd / mt  etc...


Torndeco:
Its abit rough around the edges, but it might be of use to someone.
Mainly just changes for building Poco using CMake on Windows.

Things of note
C / CXX Flag Overrides could do with someone reviewing + making sure they are ok.
C++ is really just a hobby

The windows builds could be improved by doing custom build types that way instead of Release / Debug / ReleaseWithDebInfo, you could have shared_mt static_md etc...

Torndeco:
Just done some quick changes,  C + CXX Flags aren't overridden unless someone uncomments the files.

One issue is the CMake Find OpenSSL will only find MD libraries + not MT libraries. Tried to just do a string replace, but couldn't get it to work.

------------

As for C++11 is there an easier method besides editing each cmake file for each library to set_property ?

aleks-f:
@bachp : I'll wait for your approval before merging this

bachp:
@Torndeco For setting the CXX_STANDARD there is a variable [CMAKE_CXX_STANDARD](http://www.cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD.html#variable:CMAKE_CXX_STANDARD) that sets the default standard for all targets.
I also read trough http://www.cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html to see if this might make things easier. Maybe you want to have a look too.

Regarding the FindOpenSSL issue. Can you create a seperate issue for that and explain the issue a bit more in detail (What OpenSSL used? How are the libraries called?, ...)

For the overriding of the default flags I'm not a fan of this. But I'm not sure what the general policy is.
Maybe @aleks-f  or @obiltschnig have an oppinion here or can explain what is the best practice in the Makefiles.

Torndeco:
Am just going to close this and use the changes myself...

---------------

To much debate over giving making it easier to override the default C / CXX Flags assigned via CMake.
Its similar to someone editing build/config files to change the C / CXX Flags used to compile Poco.

As for FindOpenSSL issue when building using /MT
Info @ https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L88-L91

bachp:
@Torndeco I'm sorry it was not my intention to discourage you from contributing.
But I believe overriding the C and CXX flags to your specific needs is better done in a initial cache file using the [`cmake -C`](http://www.cmake.org/cmake/help/git-master/manual/cmake.1.html) option.

The content of the file would look like this
```
set(CMAKE_CXX_FLAGS_DEBUG_INIT "/D _WINDLL /Zi /WX- /Oy- /EHsc /GS- /arch:SSE2 /fp:precise /Od /RTC1 /D_DEBUG")
set(CMAKE_CXX_FLAGS_MINSIZEREL_INIT       "/WX- /O1 /Oy- /EHsc /GS- /arch:SSE2 /fp:precise /D NDEBUG")
set(CMAKE_CXX_FLAGS_RELEASE_INIT          "/WX- /O2 /Oy- /EHsc /GS- /arch:SSE2 /fp:precise /D NDEBUG")
set(CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT   "/Zi /O2 /Oy- /EHsc /GS-  /arch:SSE2 /fp:precise /D NDEBUG")
```

This way you would not need to modify the files from Poco and don't have any conflicts with changes coming in via git.

Issue #785, opened 2015-04-20
GCC_DIAG_* definitions are missing for Windows (MinGW)


procedural:
https://github.com/pocoproject/poco/blob/1b70142/Foundation/include/Poco/Platform_POSIX.h#L47-L76

They're defined only for POSIX platforms, when trying to compile master branch on Windows with MinGW it throws errors because of undefined GCC_DIAG_* macros.

aleks-f:
Fix it and send pull request pleas

procedural:
You mean copy and paste GCC_DIAG_* macros to Platform_WIN32.h?

aleks-f:
Of course not. I mean #ifdef them out for mingw

aleks-f:
Alternatively, GCC_DIAG should be moved to it's own header and included from Platform_POSIX and Platform_Win32 (for MinGW). Reopening, perhaps someone will be interested to do this. I have no MinGW environment handy.

procedural:
How would you like to name this header file?

procedural:
PR https://github.com/pocoproject/poco/pull/787

aleks-f:
Just out of curiosity, I tried MinGW and the build is broken - when <cmath> is included, these should be in std namespace:

```
In file included from include/Poco/FPEnvironment.h:27,
                 from include/Poco/NumericString.h:25,
                 from include/Poco/NumberFormatter.h:24,
                 from include/Poco/ArchiveStrategy.h:27,
                 from src/ArchiveStrategy.cpp:17:
include/Poco/FPEnvironment_DUMMY.h: In static member function `static bool Poco:
:FPEnvironmentImpl::isInfiniteImpl(float)':
include/Poco/FPEnvironment_DUMMY.h:78: error: `isinf' is not a member of `std'
include/Poco/FPEnvironment_DUMMY.h: In static member function `static bool Poco:
:FPEnvironmentImpl::isInfiniteImpl(double)':
include/Poco/FPEnvironment_DUMMY.h:84: error: `isinf' is not a member of `std'
include/Poco/FPEnvironment_DUMMY.h: In static member function `static bool Poco:
:FPEnvironmentImpl::isInfiniteImpl(long double)':
include/Poco/FPEnvironment_DUMMY.h:90: error: `isinf' is not a member of `std'
include/Poco/FPEnvironment_DUMMY.h: In static member function `static bool Poco:
:FPEnvironmentImpl::isNaNImpl(float)':
[...]

PR #787, opened 2015-04-21
Moving GCC diagnostic macros to a separate header file. Fixes https://github.com/pocoproject/poco/issues/785

procedural:
@aleks-f feel free to guide this PR as needed.

Tested and works on Windows with MinGW-w64 compiler.


aleks-f:
Let's call it Compiler_GCC.h and include it from only one place - the [GCC segment of Platform.h](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Platform.h#L233).

procedural:
Seems like clang can't find the declaration of [GCC_DIAG_OFF](https://travis-ci.org/pocoproject/poco/jobs/59410845#L3846) (line 3846).

Uh.

Is it absolutely necessary to hide GCC warnings at all? POCO 1.6.0 dealt with it fine, why add code complexity just for flags that should be defined by the user who don't want to see these warnings.

PR #788, opened 2015-04-22
Zip64 support

bobstabler:
I implemented Zip64 support in the poco\Zip code.  As far as I know, I haven't changed any functionality, unless the caller explicitly asks for zip64, when creating a zip file using Compress(), or adds a file that is bigger that 4GB-1 or which makes the archive bigger than 4GB-1.  All of the unit tests still pass on Windows (although I only built/tested using VS2012), OSX and Linux.  I am able to read the zip64 archives that this new code is creating with WinZip, 7Zip and the Apache Java Zip code.


obiltschnig:
Thank you!

bobstabler:
You're very welcome.  We love Poco at Avid!!!

aleks-f:
@bobstabler : could you share a small paragraph on how you use poco, so we can post it [here](http://pocoproject.org/forum/viewtopic.php?f=11&t=3826)

bobstabler:
We are using Poco for a project involving clients running on MACs and Windows communicating with services running on Linux on the Amazon Cloud.  Some of the code is shared between the three platforms, so it has been very convenient to be able to maintain a single codebase.

PR #796, opened 2015-04-30
Added new static method to OpenSSLInitializer, disableSSLInitialization()

bobstabler:
This allows the caller to inform Poco that the Open SSL library has already been initialized outside of Poco, for example in main(), so it doesn't re-initialize it, and, more importantly, uninitialized it when the last OpenSSLInitializer goes out of scope.

### DIFF
--- a/Crypto/include/Poco/Crypto/OpenSSLInitializer.h
+++ b/Crypto/include/Poco/Crypto/OpenSSLInitializer.h
@@ -44,7 +44,7 @@ namespace Crypto {
 
 
 class Crypto_API OpenSSLInitializer
-	/// Initializes the OpenSSL library.
+	/// Initalizes the OpenSSL library.
 	///
 	/// The class ensures the earliest initialization and the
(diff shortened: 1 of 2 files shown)


aleks-f:
Thanks for the contribution. Would you mind adding also automatic initialization like the one we have in Net and Data libraries (see [blog](http://pocoproject.org/blog/?p=741) for details.

bobstabler:
I’ll take a look.  

P.S. If I add new files, do I have to update a whole mess of project files?

aleks-f:
On Posix, you'll have to add the .cpp file name to Makefile. On Windows, you can only add it to a VS solution version of your choice to compile/test, I'll generate the rest of the projects. Or, if you feel inclined, try [ProGen](https://github.com/pocoproject/poco/tree/develop/ProGen) yourself ;).

bobstabler:
Thanks.  I probably won't get to look at it for a while.

PR #797, opened 2015-05-01
Support for Sybase ODBC and multiple result set fixes

kostya-lnk-ms:
The change contains two main features:
1) Support for Sybase ODBC
2) Support for reading multiple result sets via the RecordSet
Bug fixes: overflow of when reading long Numeric, seg. fault on Linux when binding vector of strings. Tests have been changed to allow concurrent runs on separate hosts.


aleks-f:
I looked briefly over this pull. It's a lot of changes and, for the most part, they look like very good fixes/additions - thank you.

One thing I'm not so keen about is removing default parameters in Binder, Preparator etc - sometimes adding parameters is necessary but, if at all possible, they should not break the interface; that said, why not provide default value for numericToString constructor parameter and leave the defaults for existing parameters intact?

In any case, I need to review all this more closely and that will take some time. I will check out your branch, add it as separate branch in poco repo and we can work on that. BTW, we are looking for Poco::Data maintainer, are you interested?

aleks-f:
Created [ODBCSybase branch](https://github.com/pocoproject/poco/tree/ODBCSybase).

aleks-f:
Also, this breaks other data back-ends because of the [changes to Data::Statement](https://github.com/kostya-lnk-ms/poco/commit/48cb050daf10d46c067fa8e870c07759a3a4d218), can you please take a look at that?

kostya-lnk-ms:
I'll fix the build problems - I missed them. As to default parameters - the reason for removing them was that at all the call sites these seem to always have a defined values. I'll double check that and can revert to some defaults, if you think it is better.
Thanks for maintenance offer, I need to check if current commitments would allow me doing that.

PR #801, opened 2015-05-05
Odbc sybase - Fixed MySQL and SQLite compilation problems

kostya-lnk-ms:


aleks-f:
@intensifi : [this](https://github.com/kostya-lnk-ms/poco/commit/c40c9e9a26de65edff3ecb401bbdea744b68e03c) will affect PostgreSQL code, can you please update it. I would like to merge this branch and PostgreSQL for [1.7 release](https://github.com/pocoproject/poco/milestones/Release%201.7.0).

intensifi:
Aleks,

What is the currentDataSet concept?

Off the top of my head this doesn't sound like something postgresql supports. 

Best regards,

===
Jeff Adams

aleks-f:
It's  result of multiple query execution in one statement. See [SQLiteTest::testMultipleResults](https://github.com/pocoproject/poco/blob/develop/Data/SQLite/testsuite/src/SQLiteTest.cpp#L2519-L2558) for an example. If it is not supported, just update the interface and throw if it ever comes in as anything different than zero, like it was done in [MySQL](https://github.com/pocoproject/poco/blob/ODBCSybase/Data/MySQL/src/MySQLStatementImpl.cpp#L51-L56).

PR #807, opened 2015-05-07
Develop

mkrivos:
cmake: reenable 'd' postfix for debug poco libraries


mkrivos:
Hmm, it seems that travis has serious problem - this commit can't be intrusive ...

bachp:
Jep travis seems to have some troubles recently :(

I had a look at your change. Do you need the "d" postfix on non Windows systems too?
Because for windows it should be set here: https://github.com/pocoproject/poco/blob/develop/cmake/DefinePlatformSpecifc.cmake#L45
Maybe you can explain the problem you were facing?

@obiltschnig Do the Makefiles also create "d" sufixes? I suggest we should try to keep the CMake output consistent with the Makefile/MSProject outputs as much as possbile.

obiltschnig:
Yes, the Makefiles create a "d" suffix for debug builds of libraries.

bachp:
Ok I was not aware of that.
 I would suggest the following solution then:
https://github.com/siemens/poco/commit/28ac7e80dbd1de8515b827bcb84b14c59c3d547f

@mkrivos Could you check if this solves your problem?

mkrivos:
Seems to be OK.

PR #813, opened 2015-05-10
support new metacolumn method signature

intensifi:


intensifi:
update metacolumn method signature per ales

Issue #705, opened 2015-01-31
Error building with ENABLE_UTIL=ON and ENABLE_XML=OFF


Gamadril:
I don't need XML, but I do need UTIL for a command-line application / daemon. But seems that UTIL depends on XML. 

Building on OSX using cmake 3.0.2 with following configuration:
 ```
cmake -DENABLE_MONGODB=OFF -DENABLE_XML=OFF -DENABLE_DATA_MYSQL=OFF -DENABLE_DATA_ODBC=OFF -DENABLE_DATA=OFF -DENABLE_NETSSL=OFF -DENABLE_CRYPTO=OFF -DENABLE_ZIP=OFF -DENABLE_PAGECOMPILER=OFF -DENABLE_PAGECOMPILER_FILE2PAGE=OFF ..
```
make fails:
```
... /poco/Util/include/Poco/Util/XMLConfiguration.h:30:10: fatal error: 
      'Poco/DOM/Document.h' file not found
#include "Poco/DOM/Document.h"
         ^
1 error generated.
make[2]: *** [Util/CMakeFiles/Util.dir/src/Application.cpp.o] Error 1
make[1]: *** [Util/CMakeFiles/Util.dir/all] Error 2
make: *** [all] Error 2

```
Tested with 1.6.0 and git clone

**Comment by bachp** [comment]
Dependencies between components is not modeled in the build system at the moment.

It could be modeled using [CMakeDependentOption](http://www.cmake.org/cmake/help/git-master/module/CMakeDependentOption.html?highlight=dependent) or by just printing an error if Util is enabled but XML is not.

It might be worth to add at least such a check.

**Comment by aleks-f** [comment]
To compile Util without XML and JSON, [define POCO_UTIL_NO_JSONCONFIGURATION and POCO_UTIL_NO_XMLCONFIGURATION](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Config.h#L144-L153)

**Comment by Gamadril** [comment]
uncommented in Config.h
```
#define POCO_UTIL_NO_INIFILECONFIGURATION
#define POCO_UTIL_NO_XMLCONFIGURATION
```
still get an error:
```
Linking CXX shared library ../lib/libPocoUtil.dylib
ld: library not found for -lXML
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libPocoUtil.30.dylib] Error 1
make[1]: *** [Util/CMakeFiles/Util.dir/all] Error 2
[...]

PR #820, opened 2015-05-19
Fixed issue #705 Error building with ENABLE_UTIL=ON and ENABLE_XML=OFF

Gamadril:
Updated Util/CMakeLists.txt to link against JSON and XML libraries only if necessary

### DIFF
--- a/Util/CMakeLists.txt
+++ b/Util/CMakeLists.txt
@@ -24,7 +24,18 @@ set_target_properties( "${LIBNAME}"
     DEFINE_SYMBOL Util_EXPORTS
     )
 
-target_link_libraries( "${LIBNAME}" JSON XML Foundation)
+target_link_libraries( "${LIBNAME}" Foundation)
+if (ENABLE_XML)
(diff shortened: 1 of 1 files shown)


bachp:
Thanks. 
Looks good.

@aleks-f Please merge
I don't know if #821 is needed as-well?

aleks-f:
See [#821](https://github.com/pocoproject/poco/pull/821)

bachp:
Shouldn't this change go to develop too?

aleks-f:
Looks same like [#821](https://github.com/pocoproject/poco/pull/821). Am I missing something?

Gamadril:
This one is for the develop branch. #821 was for poco-1.6.1 branch. Since it is the current official stable one I thought it would be a good idea to update both...

aleks-f:
Ah that's what happens when working from phone - it doesn't show to which branch pull goes. Anyway, can't merge it from phone now, Travis is failing so it won't let me.

PR #823, opened 2015-05-19
Fix for PostgreSQL ODBC syntax error reporting

kostya-lnk-ms:


aleks-f:
GCC build is [failing](https://travis-ci.org/pocoproject/poco/builds/63196756)

kostya-lnk-ms:
My bad, I’ve used a local template which requires at least c++0x to compile.
Is there a plan to enable either C++0x or C++11 as a default option?

aleks-f:
Yes, with [this pull](https://github.com/pocoproject/poco/pull/778). @obiltschnig

Issue #814, opened 2015-05-10
CMake Android build


aleks-f:
Follow-up to Unable to build for android #812

peterholak:
Android NDK r13 now comes with an "official" cmake toolchain file https://github.com/android-ndk/ndk/wiki/Changelog-r13#ndk

Bjoe:
For the sake of completeness in `poco-1.9.1` and `develop` branch, there is support for Android build. There are also test builds on travis.
I think this issue is fixed.
#2555 Update CMake documentation for Android Platform is still missing.

PR #835, opened 2015-05-26
add 'log' to SYSLIBS when compiling Foundation for android

aaboyd:
This should help with #814 .

Without it you get the following error.
```
Linking CXX shared library ../lib/libPocoFoundation.so
CMakeFiles/Foundation.dir/src/AndroidLogChannel.cpp.o:AndroidLogChannel.cpp:function Poco::AndroidLogChannel::log(Poco::Message const&): error: undefined reference to '__android_log_print'
collect2: error: ld returned 1 exit status
make[2]: *** [lib/libPocoFoundation.so] Error 1
make[1]: *** [Foundation/CMakeFiles/Foundation.dir/all] Error 2
make: *** [all] Error 2
```

I was able to get the android build to work with cmake using the toolchain found https://github.com/taka-no-me/android-cmake .

### DIFF
--- a/Foundation/CMakeLists.txt
+++ b/Foundation/CMakeLists.txt
@@ -104,6 +104,10 @@ if (${CMAKE_CXX_COMPILER_ID} MATCHES "SunPro")
 	set_target_properties( "${LIBNAME}" PROPERTIES LINK_FLAGS "-library=stlport4")
 endif (${CMAKE_CXX_COMPILER_ID} MATCHES "SunPro")
 
+if(ANDROID)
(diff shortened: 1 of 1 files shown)


bachp:
Thanks. It would be nice if the android things work out of the box.
@aleks-f please merge

@aaboyd I think it would also make sense to add a CMake section to [AndroidPlatformNotes](https://github.com/pocoproject/poco/blob/develop/doc/99300-AndroidPlatformNotes.page). If you could document your steps there that would be super.

PR #843, opened 2015-05-29
Logger message: added %O for numeric OS thread id

matejk:
Log message: added %O option to insert numeric platform specific thread id. Useful when using Poco logger in non-Poco threads (like std::thread) to be able to distinguish between threads in log messages.

Thread ID implemented for Linux and OSX.

### DIFF
--- a/Foundation/include/Poco/Message.h
+++ b/Foundation/include/Poco/Message.h
@@ -128,6 +128,9 @@ public:
 	long getTid() const;
 		/// Returns the numeric thread identifier for the message.
 	
+	long getOsTid() const;
+		/// Returns the numeric OS thread identifier for the message.
+	
 	void setPid(long pid);
 		/// Sets the process identifier for the message.
 		
@@ -196,6 +199,7 @@ private:
 	Priority    _prio;
 	Timestamp   _time;
 	int         _tid;
+	int         _ostid;
 	std::string _thread;
 	long        _pid;
 	const char* _file;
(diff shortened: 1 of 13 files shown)


matejk:
The purpose of system-specific numeric thread id is to be able to use it logger to display thread ID when using non-Poco threads, like std::thread.

If currentTid() is always numeric that it can be used for this purpose, yes.

matejk:
Moved implementation from Message to Thread*.

However I noticed that only some posix platforms have different pthread and OS thread IDs and it is more convenient for tracing purposes to use OS tids.

One of the options would also be to change Posix implementation of Thread::getTid() to return OS numeric thread id instead of pthread_self (type pthread_t).

aleks-f:
The downside, as I mentioned, is a bit too many thread IDs for my taste; so, I'd modify ```Thread::currentTid()``` to return kernel ID, which decouples it from both Poco and the platform thread implementation. Threading now in practice boils down to pthread and win32 threads but who knows what future may hold. But, if we accept the proposal in this last pull and have three thread ID calls, I'm fine with that too.

obiltschnig:
Two thread IDs (one from Poco::Thread, one from system) should be enough. Three is definitely too many.

matejk:
Removed functions to get the third thread ID and rather changed implementation of tid() and currentTid() for posix implementation to return kernel threads for Linux and OS X.

If this implementation is acceptable, I'd clean create new pull request with clean code changes.

matejk:
Created new pull request #846 with cleaned up code changes.

PR #849, opened 2015-06-02
cmake improvements for msvc

pwm1234:
I made two small but important improvements for the cmake build using Visual Studio. 

The first enhancement is to add a cmake option ``ENABLE_MSVC_MP``. When ON, this option uses the /MP compiler flag for MSVC. This change required adding the option to the top-level ``CMakeLists.txt`` and then using the option in ``cmake/DefinePlatformSpecifc.cmake``. I have the default setting OFF, so it will not adversely affect anyone's current build scripts. But by enabling it, build speed scales to the number of cores used, which means on my 8-core laptop, poco builds 8 times faster when using this option. 

The second enhancement is to install the pdb files using the macro ``POCO_INSTALL_PDB`` added to the file ``cmake/PocoMacros.cmake``. Including pdb files in a development installation makes it much easier for other developers to reuse a build of poco. 

I hope you find this pull request useful.

### DIFF
--- a/Foundation/include/Poco/Message.h
+++ b/Foundation/include/Poco/Message.h
@@ -128,6 +128,9 @@ public:
 	long getTid() const;
 		/// Returns the numeric thread identifier for the message.
 	
+	long getOsTid() const;
+		/// Returns the numeric OS thread identifier for the message.
+	
 	void setPid(long pid);
 		/// Sets the process identifier for the message.
 		
@@ -196,6 +199,7 @@ private:
 	Priority    _prio;
 	Timestamp   _time;
 	int         _tid;
+	int         _ostid;
 	std::string _thread;
 	long        _pid;
 	const char* _file;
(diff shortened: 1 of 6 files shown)


bachp:
Hi @pwm1234 thanks for your pull request. I think your changes are good.

However I think it would make sense to squash your 6 commits together into two commits one for the POCO_INSTALL_PDB and one for the /MP change.

Would you be willing to do this?

pwm1234:
Yes, I am willing to squash the commits together. Unfortunately, I am a novice at squashing commits, and my attempt to do so closed this pull request. I made another pull request, [850](https://github.com/pocoproject/poco/pull/850) that is a repeat of this one with the 6 commits squashed into 2. 

Did I have to make a new pull request or is there a way I could have squashed the commits as part of this pull request? Also, as a general rule for future pull requests, should I squash commits before making the pull request? I am becoming a big fan of Poco and hope to make future contributions.
Thanks,
Phil

bachp:
Usually after squashing you need to do a force push to the branch you did the pull request from. I think this should work.

For your second question I think it is better to squash the commits into logical groups. This makes them easier to review.

We are looking forward to see more contributions from you.

PR #850, opened 2015-06-02
cmake improvements for msvc

pwm1234:
I made two small but important improvements for the cmake build using Visual Studio. 

The first enhancement is to add a cmake option ``ENABLE_MSVC_MP``. When ON, this option uses the /MP compiler flag for MSVC. This change required adding the option to the top-level ``CMakeLists.txt`` and then using the option in ``cmake/DefinePlatformSpecifc.cmake``. I have the default setting OFF, so it will not adversely affect anyone's current build scripts. But by enabling it, build speed scales to the number of cores used, which means on my 8-core laptop, poco builds 8 times faster when using this option. 

The second enhancement is to install the pdb files using the macro ``POCO_INSTALL_PDB`` added to the file ``cmake/PocoMacros.cmake``. Including pdb files in a development installation makes it much easier for other developers to reuse a build of poco. 

I hope you find this pull request useful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,9 @@ option(POCO_UNBUNDLED
 if(MSVC)
     option(POCO_MT
       "Set to OFF|ON (default is OFF) to control build of POCO as /MT instead of /MD" OFF)
+
+    option(ENABLE_MSVC_MP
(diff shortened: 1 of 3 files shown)


bachp:
Perfect. Thanks a lot.

PR #854, opened 2015-06-03
remove un-needed headers

larivierec:
These files were there multiple times and may be just a simple aesthetic problem.
They are included in [WebSocket.h](https://github.com/tml1024/poco/blob/develop/Net/include/Poco/Net/WebSocket.h)


aleks-f:
Thank you

PR #862, opened 2015-06-16
Add mysql_stmt_store_result

Torndeco:
If MySQL connection is lost after statement execution and before / during fetching the result.
Will segfault due to mysqlclient library referencing a null pointer.

Using mysql_stmt_store_result prevents this from happening and gives us extra error handling.
Downside is higher clientside memory usage
https://dev.mysql.com/doc/refman/5.0/en/mysql-stmt-store-result.html

### DIFF
--- a/Data/MySQL/src/StatementExecutor.cpp
+++ b/Data/MySQL/src/StatementExecutor.cpp
@@ -95,6 +95,9 @@ void StatementExecutor::execute()
 
 	if (mysql_stmt_execute(_pHandle) != 0)
 		throw StatementException("mysql_stmt_execute error", _pHandle, _query);
+		
+	if (mysql_stmt_store_result(_pHandle) != 0)
+		throw StatementException("mysql_stmt_store_result", _pHandle, _query);
 
 	_state = STMT_EXECUTED;
 


aleks-f:
Wouldn't indiscriminately applying this interfere with [limiting the number of fetched rows](https://github.com/pocoproject/poco/blob/375eca2ec4df30f0d86b8bca447f30ce9c96a000/Data/include/Poco/Data/Limit.h)?

Torndeco:
Not familiar with that part of the code, technically it won't break Limit Rows.
There is still the downside is increase memory usage since it will buffer all the result regardless.
Client really would prob be better of using the SQL LIMIT command itself imo.

But there is no other-way of catching / preventing the segfaults, which can be an issue with a remote Database.

aleks-f:
What happens if query returns million rows but you want to fetch only 100 at a time?

Torndeco:
Really should be using the SQL Limit Command in that case.
```
SELECT * FROM table_name LIMIT 0,100;
First 100 results
SELECT * FROM table_name LIMIT 100,200;
Second 200 results etc
```
Granted the command is slightly different depending on the SQL backend used.

Anyway you can use the small pull request or not.
Or maybe just make it an optional bool value to enable its use or not.

aleks-f:
The problem is that other back-ends may not provide same SQL functionality and, as a rule, we do not get involved in SQL parsing. This should be done as an optional feature for MySQL, false by default. See how [autoCommit/Bind/Extract](https://github.com/pocoproject/poco/blob/develop/Data/ODBC/src/SessionImpl.cpp#L143-L153) are done for ODBC.

PR #898, opened 2015-08-04
add POCO_EXPORT_CLASS_INTERFACE to ClassLibrary.h

zosrothko:
add POCO_EXPORT_CLASS_INTERFACE so that the same interface be implemented by many different libraries while having a unique name in the IoC controller.


zosrothko:
Hi
The commit 19f1b42 is a error and should be ignored/erased. It was generated by following the steps from the GitHub documentation about syncing a repository with commit behind, and obviously I missed something. Sorry for the noise.

PR #907, opened 2015-08-14
PatternFormatter Priority Name Enhancement

dedvalson:
Added a property to allow the user to specify the actual names for the priorities that will go into the log. This was required for our app to conform to RFC 5424.

The property is called "priorityNames".

It should contain a list of 8 names, separated by a period. For example:

"Fatal.Critical.Error.Warning.Notice.Information.Debug.Trace"


obiltschnig:
Made some minor changes for style and consistency: names must be separated by comma, and there's an exception if an invalid priorityNames value is provided.

PR #911, opened 2015-08-19
SQLite bump to faster version 3.8.11.1

mkrivos:
Changelog:
SQLite Release 3.8.11.1 On 2015-07-29

- Miscellaneous micro-optimizations result in 22.3% more work for the same number of CPU cycles relative to the previous release. SQLite now runs twice as fast as version 3.8.0 and three times as fast as version 3.3.9. (Measured using cachegrind on the speedtest1.c workload on Ubuntu 14.04 x64 with gcc 4.8.2 and -Os. Your performance may vary.)

- Restore an undocumented side-effect of PRAGMA cache_size: force the database schema to be parsed if the database has not been previously accessed.
- Fix a long-standing problem in sqlite3_changes() for WITHOUT ROWID tables that was reported a few hours after the 3.8.11 release.

- Added the experimental RBU extension. Note that this extension is experimental and subject to change in incompatible ways.
- Added the experimental FTS5 extension. Note that this extension is experimental and subject to change in incompatible ways.
- Added the sqlite3_value_dup() and sqlite3_value_free() interfaces.
- Enhance the spellfix1 extension to support ON CONFLICT clauses.
- The IS operator is now able to drive indexes.
- Enhance the query planner to permit automatic indexing on FROM-clause subqueries that are implemented by co-routine.
- Disallow the use of "rowid" in common table expressions.
- Added the PRAGMA cell_size_check command for better and earlier detection of database file corruption.
* Added the matchinfo 'b' flag to the matchinfo() function in FTS3.
- Improved fuzz-testing of database files, with fixes for problems found.
- Add the fuzzcheck test program and automatically run this program using both SQL and database test cases on "make test".
- Added the SQLITE_MUTEX_STATIC_VFS1 static mutex and use it in the Windows VFS.
- The sqlite3_profile() callback is invoked (by sqlite3_reset() or sqlite3_finalize()) for statements that did not run to completion.
- Enhance the page cache so that it can preallocate a block of memory to use for the initial set page cache lines. Set the default preallocation to 100 pages. Yields about a 5% performance increase on common workloads.
- Added the sqlite3_result_zeroblob64() and sqlite3_bind_zeroblob64() interfaces.
Important bug fixes:

- Fix CREATE TABLE AS so that columns of type TEXT never end up holding an INT value. Ticket f2ad7de056ab1dc9200
- Fix CREATE TABLE AS so that it does not leave NULL entries in the sqlite_master table if the SELECT statement on the right-hand side aborts with an error. Ticket 873cae2b6e25b
- Fix the skip-scan optimization so that it works correctly when the OR optimization is used on WITHOUT ROWID tables. Ticket 8fd39115d8f46
- Fix the sqlite3_memory_used() and sqlite3_memory_highwater() interfaces so that they actually do provide a 64-bit answer.

### DIFF
--- a/Data/SQLite/src/sqlite3.h
+++ b/Data/SQLite/src/sqlite3.h
@@ -23,7 +23,7 @@
 **
 ** The official C-language API documentation for SQLite is derived
 ** from comments in this file.  This file is the authoritative source
-** on how SQLite interfaces are suppose to operate.
+** on how SQLite interfaces are supposed to operate.
 **
 ** The name of this file under configuration management is "sqlite.h.in".
 ** The makefile makes some minor changes to this file (such as inserting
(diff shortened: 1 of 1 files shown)


mkrivos:
All Poco sqlite testunit passed.

PR #920, opened 2015-08-22
New XmlStream API

mkrivos:
Pocoized library from http://www.codesynthesis.com/projects/libstudxml/ + tests.
Any ideas to improve this patch are welcomed.


mkrivos:
Compiles Ok on linux64 gcc 4.9.3 and win64 vs 2013

mkrivos:
For now, I am done.

obiltschnig:
Need to check with libstudxml authors if it's OK to re-license under Boost instead of MIT license.

obiltschnig:
Got the OK to use the Boost license from the libstudxml author.

obiltschnig:
One other issue: should we really include the serialization part (based on genX). It's API is not much different from the XMLWriter, so I wonder if it makes sense pulling in additional external dependencies (genX), which we'll have to maintain, etc.

mkrivos:
Ok, can I remove Serializer part ?

obiltschnig:
Yes, please remove the serializer part including genX.

mkrivos:
Done

obiltschnig:
Made some more style fixes today. Also, we can get rid of the QName class. It's equivalent to the existing Name class. See #923

mkrivos:
Ok I can look at, but at sanday because I'm on vacation for now.

obiltschnig:
Don't worry, I'll take care of that.

PR #922, opened 2015-08-26
fixed sql lite statement for empty extracts

dimanikulin:
When we call statement execution (SQL lite) for  "PRAGMA incremental_vacuum(1024);" it fails with null std iterator exception. It happens because variable "extracts" is empty and construction "*extracts.begin()" is invalid.

### DIFF
--- a/Data/SQLite/src/SQLiteStatementImpl.cpp
+++ b/Data/SQLite/src/SQLiteStatementImpl.cpp
@@ -274,7 +274,6 @@ std::size_t SQLiteStatementImpl::next()
 		_stepCalled = false;
 		if (_affectedRowCount == POCO_SQLITE_INV_ROW_CNT) _affectedRowCount = 0;
 
-		//_affectedRowCount += (*extracts.begin())->numOfRowsHandled();
 		if (extracts.size())
 			_affectedRowCount += (*extracts.begin())->numOfRowsHandled();
 		else
--- a/Data/SQLite/src/SQLiteStatementImpl.cpp
+++ b/Data/SQLite/src/SQLiteStatementImpl.cpp
@@ -273,7 +273,16 @@ std::size_t SQLiteStatementImpl::next()
 		}
 		_stepCalled = false;
 		if (_affectedRowCount == POCO_SQLITE_INV_ROW_CNT) _affectedRowCount = 0;
-		_affectedRowCount += (*extracts.begin())->numOfRowsHandled();
+
+		//_affectedRowCount += (*extracts.begin())->numOfRowsHandled();
+		if (extracts.size())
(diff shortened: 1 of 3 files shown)


dimanikulin:
I saw there are some checks unsuccessful but I can not understand if they are critical or not

dimanikulin:
Now It passed all the checks, can it be merged?

aleks-f:
could you please add a test case for this?

dimanikulin:
Test case is ready

dimanikulin:
There is travis-ci check unsuccessful, is it critical?

aleks-f:
I did not examine it. Travis often fails for reasons non-related to code. If you can take a look at it, it would be good. Did you compile and run test on linux/mac?

dimanikulin:
Unfortunately I do not have a linux/mac environment. It was compiled and checked on Win only.

aleks-f:
It compiles, but test fails:
```There was 1 failure: 
 1: N7CppUnit10TestCallerI10SQLiteTestEE.testIncrementVacuum
    "pragma_mode==2"
    in "src/SQLiteTest.cpp", line 3461```

dimanikulin:
Yes It fails if "dummy.db" is present at moment of "testIncrementVacuum" execution. It must be executed on empty DB without any table created.

aleks-f:
that should be handled by test - delete and recreate it.

dimanikulin:
I tried that way (removing DB file and creating), but DB file for some reason was blocked by "TestSuite.exe" process and from source code it was not able to delete it.

dimanikulin:
Can these changes be merged to poco-1.6.1-release?

zosrothko:
Hi
@dimanikulin This IncrementVacuum test is always failing on Ubuntu and blocks the CI on travis. It also fails on Cygwin. Thus you can install Cygwin on your platform and check why it is failing and fix it on Cygwin. 
@aleks-f Meantime, it would be nice if this test be commented out from SQLiteTest so that Travis wo'nt fail by the end.

dimanikulin:
@zosrothko I've already tried to fix it but without success. 

This test requires a new DB file as "Increment AutoVacuum" can be set only if no table is created.
So as solution I tried to delete DB file created by previous tests and to create new one.
I got "File Access Error".
I thought It is because session was not closed before file is removed but It is not.
The DB file was blocked by testSuite.exe process even I used my own DB file name for this test. 
And I was able to delete the file only after testSuite.exe exited.
I tried to delete db file inside of other process and it was successful.
I was not able to find a place where testSuite.exe takes exclusive ownership of any DB files.

As a workarounds we can use:
- setting "Increment AutoVacuum" on first test. But it can impact on that test.
- to clear DB file at time manually and run this test manually.

So let it be just commented :-)

Issue #912, opened 2015-08-19
Generating custom open SSL cipher keys


benoitbleuze:
Hi all,

I have a problem with an assumption made in the Cipher Key implementation. First a disclaimer, I am far from an open SSL expert, and that was one among many reason why our team started using poco, which is a great piece of software.

If one looks at [CipherKeyImpl.cpp](https://github.com/pocoproject/poco/blob/develop/Crypto/src/CipherKeyImpl.cpp)
one would find that the key and Iv are created by using user-provided password, salt (if not empty), and number of iteration for the generation.

This is fine, but insufficient for my use. Open SSL can accept more arguments, such as different digest algorithm. I must read encrypted data that use keys that employ a SHA1 digest. 

Here is the part of the code "tailored" for MD5:

```cpp
void CipherKeyImpl::generateKey(
	const std::string& password,
	const std::string& salt,
	int iterationCount)
{

// ...

// Now create the key and IV, using the MD5 digest algorithm.
	int keySize = EVP_BytesToKey(
		_pCipher,
		EVP_md5(),
		(salt.empty() ? 0 : saltBytes),
		reinterpret_cast<const unsigned char*>(password.data()),
		static_cast<int>(password.size()),
		iterationCount,
		keyBytes,
		ivBytes);
```

My obvious solution would be : 

* Change the interface of this function, and all the constructors needed to add a new parameter being `digest`. 
* Choose a data structure to identify digests (enum? string, the same way the cipher names are encrypted?)
* still keep a default to MD5 to not break backward compatibility

Is it a change the poco team would be interested upstream? if yes do you have any suggestion to improve, or fix issues I would not have found when thinking about this improvement?

benoitbleuze:
I will bump my own post, by adding I have seen there is a constructor in `CipherKey` that takes a `key` and an `iv` vectors, and one could use this constructor to use any digest they want, but a very good feature of the poco library is to abstract oneself of any direct linking dependency on open-ssl. 


Also since last time, I found `EVP_get_digestbyname()` which is the best candidate to use the open ssl sanctioned list of digests, without creating a list of digests in poco itself. I would use these strings (similarly to the cipher name) to set the digest, raising the same exception as the one raised for unknown cipher names.

Is anyone in the core team interested in receiving such a patch?

obiltschnig:
Yes, that sounds useful. Pull request welcome ;-)

benoitbleuze:
Patch merged upstream, I guess we can close this issue then :-)

PR #925, opened 2015-08-31
Crypto: Add any digest to  key generation.

benoitbleuze:
See #912. This pull request adds the possibility to add a custom digest for key generation in symmetric ciphers.

A question though: how is memory management done? The _pCipher pointer is not directly deleted, I would assume because it is delivered by the function EVP_get_cipherbyname(), and this pointer would be deleted by the OpenSSLInitializer::uninitialize() method? I have to admit this is a bit obscure to me, why not use RAII in all the cipher classes?

### DIFF
--- a/Crypto/include/Poco/Crypto/CipherKey.h
+++ b/Crypto/include/Poco/Crypto/CipherKey.h
@@ -47,6 +47,16 @@ class Crypto_API CipherKey
 	///     std::string salt("asdff8723lasdf(**923412");
 	///     CipherKey key("aes-256", password, salt);
 	///
+	/// You may also control the digest and the number of iterations used to generate the key
+	/// by specifying the specific values. Here we create a key with the same data as before,
+	/// except that we use 100 iterations instead of DEFAULT_ITERATION_COUNT, and sha1 instead of
+	/// the default md5:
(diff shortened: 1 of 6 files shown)


aleks-f:
I'm not sure (did not look deeper into it), but there are EVP*cleanup() and *destroy() API calls. It is possible that we are leaking here. Also, it would be good to have some tests for this.

benoitbleuze:
For the tests, what do you prefer: I could either implement just a password digest test similar to `void CryptoTest::testPassword()`, which would hash the password with different digests, and stop there.

Or do you wish to also add another one similar to `CryptoTest::testEncryptDecryptWithSalt()`, which would encrypt and decrypt with a password, salt and different digest?

For the memory management, I will have a look at the openssl API, but it is really not a library I am used to, and it is really not the easiest documentation to read...

benoitbleuze:
I did run the tests with Valgrind, and there is no memory leak, the calls to `Poco::Crypto::initializeCrypto();` and `Poco::Crypto::uninitializeCrypto();` in _driver.cpp_ in the test suite takes care of everything. So as long as people use these functions, all is green in terms of memory management.

PR #929, opened 2015-09-01
fix: CMake compilation for QNX 6.6

csabakeszegh:
QNX cmake var should set to 1, on command line, or in a toolchain file
do not set _XOPEN_SOURCE macro when building for QNX
set SYSLIBS cmake var (see: build/config/QNX:SYSFLAG)
tested: Ubuntu 14.04, QNX 6.6.0, cmake 3.2.3


bachp:
From my point of view if it breaks none of the other builds this can be merged.

PR #938, opened 2015-09-15
fix: cmake: set public compiler flags on library targets

csabakeszegh:
POCO_STATIC: has to be set when using poco as static library
POCO_NO_AUTOMATIC_LIBS: CMake config module will find the correct libraries.
  Don't need to guess in headers.


bachp:
Good catch.
The exported targets are not very well tested.
Exporting the definitions makes sense.

@aleks-f good to merge

PR #943, opened 2015-09-19
Proposed changes for fixing NetSSL_OpenSSL and adding portability function in Sharedmemory

zosrothko:
Hi

here follow proposed changes for fixing NetSSL_OpenSSL compilation of TLSv1_1 and TLSv1_2.

The SahredLibrary::getOSName(const std::string& name) bring a better portability in SharedMemory for naming the real OS dependent library. For exemple, getOSName("FOO") will return

on Linux/Unix:   libFOO.so
on Cygwin        cygFOO.dll
on Windows      FOO.dll 

so that loading(getOSName("FOO")) will work on all platforms.

May be it would be also interesting to add an paramter asto the suffix function as suffix( const char* version = NULL) so that the version numbering be handled properly.

### DIFF
--- a/Foundation/include/Poco/SharedLibrary.h
+++ b/Foundation/include/Poco/SharedLibrary.h
@@ -121,7 +121,7 @@ public:
 	static std::string prefix();
 		/// Returns the platform-specific filename prefix
 		/// for shared libraries.
-		/// Most platforms would return an empty string, but
+		/// Most platforms would return "lib" as prefix, while
 	    /// on Cygwin, the "cyg" prefix will be returned.
 		
 	static std::string suffix();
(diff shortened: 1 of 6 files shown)


zosrothko:
Yes, I have rollbacked to the inital version and fix the  guarded TLSv1_1/TLSv1_2 code in Context.cpp

zosrothko:
May be it is ok for you to change the status of the Cygwin port from Experimental to Operational??

aleks-f:
I'm not sure I'm following - this pull includes poco embedded headers and I can't see that changed. As for change to operational - where, in the documentation?

zosrothko:
Sorry to be vague about, but I was talking about the status of the Cygwin port on the Wiki page  (by the way, I have updated it)

PR #945, opened 2015-09-19
Refactory BinaryReader and BinaryWriter

aaron0x:
1. move flipBytes(float/double) to ByteOrder.
2. remove redundant code from BinaryWriter and BinaryReader.


obiltschnig:
Thanks. One issue: you should not implement 

BinaryWriter& BinaryWriter::operator << (const std::string& value)

using:

BinaryWriter& BinaryWriter::operator << (const char* value)

to avoid having to do a strlen on std::string content. Or factor out another method taking a const char* and std::size_t length as argument.

aaron0x:
Good point!

Issue #944, opened 2015-09-19
CMake build produces libPocoFoundation.1.6.1.dylib (or .so)


mikedld:
Autoconf-based build only produces libPocoFoundation.dylib and libPocoFoundation.31.dylib (and their debug counterparts with "d" suffix) on Mac (or .so on other *NIX systems), but CMake-based build also has libPocoFoundation.1.6.1.dylib with former two being symlinked to it. This discrepancy was introduced by pull request #574 (commit b69fafb) and is about Foundation library only. The issue is that when one wants to use `$<TARGET_FILE:Poco::Foundation>` (e.g. to copy the file for distribution), path to libPocoFoundation.1.6.1.dylib is returned, but linking is being made against the libPocoFoundation.31.dylib.

bachp:
This sounds like a Mac only Problem, and at the moment I'm not really equipped to debug this.

By saying it is about the Foundation library only, do you mean the other libraries like PocoUtil is behaving differently?

Could you provide me the content of the generated `PocoFoundationTargets.cmake` file and another generated file like `PocoUtil.cmake` that is behaving as expected?

mikedld:
All the other libraries are fine (read: have only libPocoFoundation.dylib and libPocoFoundation.31.dylib). I don't believe it's a Mac-only issue, but I didn't try reproducing it on other *NIX systems.

Here's the info you request: https://gist.github.com/mikedld/b88a9d351ed04d2142b3 (note the different is actually in `Poco*Targets-<config>.cmake` files).

mikedld:
As hinted in the comment to the aforementioned commit, changing `VERSION ${PROJECT_VERSION}` back to `VERSION ${SHARED_LIBRARY_VERSION}` fixes it for me.

PR #946, opened 2015-09-20
Change Foundation library version back to SHARED_LIBRARY_VERSION

mikedld:
Foundation is the only library using PROJECT_VERSION as library version, the change accidentally introduced by commit b69fafb84c. This causes difference in library being linked to (libPocoFoundation.31.dylib on Mac) and library returned by `$<TARGET_FILE:Poco::Foundation>` for imported target (libPocoFoundation.1.6.1.dylib), causing inconvenience. Also, this differs from autoconf-based builds where only "31" (soversion) library is being created and not "1.6.1" (project version) one.

This closes issue #944.


mikedld:
Just wondering, will this change make it into 1.6.2? If not, what's the next stable version number it'll be in, 1.6.3 or 1.7?

bachp:
I would say this should be included in 1.6.2, as it fixes an obvious error.
@obiltschnig What you think? 1.6.2 is not out as far as I know?

obiltschnig:
I'm okay with that. Could you merge/cherry-pick the changes into the 1.6.2 branch?

bachp:
@obiltschnig I don't have commit access. Could you do that?

PR #975, opened 2015-10-09
Odbc sybase

kostya-lnk-ms:
Fixed null pointer exception when a BLOB is fetched.
Fixed CLOB/BLOB multirow fetching (which was broken and worked for 1 row query results only)
Added more exhaustive BLOB/CLOB testing


kostya-lnk-ms:
Looks like build system malfunction - it can't download mongo and some internal error from travis.

aleks-f:
@kostya-lnk-ms : before I merge, can you check why appveyor fails?

kostya-lnk-ms:
Looks like some license problems - it throws Internal Server error 500 and says something about missing license - I take it as it has nothing to do with my github account (which license?)
https://ci.appveyor.com/project/obiltschnig/poco/build/542

PR #980, opened 2015-10-12
Multiplatform make support for PostgreSQL

intensifi:


aleks-f:
I guess travis and appveyor builds will have to be updated to install libpq prior to building

PR #998, opened 2015-10-27
Always return non-const object from AutoPtr/SharedPtr

mikedld:
Returning pointer or reference to const object from const AutoPtr/SharedPtr methods is incorrect: constant smart ptr should mean constant raw pointer and reference counter (which are its members), but not constant pointee (which is not the member). This change also brings Poco smart ptrs in line with std:: smart ptr classes which don't make pointees transitively constant either.

Another use-case permitted with this change: it's now possible to declare AutoPtr/SharedPtr const on stack or as member variable, to prevent the pointer change, and still be able to call non-const methods on pointee.

### DIFF
--- a/Foundation/include/Poco/AutoPtr.h
+++ b/Foundation/include/Poco/AutoPtr.h
@@ -182,7 +182,7 @@ public:
 		return AutoPtr<Other>(pOther, true);
 	}
 
-	C* operator -> ()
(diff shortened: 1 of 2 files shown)


aleks-f:
@obiltschnig: are you good with this? I think we can merge it

Issue #1009, opened 2015-11-02
Logger API should take AutoPtr instead of plain pointer


DEvil0000:
In all the examples for Logging you use AutoPtr to store the channels. This makes perfect sense. But the Logger API only takes plain pointers and those are not counted with AutoPtr.
So the examples only work if you store the Channels on the stack where you do not remove them (like main function for example), if you store them differently or randomly.
I was getting a segfault because of this in log(Message&).
Consider the following code:

> void configurePocoLogger()
> {
>   AutoPtr<NullChannel> pNULL(new NullChannel);
>   AutoPtr<SimpleFileChannel> pFile(new SimpleFileChannel);
>   pFile->setProperty("path", "/tmp/beforeValid.log"); //"/tmp/beforeValid.log"
>   pFile->setProperty("rotation", "never");
>   Poco::Logger::root().setChannel(pFile);
>   Poco::Logger::root().setLevel(Poco::Message::PRIO_TRACE);
> }
> int main(int argc, char*\* argv)
> {
>     configurePocoLogger();
>     Logger& logger = Logger::root();
>     logger.warning("Testing string function");
>     return 0;
> }

After configurePocoLogger() is finished AutoPtr will remove the object because a plain ptr is not counted.
Actually the code above somehow works with gcc - the code actually failing was like the following what is obviously wrong but should be the same in my eyes:

> void configurePocoLogger()
> {
>   AutoPtr<NullChannel> pNULL(new NullChannel);
>   Channel\* pCh = pNULL;
>   if (true)
>   {
>       AutoPtr<SimpleFileChannel> pFile(new SimpleFileChannel);
>       pFile->setProperty("path", "/tmp/beforeValid.log"); //"/tmp/beforeValid.log"
>       pFile->setProperty("rotation", "never");
>       pCh = pFile;
>   }
>   Poco::Logger::root().setChannel(pCh);
>   Poco::Logger::root().setLevel(Poco::Message::PRIO_TRACE);
> }
> int main(int argc, char*\* argv)
> {
>     configurePocoLogger();
>     Logger& logger = Logger::root();
>     logger.warning("Testing string function");
>     return 0;
> }

I think for the first code the compiler somehow detects that the stack variable is still used so I am not 100% sure but I think it at least could go wrong.

DEvil0000:
may be related to #292

aleks-f:
[...]

PR #1010, opened 2015-11-02
Logger & Channel classes now using AutoPtr<Channel> instead of Channe…

DEvil0000:
…l\* (API improvement)
ticket #1009


jmansion:
Surely we should not expose autoptr in an interface?

unique_ptr, maybe.

obiltschnig:
Poco::AutoPtr is in no way related to std::auto_ptr, despite the unfortunate similar name.

aleks-f:
https://github.com/pocoproject/poco/pull/2189

Issue #685, opened 2015-01-23
poco logger  how to record the source code line ?


lionm117:
hey ,guys , when i try to use logger to record some info into the log file ,  i wanna keep the code line ,and source file name  .  but  the result was never what i had expected . 
was it the way i used wrong ? or  it was a bug ? 
if it was not a bug , how could i  keep down  the souce code line number automatic  ?　
what do you think about it ?  

void Logger_test()
{
	AutoPtr<FileChannel> pchannel(new FileChannel);
	pchannel->setProperty("path", "sample file.log");
	pchannel->setProperty("rotation", "2000 K");
	pchannel->setProperty("archive", "timestamp");

	AutoPtr<PatternFormatter>  ptnfmt(new PatternFormatter("%L%H:%M:%S-%u-%U : %t"));
	AutoPtr<FormattingChannel> fmtchnl(new FormattingChannel(ptnfmt, pchannel));
	Logger::root().setChannel(fmtchnl);
	Logger& logger = Logger::get("TestLogger"); // inherits root channel 
	static string filename = "E:logger_poco.cpp";
	for (int i = 0; i < 1000; ++i)
	{
		logger.information("Testing SimpleFileChannel" ,filename , 60);
		Sleep(1500);
	}
}

int _tmain(int argc, _TCHAR* argv[])
{
	Logger_test();
	cout<<"this is the end"<<endl;
	int kk;
	cin >> kk;
	return 0;
}

expected  result:  " 16:55:17-60- E:logger_poco.cpp : Testing SimpleFileChannel "

result :  " 16:55:17-0- : Testing SimpleFileChannel "

bachp:
Could the [poco_* convenience macros](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Logger.h#L520) be what you are looking for?

DEvil0000:
For the normal printf like logger API this is possible but for LogStream I do not see a chance to get file and line into it.
1) I would suggest a flag in the Logger constructor/configuration if it should be included by default.
2) I would suggest to extend the following LogStream API calls with LINE/FILE parameters
priority()
error()
warning()
.....

PR #1011, opened 2015-11-03
improving LogStream API to make it possible setting source file & line

DEvil0000:
see comment in #685


aleks-f:
@DEvil0000 
- why directly in LogStream and not delegating this to Logger?
-we already have [NDC](https://github.com/pocoproject/poco/blob/poco-1.7.9/Foundation/include/Poco/NestedDiagnosticContext.h#L73) that can contain this kind of information, perhaps better to add some NDC support to the logging framework

In any case, this pull is out of sync, so I'll close it; feel free to propose something along the above lines and we'll look into it

Issue #894, opened 2015-08-03
Error when linking 32-bit on a 64-bit machine


avishaib:
When compiling 32-bit poco on a 64-bit machine (tested on Debian 7), link fails.
1. The configure script is set with --cflags=-m32
2. OSARCH_64BITS is set to 0
3. POCO_TARGET_OSARCH is set to i386

The compiler creates 32-bit object files, but the -m32 flag is missing from the g++ command when linking the libraries.

FIX:
It can be fixed by adding -m32 to SHLIBFLAGS in build/rules/global in case OSARCH_64BITS==0.

Tigzy:
Hey there, it's 2018 and I'm still having the issue. I'm unable to link my 32 bits compiled object from a 64 bits machine. Is there a reason that isn't released in official package yet?

Tigzy:
The latest, 1.9.0 doesn't have the fix

Tigzy:
Up?

Tigzy:
using: 
`./configure --shared --cflags='-g -O0 -w -m32' --omit=Data/MySQL,Data/ODBC`
I've tried to define 
OSARCH_64BITS=0
POCO_TARGET_OSARCH='i386'
still having build errors at link time (output not compatible with i386:x86-64)

zosrothko:
Could you add the file build/config/Linux32-gcc with the content below and use the following command
`./configure --config=Linux32-gcc --shared --cflags='-g -O0 -w -m32' --omit=Data/MySQL,Data/ODBC`

```
#
# Linux
#
# Make settings for crossbuilding Poco Linux x86 on a Linux x86_64 OS.
#
#

include $(POCO_BASE)/build/config/Linux

#
# Crossbuild Settings
#
POCO_TARGET_OSNAME  = Linux
POCO_TARGET_OSARCH  = x86

#
# Compiler and Linker Flags
#
CFLAGS64        += -m32
CXXFLAGS64      += -m32
SHLIBFLAGS64    += -m32
LINKFLAGS64     += -m32
```

Tigzy:
testing, be back soon.

zosrothko:
You can add those options on the configure command
```
  --include-path=<path>
    Add search path for header files.

  --library-path=<path>
    Add search path for library files.
```

Tigzy:
Yes but these will be static. I have no way to specify both 32 and 64 bits library-paths. If I do this, I'm pretty sure it'll try to link to the wrong version of the lib.

[...]

PR #1018, opened 2015-11-06
Fully support cross-compiling with Linux GNU build system using OSARCH_64BITS=[0,1]

kblaschke:
Building POCO on a Linux host with an architecture different from the host system's (e. g. using `--cflags="-m32"`), dynamic linking of the POCO libraries fails. This patch adds the proper compiler and linker flags to the Linux configuration (including auto.detection of the host's arch if not specified), and passes `LINKFLAGS` to the dynamic linker command line.

Note: The build/rules/global script always adds either `*FLAGS32` or `*FLAGS64` to the flags list, and defaults to the 32-bit variant if `OSARCH_64BITS` is undefined. I left the file unchanged to avoid possible problems with other OS configurations.

This patch will fix issue #894

### DIFF
--- a/build/config/Linux
+++ b/build/config/Linux
@@ -39,14 +39,14 @@ SHAREDLIBLINKEXT = .so
 # Compiler and Linker Flags
 #
 CFLAGS          = 
-CFLAGS32        =
-CFLAGS64        =
+CFLAGS32        = -m32
+CFLAGS64        = -m64
 CXXFLAGS        = -Wall -Wno-sign-compare
-CXXFLAGS32      =
-CXXFLAGS64      =
+CXXFLAGS32      = -m32
+CXXFLAGS64      = -m64
 LINKFLAGS       =
-LINKFLAGS32     =
-LINKFLAGS64     =
+LINKFLAGS32     = -m32
+LINKFLAGS64     = -m64
 STATICOPT_CC    =
 STATICOPT_CXX   =
 STATICOPT_LINK  = -static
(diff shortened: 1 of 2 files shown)


RangelReale:
This patch breaks static library building, because it passes -m32 / -m64 to "ar", which doesn't recognize this parameter.

zosrothko:
The -m32 and -m64 options are targeting generation on an 32 bit arch and 64 arch respectively. Since those options are used at generation time, they should be added IMO to CFLAGS32/64 and CXXFLAGS32/64 only .

kblaschke:
The (GNU) linker also needs the architecture parameter to determine the libraries which need to be linked. If not given, the linker will use the first library found in it's search path (mostly the native system arch) and try to use it to link the compiled object files, which will naturally fail if these were generated with a different arch. Passind either CFLAGS or CXXFLAGS to the linker will mostly work, but the linker doesn't accept some compiler arguments and vice versa.

There should be four separate flag lists, each with 32/64 suffixes:
- CFLAGS for the C ompiler
- CXXFLAGS for the C++ compiler
- LDFLAGS for the dynamic linker
- LINKFLAGS for the static linker/archiver

aleks-f:
ok, can you submit a pull fixing that?

zosrothko:
Its already in the branch develop except that the -m32/m64 options should be removed from LIBFLAGS32/64.

SHLIBFLAGS32/64 goes for linking a shared library used by rule/lib
LINKFLAGS32 goes for linking a executable used by rule/exec

One should add also DYLIBFLAGS32/64 for building a dynamicaly loadable lib.

PR #1027, opened 2015-11-11
Only escape string values in Poco::JSON::Stringifier::stringify()

kblaschke:
As the Poco::JSON implementation uses Dynamic::Var, it allows to create specialized implementations of VarHolder to place other types somewhere into a JSON object. This allows for efficient storage and serialization, since there is no need to first convert a specialized class into a JSON::Object structure and then serialize it. Instead, the class can implement the `convert<std::string>()`  method to build a JSON string out of the class members.

This patch should not change the behaviour of Stringifier while only using the Poco::JSON classes.


aleks-f:
Dates and Times should be in quotes as well.

aleks-f:
[fixed](https://github.com/pocoproject/poco/commit/0286d245abc1acfce937dd8c7d69488a3e6c9382)

PR #1029, opened 2015-11-12
tvOS / WatchOS bitcode enabled

danoli3:
Required by default for the platforms.

Added the flags to correctly build the Poco libraries for production with Bitcode.

> For watchOS and tvOS apps, bitcode is required. If you provide bitcode, all apps and frameworks in the app bundle (all targets in the project) need to include bitcode.


danoli3:
This is good! However I just tested x86_64 and it was missing it... Noticed OSFLAGS was overloaded! PR'ing one for the simulators !

PR #1030, opened 2015-11-12
tvOS / WatchOS bitcode enabled for simulators

danoli3:
Required by default for the platforms.

Added the flags to correctly build the Poco libraries for production with Bitcode.

> For watchOS and tvOS apps, bitcode is required. If you provide bitcode, all apps and frameworks in the app bundle (all targets in the project) need to include bitcode.

--- Related to https://github.com/pocoproject/poco/pull/1029


danoli3:
@obiltschnig Yeah this fixes it for simulators!

PR #1034, opened 2015-11-17
Odbc fixes and improvements

kostya-lnk-ms:
Added support for DB2 XML and DBCLOB column types
Added ability to hint the driver what's underlying null type is, some drivers (like Sybase) are unable to cast int to date etc.
Fixed BLOB/CLOB multirow extraction due to improper default value construction
Made connection failures containing an inner ODBC exception with all the diagnostics
Performance fixes for insertion case - if a statement is insert-only it shan't keep asking a driver about parameter info for every row.


kostya-lnk-ms:
Something wrong is going on with the build system

558The command "sudo apt-get install -qq -y g++-4.8" exited with 100.
5590.01s$ export CC="gcc-4.8"

Has anyone got an idea what that could be?

zosrothko:
Yes...one of the problems is coming from a change in the Linux config which now sets the OSARCH_64BITS to 1, leading to executables and libraries ending with 64 while the build/script/runtests.sh is expecting to run testrunner and not testrunner64

The other problem is that Travis is using an old wget which has problem with the cmake.org server with a unrecognized certificate.

I am working to fix those issues but did not get yet a complete run.

kostya-lnk-ms:
I see no build/test errors apart from infrastructure ones.

kostya-lnk-ms:
For some reasons only gcc test run has failed, where some of the SQLite tests produced the following error 
15: N7CppUnit10TestCallerI10SQLiteTestEE.testJSONRowFormatter
    "N4Poco4Data6SQLite17DBLockedExceptionE: The database is locked"
    in "<unknown>", line -1
Looks like the existing test/implementation race condition.

PR #1048, opened 2015-11-27
add SQLiteConfiguration

byteman:


obiltschnig:
I like the idea, but I don't want to create a dependency to Data in the Util library.

aleks-f:
It does not violate an important design principle, we already have dependency on XML and JSON there; if properly ifdef'd out by default, I think this would be fine. Why are tests in Json configuration suite?

obiltschnig:
Well, it does violate my "minimize dependencies" policy ;-) Why not put it into Data and make Data dependent on Util? Although I don't really like that idea either. Maybe factor out the Configuration classes that needs additional dependencies into a separate library?

aleks-f:
I am not happy with current Util dependencies, either; so, separating configuration library would be fine with me; I would definitely not introduce Data dependency on Util

byteman:
" Configuration classes that needs additional dependencies into a separate library" 

i like the idea

Issue #1050, opened 2015-11-28
Poco fails to build with gcc's -Wshadow


vmiklos:
Hi,

Building poco with

```
./configure --cflags="-Werror -Wshadow" && make
```

results currently in:

```
In file included from include/Poco/NumericString.h:24:0,
                 from include/Poco/NumberFormatter.h:24,
                 from include/Poco/ArchiveStrategy.h:27,
                 from src/ArchiveStrategy.cpp:17:
include/Poco/Buffer.h: In constructor ‘Poco::Buffer<T>::Buffer(std::size_t)’:
include/Poco/Buffer.h:41:30: error: declaration of ‘capacity’ shadows a member of 'this' [-Werror=shadow]
  Buffer(std::size_t capacity):
                              ^
cc1plus: all warnings being treated as errors
/home/vmiklos/git/poco/build/rules/compile:53: recipe for target '/home/vmiklos/git/poco/Foundation/obj/Linux/x86_64/debug_shared/ArchiveStrategy.o' failed
make[1]: *** [/home/vmiklos/git/poco/Foundation/obj/Linux/x86_64/debug_shared/ArchiveStrategy.o] Error 1
make[1]: Leaving directory '/home/vmiklos/git/poco/Foundation'
Makefile:116: recipe for target 'Foundation-libexec' failed
make: *** [Foundation-libexec] Error 2
```

This is a problem for me, as I want to use poco in a project that routinely builds with -Wshadow and -Werror. Currently I work this problem around by installing the poco package from the distro package, that way gcc know that the poco includes are provided by the system and it does not trigger the warnings.

Question is: if I invest time in fixing the -Wshadow warnings, would such changes be accepted? If so, I'm happy to go ahead, create fixes and send pull requests, but first I wanted to check if it would be accepted before I spend time on it. :-)

Thanks,

Miklos

aleks-f:
yes, of course, just stick with the coding style and send the pull

vmiklos:
Closing this issue -- my original problem (that I could not build https://github.com/libreoffice/online with -Wshadow due to Poco) is resolved already. Reading https://github.com/pocoproject/poco/pull/1277#issuecomment-226805606, it doesn't make sense to fix all -Wshadow warnings in internal test code as well. (So that a complete -Werror + -Wshadow build would pass.) Thanks for all the patch reviews. :-)

PR #1052, opened 2015-11-30
GH #1050 Foundation: fix gcc -Wshadow warnings

vmiklos:
http://www.appinf.com/download/CppCodingStyleGuide.pdf indicates that "it", "p" and "r" prefixes are welcome, so in a number of cases I could just fix the warnings by applying those prefixes in case the parameter name was the same as the getter member function name (and one of them needed renaming, so I had to rename the parameter name to avoid API/ABI changes).

```
make Foundation-libexec
```

now succeeds for me with -Wshadow + -Werror, using gcc.


aleks-f:
few travis builds are timing out for some reason ...

PR #1055, opened 2015-12-01
Cross build a Poco i686 32 bits on a Linux x86_64 64 bits OS

zosrothko:
Hi

All previous fixes about cross build a Poco i686 32 bits on a Linux x86_64 64 bits OS are wrong.
Setting -m64 on a Linux x86_64 does not help since it is the default target for gc x86_64
Setting -m32 on a Linux i686 does not help neither since it is also the default target for gcc i686

Thus 
1/ the Linux configuration has been rollbacked to its original version which works as a generic Linux configuration on i686 as well as x86_64.
2/ A new configuration Linux32-gcc is created that specifies the -m32 option for all XXX64 flags when running on a Linux x86_64
3/ steps for cross compiling
git clone -b develop https://github.com/pocoproject/poco.git poco
cd poco
mkdir Linux32-gcc
cd Linux32-gcc
../configure --config=Linux32-gcc --everything --omit=NetSSL_OpenSSL


aleks-f:
can you check why travis is failing?

zosrothko:
There is only on jobs that fails among teh 3 that are running tests and end of log states
*** Zip


testSingleFile: 
testDirectory: 
testManipulator: 
testManipulatorDel: 
testManipulatorReplace: 
testSetZipComment: 


No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated


I already signaled to the Travis team to increase the no output timeout from 10 mn to 30 mn.

aleks-f:
the name of this config is misleading, why did you change it from 32 to gcc? couldn't his be properly separated to 32/64 bit as is for clang on Darwin?

zosrothko:
In the build/config, there are

Darwin
Darwin-clang
Darwin-gcc
Darwin32
Darwin32-clang
Darwin32-gcc
and so on

I just try to follow the explicit naming rules used already in config... but you can change it as you want

zosrothko:
got an answer from Travis about the timeout
"No output errors are when something your build is running is not putting out any output so Travis terminates the build, even if it is actually running. To solve this problem, we have a tool called travis_wait that you add in front of whatever line in your .travis.yml that is stalling the build which will output a line to the log to override this error. The default time is 20 minutes but since you need about 30, just add travis_wait 30 (travis_wait accepts a numerical value as a variable to increase the duration) in front of the line in your yml file where you run Zip. "

So could you add to the .travis.yml the command  travis_wait 30 just before running the tests.

aleks-f:
Darwin and Darwin32 are just forwards with tweaks, while here now we have Linux and Linux-gcc, which are both full config files.

zosrothko:
Ok, it is a typo from my side, the Linux32 should be renamed Linux32-gcc as states in the subject. Should I  follow the Darwin forwards and tweaks?

By the way, Darwin32-gcc forwards to Darwin-clang?

aleks-f:
Yes, Darwin convention should be followed. gcc forwarding to clang is obviously a mistake, it's in both 32 and 64-bit

aleks-f:
btw, that thing with time out did not work, btw, can you look at it and see what can be done, I'm not familiar with travis

zosrothko:
I have seen the errored jobs and exchange with Travis.org about. They did not answered yet.

Issue #1039, opened 2015-11-20
race condition in AsyncChannel close/log


DEvil0000:
> void AsyncChannel::close()
> {
>     if (_thread.isRunning())
>     {
>         while (!_queue.empty()) Thread::sleep(100);
>         do 
>         {
>             _queue.wakeUpAll(); 
>         }
>         while (!_thread.tryJoin(100));
>     }
> }

close() -> thread.isRunning() -> true
[thread context change]
some thread calling log().. exec until _queue.enqueue
[thread context change]
AsyncChannel thread finishes...
[thread context change]
close() -> while (!_queue.empty()) -> thread stuck

edit:
this can also cause shutdown issues in the following scenario:
main thread calls exit()
singleton Logger shutdown
other thread logging something
AsyncChannel creates the thread again
AsyncChannel has still some work while main thread destroys stuff with exit()
-> segfault

DEvil0000:
you can also hit this on reconfigure the logger while a different thread is logging:
>#0  0xb7fdbd50 in __kernel_vsyscall ()
#1  0xb7e4c5f6 in nanosleep () at ../sysdeps/unix/syscall-template.S:81
#2  0xb7bc753a in Poco::ThreadImpl::sleepImpl (milliseconds=100) at src/Thread_POSIX.cpp:285
#3  0xb7b252f9 in Poco::Thread::sleep (milliseconds=100) at include/Poco/Thread.h:308
#4  0xb7b24b8e in Poco::AsyncChannel::close (this=0x8142c00) at src/AsyncChannel.cpp:103
#5  0xb7b248d8 in Poco::AsyncChannel::~AsyncChannel (this=0x8142c00, __in_chrg=<optimized out>) at src/AsyncChannel.cpp:64
#6  0xb7b24a45 in Poco::AsyncChannel::~AsyncChannel (this=0x8142c00, __in_chrg=<optimized out>) at src/AsyncChannel.cpp:71
#7  0x0807e5a5 in Poco::RefCountedObject::release (this=0x8142c04)
    at /development/Lib/poco-1.6.1/Foundation/include/Poco/RefCountedObject.h:84
#8  0xb7b65777 in Poco::Logger::setChannel (this=0x8143ce0, pChannel=0x8153a38) at src/Logger.cpp:48
#9  0xb7b6583c in Poco::Logger::setProperty (this=0x8143ce0, name=..., value=...) at src/Logger.cpp:75
[...]

PR #1058, opened 2015-12-02
Fix async channel shutdown race condidion

DEvil0000:
#1039


DEvil0000:
seams like branch is out of sync - sorry for that.

aleks-f:
@obiltschnig is this pull still relevant?

DEvil0000:
I did not test it again with new versions but from looking at the code I would say its still needed. I did just not find time to fix my messed up PR.

PR #1060, opened 2015-12-04
GH #1050 JSON: fix gcc -Wshadow warnings

vmiklos:
````
make JSON-libexec
````

now succeeds for me with -Wshadow + -Werror, using gcc.

### DIFF
--- a/Foundation/include/Poco/Nullable.h
+++ b/Foundation/include/Poco/Nullable.h
@@ -74,9 +74,9 @@ public:
 	{
 	}
 
-	Nullable(const C& value): 
+	Nullable(const C& rValue): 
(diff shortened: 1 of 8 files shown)


vmiklos:
Hmm, it says the test timed out after 50 minutes, I wonder if that's a false negative then.

aleks-f:
There's something about these builds that makes them timeout. The [fix](https://github.com/pocoproject/poco/commit/2266fc19be86f350461981c648e13fbb30455d84) suggested by Travis did not help. @zosrothko is in contact with Travis to resolve

vmiklos:
Let me rebase this patch then, now that develop is green. :-)

PR #1067, opened 2015-12-07
improvment in running tests 

zosrothko:
Hi

This PR is covering  the following points:
1/ For a UNIX_FAMILY platform, use of POCO_BASE for finding out the configuration file of the testsuite of NetSSL_OpenSSL
2/ Add a filter on runtests2.cmd on Windows.
2/ Limit verbosity of msbuild at minimal

It is worthwhile to note that now, one can build Poco in a separate directory and run the tests from $POCO_BASE. This can be applied to multiple parallele branches. Here a sample

git clone https://github.com/pocoproject/poco.git poco
cd poco
git checkout MyBranch
mkdir MyBranch
cd MyBranch
../configure --everything
make -s -j4
cd ..
. env.sh
export POCO_BUILD=`pwd`/MyBranch
./build/script/runtests.sh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,6 +120,7 @@ RemotingNG
 XSD
 stage/
 releases/
+Zip/testsuite/data*.bin
 
 
 # openssl binaries #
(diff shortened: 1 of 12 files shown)


aleks-f:
Can you also add those options to [PowerShell script](https://github.com/pocoproject/poco/blob/develop/buildwin.ps1). The idea of PowerShell was to have a proper build script and get rid of the awful batch, but it keeps living.

zosrothko:
Added limitation of verbosity to minimal and nologo to buildwin.ps1. Back port of some CMake fixes from poco-1.6.2 fixes.

Issue #1062, opened 2015-12-04
MongoDB Document toString with element is date object


manthanrtilva:
Hi,

When i run below code.

	Poco::MongoDB::Connection connection("localhost",27017);
	Poco::MongoDB::Database db("a");
	Poco::MongoDB::ResponseMessage response;
	Poco::SharedPtr<Poco::MongoDB::InsertRequest> insertPlayerRequest = db.createInsertRequest("b");
	Poco::Timestamp t;
	insertPlayerRequest->addNewDocument()
		.add("lastname", "Valdes")
		.add("firstname", "Victor")
		.add("time", t);
	connection.sendRequest(*insertPlayerRequest);
	
    Poco::SharedPtr<Poco::MongoDB::QueryRequest> queryPtr = db.createQueryRequest("b");
    queryPtr->setNumberToReturn(1);
    connection.sendRequest(*queryPtr, response);
    Poco::Dynamic::Var result;
    if ( response.hasDocuments() )
    {
    	std::cout<<response.documents()[0]->toString()<<std::endl;
    }

its printing output like below.

{"_id":56619736ce47c3e150903db6,"firstname":"Victor","lastname":"Valdes","time":2015-12-04T13:37:58.161000Z}

I think, as we using toString method time element should be surrounded by double quote.

Let me know if my understanding is wrong or there is any issue in library.

manthanrtilva:
You just need to update in Element.h file toString() method of  struct ElementTraits<Timestamp>.

Correct me if I am wrong? So I can continue with my development.

fbraem:
Correct. Just need some time to see if this change doesn't break anything.

aleks-f:
@fbraem you can use [Var::isDateTime()](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Dynamic/Var.h#L505-L511),  those were introduced for [JSONRowFormatter](https://github.com/pocoproject/poco/blob/develop/Data/src/JSONRowFormatter.cpp#L144)

PR #1047, opened 2015-11-27
AppVeyor: full build with msbuild

zosrothko:
Hi

This PR changes the AppVeyor CI from cmake to msbuild with a complete testing matrix with the following dimensions:
configuration:    (release, debug)
link               :    (shared, static_md, static_mt)
builder         :     (VS2013, VS2015)

Ther is a lot a minor commits just for adjusting the appveyor.yml to the AppVeyor infrastructure as well as typos and beginner's errors (I never used PS before).

runtest2.cmd is commented out because it goes in loop and makes the job fails by timeout. This will be investigated later on.

buildwin.cmd has been adjusted by adding the VERBOSITY parameter and the LOGGER parameter for using the AppVeyor logging that displays the msbuild messages in the MESSAGE panel.

The full CI takes around 6 hours.

Any member of the GitHub/Kampbell organization has access to the AppVeyor jobs for checking the outcome.

PR #1068, opened 2015-12-08
MongoDB enhancements

### LINKED ISSUE #1051: MongoDB Tests on Mac don't fail gracefully
On Mac (Yosemite), with 1.6.1, I'm seeing the `MongoDBTest` throw an `InvalidArgumentException` when the `suite()` method creates the *second* `MongoDBTest` object. The first object attempts to connect to `localhost:27017` and fails, because I don't have a MongoDB running. This seems to leave the underlying `_mongo` `MongoDB::Connection` (specifically the socket I think) in a weird state, which - because it is global - causes the second constructor call to fail with the `InvalidArgumentException`. If `_mongo.disconnect()` is called upon failure, the socket seems to get itself into a nice state and the creation of the other objects fails gracefully, and all of the tests are skipped as expected (with one exception - see below). It seems very strange to me to have one global `MongoDB::Connection` object, which is modified by each of the constructors, only to then be used by the tests one at a time (but after the other test constructors have modified it). Surely the `MongoDBTest` object owning the connection and the `_connected` bool would make more sense?

For some reason this failure doesn't occur on Linux.

On a related note, the `ConnectionPool` test doesn't catch the `ConnectionRefusedException` (and doesn't skip itself if `_connected` is false), so that still fails the tests.

fbraem:
Fixes #1051 and #1062

### DIFF
--- a/MongoDB/include/Poco/MongoDB/Element.h
+++ b/MongoDB/include/Poco/MongoDB/Element.h
@@ -27,6 +27,7 @@
 #include "Poco/Nullable.h"
 #include "Poco/NumberFormatter.h"
 #include "Poco/DateTimeFormatter.h"
+#include "Poco/UTF8String.h"
 #include "Poco/MongoDB/MongoDB.h"
 #include "Poco/MongoDB/BSONReader.h"
 #include "Poco/MongoDB/BSONWriter.h"
@@ -79,7 +80,7 @@ inline std::string Element::name() const
 typedef std::list<Element::Ptr> ElementSet;
 
 
-template<typename T> 
+template<typename T>
 struct ElementTraits
 {
 };
@@ -109,51 +110,11 @@ struct ElementTraits<std::string>
 
 	static std::string toString(const std::string& value, int indent = 0)
 	{
-		std::ostringstream oss;
-
-		oss << '"';
-
-		for(std::string::const_iterator it = value.begin(); it != value.end(); ++it)
-		{
-			switch (*it)
-			{
-			case '"':
-				oss << "\\\"";
-				break;
-			case '\\':
-				oss << "\\\\";
-				break;
-			case '\b':
-				oss << "\\b";
-				break;
-			case '\f':
-				oss << "\\f";
-				break;
-			case '\n':
-				oss << "\\n";
-				break;
-			case '\r':
-				oss << "\\r";
-				break;
-			case '\t':
-				oss << "\\t";
-				break;
-			default:
-				{
-					if ( *it > 0 && *it <= 0x1F )
-					{
-						oss << "\\u" << std::hex << std::uppercase << std::setfill('0') << std::setw(4) << static_cast<int>(*it);
-					}
-					else
-					{
-						oss << *it;
-					}
-					break;
-				}
-			}
-		}
-		oss << '"';
-		return oss.str();
+		std::string result;
+		result.append(1, '"');
+		result.append(UTF8::escape(value));
+		result.append(1, '"');
+		return result;
 	}
 };
 
@@ -231,7 +192,11 @@ struct ElementTraits<Timestamp>
 
 	static std::string toString(const Timestamp& value, int indent = 0)
 	{
-		return DateTimeFormatter::format(value, "%Y-%m-%dT%H:%M:%s%z");
+		std::string result;
+		result.append(1, '"');
+		result.append(DateTimeFormatter::format(value, "%Y-%m-%dT%H:%M:%s%z"));
+		result.append(1, '"');
+		return result;
 	}
 };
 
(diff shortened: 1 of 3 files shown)


aleks-f:
@bufferoverflow there's AppVeyor cmake install error: 
```
cinst cmake
Installing the following packages:
cmake
By installing you accept licenses for the packages.
cmake not installed. An error occurred during installation:
 The remote server returned an error: (500) Internal Server Error.
The install of cmake was NOT successful.
cmake not installed. An error occurred during installation:
 The remote server returned an error: (500) Internal Server Error.
```

aleks-f:
@zosrothko : are you familiar with AppVeyor?

zosrothko:
No, but I am learning... as for Travis and PowerShell... I am working full time 5d/week on Poco since 2 months...

AppVeyor is working fine now for cmake... Can you have a look on hte Poco build #113,  running an updated version of appveyor.yml ...

aleks-f:
for #113 see with @fbraem. obviously failed tests are not something we want

zosrothko:
Sure but there is an issue with Poco/CMake/CTest configuration

In release mode, 2 of the 3 failing tests are the testProxy from NetSSL
: There were 2 errors: 
7:  1: class CppUnit::TestCaller<class HTTPSClientSessionTest>.testProxy
7:     "class Poco::Net::HTTPException: HTTP Exception"
7:     in "<unknown>", line -1
7:  2: class CppUnit::TestCaller<class HTTPSStreamFactoryTest>.testProxy
7:     "class Poco::Net::HTTPException: HTTP Exception"
7:     in "<unknown>", line -1

The point is that there is no way to exclude those 2 tests from the run. CTest can exclude only component, but not some tests from one component as on Travis.

The 3rd test failing in in Foundation
There was 1 failure: 
1:  1: class CppUnit::TestCaller<class TimedNotificationQueueTest>.testDequeue
1:     "ts1.elapsed() >= 0"
1:     in "C:\projects\poco-olq4c\Foundation\testsuite\src\TimedNotificationQueueTest.cpp", line 97

So I do not see a proper solution except surrounding the testProxy by a try/catch HTTPException displaying the what() when no proxy is avalable.

Note also that for now, none of the Data tests are run.. and also that the current appveyor.yml is not running any test at all... that's why appveyor is always green

aleks-f:
is #1047 adding tests to AppVeyor? In regards to proxy test failing, I don't think it is expected to fail

zosrothko:
NetSSL/testProxy is failing on my local win7 with VS2013 debug & release with the below exception
`testProxy: HTTP Exception: Cannot establish proxy connection: Forbidden`

It is also failing on travis and has been excluded from the travis. See travis/runtets.sh
`export CPPUNIT_IGNORE="\
  N7CppUnit10TestCallerI8PathTestEE.testExpand \
  N7CppUnit10TestCallerI13RawSocketTestEE.testEchoIPv4 \
  N7CppUnit10TestCallerI13RawSocketTestEE.testSendToReceiveFromIPv4 \
  N7CppUnit10TestCallerI14ICMPClientTestEE.testPing \
  N7CppUnit10TestCallerI22HTTPSClientSessionTestEE.testProxy \
  N7CppUnit10TestCallerI22HTTPSStreamFactoryTestEE.testProxy \
  N7CppUnit10TestCallerI19MulticastSocketTestEE.testMulticast \
  N7CppUnit10TestCallerI13NTPClientTestEE.testTimeSync"`

aleks-f:
try to run as administrator

zosrothko:
Same result
Z:\git\poco>build\script\runtests2.cmd NetSSL_OpenSSL Proxy


*** NetSSL_OpenSSL


testProxy: HTTP Exception: Cannot establish proxy connection: Forbidden

BTW, I have no idea if running with administrator UAC on AppVeyor is possible...but IMO, it would be better to fix this testProxy without admin UAC.

aleks-f:
I have never encountered problem with proxy. UAC was needed only for ping because it uses raw sockets.

zosrothko:
@obiltschnig @aleks-f Could you check those 2 testProxy from NetSSL_OpenSSL/testsuite and adapt them at your convenience so that they do not failed when no proxy are defined/activated/running?

`HTTPSClientSessionTest>.testProxy`
`HTTPSStreamFactoryTest>.testProxy`

zosrothko:
Foundation\testsuite\src\TimedNotificationQueueTest.cpp failure has been fixed, thus it remains only the 2 testProxy failures to fix for a complete successful round on AppVeyor for release mode.

aleks-f:
Ok, those are failing just for NetSSL, right?

zosrothko:
Yes only NetSSL

PR #1073, opened 2015-12-10
Don't leak <windows.h> (Poco/UnWindows.h) to library users

RangelReale:
Platform_WIN32.h includes &lt;Poco/Unwindows.h&gt; and in consequence, &lt;windows.h&gt; is included in all source code that uses Poco on Windows, increasing the compilation time.

This patch moves the OS detection (the thing that needs windows.h) to a separate header, that must be included if these are needed (in Poco, only Net/src/NetworkInterface.cpp).

Also, client code that assumed windows.h is always included, must include &lt;Poco/UnWindows.h&gt; explicitly. In my application that was 4 source files only.

With this patch, testing in MSVC with /showIncludes, resulted in 15% less files included.


aleks-f:
I'm fine with moving the OS version detection to separate header, that was only needed for NetworkInterface so far and I think last version needing it was XP or Vista (although the need may arise again in the future so it's not a bad idea to keep it). 

Minimizing the includes is a worthy goal but UnWindows.h was introduced to shield us from accidental use of the evil macros for API calls which have narrow and wide string arguments version, i.e. to force the explicit use of W or A version of API calls (although the list there is incomplete and should be completed for full protection). I'm not sure whether we should remove the global inclusion. @obiltschnig ?

obiltschnig:
I would be in favor of getting rid of including <windows.h> globally. There are a few places, though were we need Windows API definitions in the header, either for class members or for inlined functions. A possible way to solve this is to use the pimpl idiom, but I don't like that for performance reasons (for things like Mutex::lock() and unlock() it makes a noticeable difference being able to inline the methods).
Maybe we could just extract the relevant definitions from windows.h, but that's not a perfect solution either.

RangelReale:
Yes, all platform-dependent files that are not already pimpl, will continue to leak windows.h.
I've looked at converting them to pimpl, but they directly map platform-specific constants to the base class, so it would not be feasible.

In Poco itself very few places needed to be changed after I removed UnWindows.h from the platform file. But of course some classes are still including UnWindows.h from the platform-specific files, so I can't be too sure.

jbulow:
Poco 1.10.1 leaks macros in windows.h to library user. Breaks code for library users.

PR #1089, opened 2015-12-15
minor change to support cmake2.8 builds

guruofquality:
Reduce cmake version requirement to 2.8.12 (ubuntu trusty), and conditionalize export() in PocoMacros to support 2.8 syntax. This supports building recent versions of poco (1.6.2) using the cmake2.8 series, and using the project config files in a client project under cmake 2.8.

Realizing that cmake2.8 is pretty old now, and cmake3.0 offers many great new features, cmake2.8 is still a default for ubuntu trusty (LTS). Supporting cmake2.8 would allow me to use Poco 1.6.2 as a cmake subproject on ubuntu trusty. Currently, I still use poco 1.5.4 because of the cmake compatibility.

Anyway, far from going backwards, the changes to support this are really quite small, and the resulting project config files even work on a client project using cmake2.8 (provided that there is a copy of CMakeFindDependencyMacro.cmake).

Its there is no interest, I certainty understand -- please close the issue. But I wanted to share this patch just in case there was. Looks like cmake 2.8 was mentioned here: https://github.com/pocoproject/poco/wiki/CMake-Status#todo

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 # ENABLE_{COMPONENT}
 # ENABLE_TESTS
 
-cmake_minimum_required(VERSION 3.0.0)
+cmake_minimum_required(VERSION 2.8.12)
 
 project(Poco)
 
--- a/cmake/PocoMacros.cmake
+++ b/cmake/PocoMacros.cmake
@@ -202,10 +202,19 @@ write_basic_package_version_file(
   VERSION ${PROJECT_VERSION}
   COMPATIBILITY AnyNewerVersion
 )
-export(EXPORT "${target_name}Targets"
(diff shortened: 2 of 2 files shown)


RangelReale:
Just to point it out, the upcoming c++11 branch requires CMake 3.1, because of C++11 support.

In my ubuntu, I used [this](http://askubuntu.com/questions/610291/how-to-install-cmake-3-2-on-ubuntu-14-04) ppa to get a newer CMake, which worked flawlessly.

guruofquality:
> Just to point it out, the upcoming c++11 branch requires CMake 3.1, because of C++11 support.
In my ubuntu, I used this ppa to get a newer CMake, which worked flawlessly.

Thanks, when the time comes, that sounds like a good strategy for my older ubuntu users.

PR #1095, opened 2015-12-22
GNU Make/c++11: computing the version of gcc: use predefined constant instead of --version

zosrothko:
Hi

One should use  gcc predefined constants `__GNUC__`,   `__GNUC_MINOR__`  and `__GNUC_PATCHLEVEL__ ` instead of --version... See embedded comments.


aleks-f:
@zosrothko and @RangelReale : please see why this is failing and let me know when it's good to merge

zosrothko:
Jobs are failing for a serious issue on data type in NumberFormatter I am working on...BTW, you can merge this PR cause otherwise the g++ 4.8 version need to compile will never be applied.

Issue #1102, opened 2015-12-24
c++11 branch & gcc 4.6:compilation failed


zosrothko:
Hi
The c++ branch at pocoproject/poco does not compile with gcc 4.6 while it should
** Creating dependency info for src/Format.cpp

/usr/include/c++/4.6/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the upcoming ISO C++ standard, C++0x. This support is currently experimental, and must be enabled with the -std=c++0x or -std=gnu++0x compiler options.

zosrothko:
Fixed by https://github.com/RangelReale/poco/commit/0d03ab22bd00041dc7c0f6e025eaef6f8a4d4050

ARAVINDAARONE:
error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
Thread_simple1.cpp: In function 'int main()':
Thread_simple1.cpp:32:5: error: 'thread' was not declared in this scope
     thread t1(func);
     ^
Thread_simple1.cpp:32:12: error: expected ';' before 't1'
     thread t1(func);
            ^
Thread_simple1.cpp:33:5: error: 't1' was not declared in this scope
     t1.join();


Hello i facing issue while running my thread program in VS Code
can you support me in resolving this issue

matejk:
@ARAVINDAARONE , which version of Poco did you try to compile and which compiler did you use?

Poco 1.13 and newer requires compiler that supports standard C++17.

ARAVINDAARONE:
i use C++ 14 which is mingw compiler

matejk:
How do you build Poco? With CMake?

ARAVINDAARONE:
not with CMake, but with the terminal ctr run

matejk:
Then I assume that you did not specify the c++ standard option. It must be set to `std=c++17`.

ARAVINDAARONE:
These are the cpp_properties.json file looks like,

{
    "configurations": [
        {
            "name": "windows-gcc-x86",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "compilerPath": "C:/Softwares/Mingw_posix_Thread_C++/x64-4.8.1-release-posix-sjlj-rev5/mingw64/bin/gcc.exe",
            "cStandard": "c23",
            "cppStandard": "c++17",
[...]

PR #1104, opened 2015-12-29
C++11/Travis: improving Travis tests coverage for c++11

zosrothko:
Hi

This PR covers the following points:
1/ Fix the NumberFormatter compile error by removing `<cstdint>` from Types.h
2/ Improvment of c++1/c++14 setup according the clang/gcc and the underlying platform
3/ Fix a missing initialization in AtomicCounterTest
4/ All other changes are related to improving Travis for testing the c++11 branch as show here
https://travis-ci.org/Kampbell/poco job 227 branch c++11

Remaining issues to fix
a/ In Foundation
There were 4 failures: 
 1: N7CppUnit10TestCallerI20DirectoryWatcherTestEE.testAdded
    "_events.size() >= 1"
    in "src/DirectoryWatcherTest.cpp", line 61
 2: N7CppUnit10TestCallerI20DirectoryWatcherTestEE.testRemoved
    "_events.size() >= 1"
    in "src/DirectoryWatcherTest.cpp", line 92
 3: N7CppUnit10TestCallerI20DirectoryWatcherTestEE.testModified
    "_events.size() >= 1"
    in "src/DirectoryWatcherTest.cpp", line 124
 4: N7CppUnit10TestCallerI20DirectoryWatcherTestEE.testMoved
    "_events.size() >= 2"
    in "src/DirectoryWatcherTest.cpp", line 159

b/ Linux 4.6 issue #1102
** Creating dependency info for src/Format.cpp

/usr/include/c++/4.6/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the upcoming ISO C++ standard, C++0x. This support is currently experimental, and must be enabled with the -std=c++0x or -std=gnu++0x compiler options.      
    
c/ SIGPIPE
At least 2 tests of Net and NetSSL are returning an exception under ctest, probably when a server is shutdown. To be investigated.


Foundation/include/Poco/Types.h
```diff
@@ -21,30 +21,12 @@
 
 
 #include "Poco/Foundation.h"
-#if defined(POCO_ENABLE_CPP11)
-#include <cstdint>
```
RangelReale: I added <cstdint> because it is a header guaranteed to be available in C++11.
(http://en.cppreference.com/w/cpp/types/integer)
so it would be a cross-platform way to defined the types.

RangelReale:
The Format.h problem can be solved wrapping the <type_traits> include in the C++11 ifdef, like this:
https://github.com/RangelReale/poco/commit/0d03ab22bd00041dc7c0f6e025eaef6f8a4d4050

RangelReale:
For Clang detection, I was thinking of an approach of doing separate files for Clang and AppleClang, as they return completly different values for the same variables.

This is the patch I was preparing:
https://github.com/RangelReale/poco/commit/76b7a20b9945ccf5bc7d622fa14074488b9a4bfd

zosrothko:
While the job on OSX/clang succeed in the pocoproject PR travis run, it fails on mine with

Assertion failed: (ec == 0), function unlock, file /SourceCache/libcxx/libcxx-120/src/mutex.cpp,testActivity: 

It seems that the c++11 runtime needs to be harden...

zosrothko:
The remaining 3 issues are now fixed and Travis is green on all jobs...

 A small remark, it take 10 mns more time to build&test with gcc 4.8 than gcc 4.6 using gnu make -- around 30% more -- while there is only a 2 mns difference using CMake

PR #1106, opened 2015-12-29
Better C++11 compiler detection

RangelReale:
* Use separate C++11 flag detection for AppleClang
* Fix Format.h for non-C++11 compilers
* Use -ge for compiler version comparison

Compiled on OSX Yosemite (AppleClang 700+), and on Linux with GCC 4.6, 4.8, 4.9 and 5.1.
Clang on Linux has incompatibilites with GCC headers on C++11 mode, so couldn't be tested.

### DIFF
--- a/Foundation/include/Poco/Format.h
+++ b/Foundation/include/Poco/Format.h
@@ -23,7 +23,9 @@
 #include "Poco/Foundation.h"
 #include "Poco/Any.h"
 #include <vector>
+#if defined(POCO_ENABLE_CPP11)
 #include <type_traits>
+#endif
 
 
 namespace Poco {
--- a/Foundation/include/Poco/Platform_POSIX.h
+++ b/Foundation/include/Poco/Platform_POSIX.h
@@ -61,8 +61,8 @@
 	#define POCO_ENABLE_CPP11
 #endif
 
-// Enable C++11 support for Clang 3.4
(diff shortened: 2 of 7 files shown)


zosrothko:
For detecting the conformance of a c++ compiler, the standard mandates 

`The C++11 FDIS says in "16.8 Predefined macro names [cpp.predefined]
The name __cplusplus is defined to the value 201103L when compiling a C++ translation unit. (Footnote: It is intended that future versions of this standard will replace the value of this macro with a greater value. Non-conforming com- pilers should use a value with at most five decimal digits.)`

But this does not work neither perfectly among all compilers and compiler versions for c++11 and c++14... It is worthwhile to notice that fixing the wrong value of _cplusplus for gcc 2.7 took 10 years to solve: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=1773

aleks-f:
so, is this supposed to be merged? it's still falling on travis (the other commit by @zosrothko succeeds)

zosrothko:
It is already merged in the #1104

RangelReale:
I think this patch is complementar to the other one.

aleks-f:
there are conflicts now. please resolve and let me know

zosrothko:
it has already  been integrated in the #1104 
@RangelReale : look at the merged pocoproject/poco@c++11 branch and tell what is missing?

RangelReale:
Ah you merged this PR, allright then!

PR #1131, opened 2016-01-20
Make the Token ignore property a dynamic property.

zosrothko:
Hi

This PR make the ignore property for a token a dynamic programmable property. The use case that covers this PR is the following. When parsing a meta language, one needs sometime to parse portion of code of the language itself. In such case, all token should be recognized and passed thru the process without alteration. This includes the Comment token as well as the Whitespace token. This PR make the choice to ignore or not such tokens when parsed by the StreamTokenizer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,7 @@ Thumbs.db
 *.suo
 *.sdf
 *.opensdf
+*.opendb
 *.tlb
 *.tlh
 *.ilk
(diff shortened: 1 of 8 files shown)


zosrothko:
Hi
this Travis job #1592 TEST_NAME="Linux clang 3.4 (make)" started by the the Pull Request #1131 "Make the Token ignore property a dynamic property". failed on the following test with a segfault while others Travis jobs from the same test with Linux clang 3.4(make) are successfuls.

`testMaxKeepAlive: `
`Segmentation fault (core dumped)`

I would like to restart it but does not have the sufficient permission. Could you restart it?

zosrothko:
@obiltschnig Could you consider this PR for the next release?

Issue #860, opened 2015-06-09
Poco::MongoDB connection pool and multithreading


matejk:
Hi,

Based on an example in the test suite for MongoDB we wanted to use connection pool and related MongoDB::PooledConnection in an a multi-threaded application. The concept seems ideal for that since the ObjectPool does all necessary housekeeping of available objects (connections to MongoDB database in this case).

However, strange behaviour was observed because requests in some of the threads received unexpected responses when using this approach with MongoDB 3.

A multi-threaded test case was prepared to reproduce the situation:

It creates new object in the main thread and spawns new thread to query and remove the object. In 6 cases out of 10 either creation of new object or query fails. Instead of receiving the expected "ok" response it gets a result from query which was requested by another thread or similar. Occasionally the test hangs because it expects response which is not delivered on that socket.

The connection is taken from the pool and then released for every request that is executed.

In the [MongoDB documentation](http://docs.mongodb.org/meta-driver/latest/legacy/mongodb-wire-protocol) it is stated that requests can have client-defined requestID attached.

Do we need to use that?

Will MongoDB return result on the same socket where it was requested?

Any help or clarification would be appreciated.

Thanks,

Matej

matejk:
Example code that demonstrates what's described above is here: https://gist.github.com/matejk/622aca94540a7c5d3f95

matejk:
@fbraem: do you know what might be the problem here?

fbraem:
This is something I still need to investigate ...

fbraem:
I've tried to do the same on my machine (which has MongoDB 2.6.10) and I don't see these issues. For some reason I can't get MongoDB 3.2 running, so can you try my test program: https://gist.github.com/fbraem/ee1a2452a4519bb925d9 ?

matejk:
Hi @fbraem. Sorry that it took quite long to respond.

Your test program works fine with MongoDB 3.2.

I compared your code with mine.  I use connections from the pool in a different way as you do. I'll continue digging what the problem is.

matejk:
I discovered the source of the problem from the differences between our examples. In my example the pooled connection is allocated in a function and returned.

MongoDB::PooledConnection does not have move ctor/operator and then dtor (and resource deallocation) is called when returning from the function.

When disabling copy ctor/operator and enabling move ctor/operator my example works also.

I'll provide a pull request for that.

@fbraem @aleks-f : Is there a policy how to use C++11 features like move ctor in Poco? Does it have to be ifdef-ed?

[...]

PR #1145, opened 2016-02-03
MongoDB: disable unwanted release of resources in PooledConnection

matejk:
Using copy ctor or assignment operator could release the connection when not desired.

Disabled copy semantics.

Added move semantics if C++11 functionality is enabled.

Fixes: #860


matejk:
Pull request contains unwanted clutter. Will create new one.

PR #1148, opened 2016-02-06
Some of the samples need to link with PocoJSON

swt2c:


obiltschnig:
Could you please change the pull request so that PocoJSON comes before PocoFoundation in target_libs. Then it will also work with static linking. Thank you!

swt2c:
Done.

obiltschnig:
Thanks!

PR #1151, opened 2016-02-07
bump to sqlite 3.8.2

mkrivos:
All tests passed on gentoo linux64.


mkrivos:
http://sqlite.org/releaselog/3_10_2.html

PR #1154, opened 2016-02-17
Improved support for MinGW-w64

teminian:
Added following #defines which is not defined in MinGW header:
    #define _RC_CHOP 0x00000300
    #define _RC_UP 0x00000200
    #define _RC_DOWN 0x00000100
    #define _RC_NEAR 0x00000000

Use isfinite(), instead of _finite(), for isInfiniteImpl().
Use isnan(), instead of _isnan(), for isNaNImpl().
Use copysign(), instead of _copysign(), for copySignImpl().


jamilservicos:
my solution: https://github.com/jamilservicos/custom-poco/blob/mingw/Foundation/include/Poco/FPEnvironment_WIN32.h

PR #1157, opened 2016-02-20
Fixing a bug in the NetSSL_Win module (Host name verification failed error)

erl987:
This is fixing a bug that occurs when sending e-mails via SMTP with TLS/SSL encryption. See this discussion:

http://pocoproject.org/forum/viewtopic.php?f=12&t=6708


erl987:
This small fix has quite some impact on using e-mail sending with WinSSL. It still did not make it into the release. Are there any plans to fix that also in the release branch?

PR #1162, opened 2016-02-29
Comment out debugging trace in build/rules/global

zosrothko:


zosrothko:
Using POCO_VERBOSE instead of VERBOSE to avoid name clash with an environment variable other than the Poco one.

obiltschnig:
fixed 'manually' in 1.7.0

Issue #1167, opened 2016-03-02
CMake & POCO_UNBUNDLED: expat sources are compiled in libPocoXML


matejk:
Even though POCO_UNBUNDLED is used on Linux (Ubuntu 15.10) to compile Poco on develop branch, the expat sources are compiled and linked in libPocoXML anyway.

Debian lintian reports:

```
E: libpoco-1.7-clavis: embedded-library usr/lib/libPocoXML.so.40: expat
```

The reason for that is in how SRCS variable is set in the CMakeLists.txt.

I will fix it and prepare a pull request.

PR #1168, opened 2016-03-02
Compile and link bundled expat only if POCO_UNBUNDLED is not set.

matejk:
Bundled expat sources and headers moved to a subdirectory and are referred to only when compiling with unset POCO_UNBUNDLED.

Fixes: #1167


matejk:
Moving files to subdir breaks Makefile build (dependencies).

PR #1169, opened 2016-03-02
XML, CMake: Remove expat CPP sources from globbed C++ source file list.

### LINKED ISSUE #1167: CMake & POCO_UNBUNDLED: expat sources are compiled in libPocoXML

matejk:
List of C++ source files for libXML when building with CMake always contained an expat C++ source file xmlparser.cpp.

This file is now removed from default globbed source list and only added if POCO_UNBUNDLED is not set.

Fixes #1167

### DIFF
--- a/XML/CMakeLists.txt
+++ b/XML/CMakeLists.txt
@@ -1,8 +1,13 @@
 set(LIBNAME "XML")
 set(POCO_LIBNAME "Poco${LIBNAME}")
 
+# Expat CPP sources to be excluded from globbed SRCS_G
+# They are added back on the list if POCO_UNBUNDLED is not set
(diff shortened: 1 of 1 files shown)


matejk:
@obiltschnig This one could be a candidate for milestone 1.7.0?

PR #1176, opened 2016-03-07
Add support for IBM XLC on AIX platforms

mathausmendel:
This pull request adds support for IBM XLC on AIX platforms.

I tried to keep the changes to a minimum due to not being able to test in all the platforms POCO supports. So if anyone could do it to check if we are still good on Windows it would be great.

Also I'm aware that some warnings are disabled on XLC, but I did that in order to avoid warnings that are meaningless (like compiler memory usage for certain optimizations and such). If required, I can remove or document them.

I tested it on Linux with both GCC/Clang and on AIX with IBM XLC 13.1 and everything seems to be working fine.


obiltschnig:
Yes, that's the way to do it.

mathausmendel:
I corrected the code as we discussed. Hope the PR is now good enough to be merged.

PR #1177, opened 2016-03-08
Pr/1113 after merging develop

aleks-f:

PR #1186, opened 2016-03-10
Pr/1113 last fixes for Redis and Makefile

zosrothko:


zosrothko:
@aleks-f I think you can close also the PR #1177 and delete this pr/11113 branch.

aleks-f:
@zosrothko: not really, when I merged this into develop and then develop back to pr/1113, there were conflicts - some MySQL tests did not have `/Poco/CppUnit` include path, see [SQLExecutor in develop](https://github.com/pocoproject/poco/blob/develop/Data/MySQL/testsuite/src/SQLExecutor.cpp#L12) for an example. So, I fixed those conflicts after merge and it;s building now. if it passes, I'll merge it into develop.

aleks-f:
one travis  gcc run failed with a coredump in Net tests, for whatever reason

PR #1195, opened 2016-03-14
GH #1050 NetSSL_OpenSSL: fix gcc -Wshadow warnings

vmiklos:
Hi,

Quite similar as the previous ones, this time NetSSL_OpenSSL is now -Wshadow warning-free.

BTW, a related question. :-) 08a748dae166e9aa735ae90fc5d91a420dc132f2 added a nice little new feature on 2015-03-07, still I don't see it in the poco-1.7.1 branch, which has changes from 3 days ago. Is it expected that some changes don't end up in a release branch even after more than a year? ;-)

Thanks.

### DIFF
--- a/Foundation/include/Poco/Optional.h
+++ b/Foundation/include/Poco/Optional.h
@@ -62,9 +62,9 @@ public:
 	{
 	}
 
-	Optional(const C& value): 
+	Optional(const C& rValue):
 		/// Creates a Optional with the given value.
-		_value(value), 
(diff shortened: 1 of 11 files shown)


aleks-f:
@vmiklos: the reason is because 1.6.x branch became 1.7 and that contribution was never ported there. it will be in 1.8

vmiklos:
Aha! Thanks for explaining.

PR #1218, opened 2016-03-29
Fixes crash on streamed unzip

stracol:
Fixes a crash if the incoming stream has less than the header size available.

### DIFF
--- a/Zip/src/ZipDataInfo.cpp
+++ b/Zip/src/ZipDataInfo.cpp
@@ -41,9 +41,15 @@ ZipDataInfo::ZipDataInfo(std::istream& in, bool assumeHeaderRead):
 	_valid(false)
 {
 	if (assumeHeaderRead)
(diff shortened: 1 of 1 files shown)


stracol:
Hi,

The appveyor job failed on Environment: builder=msbuild, vsver=120, linkmode=shared; Platform: x64.
with the test *testReuseSocket* in *NetSSL_Win* with logs:

>  
There was 1 error: 
 1: class CppUnit::TestCaller<class TCPServerTest>.testReuseSocket
    "class Poco::IOException: I/O error"
    in "<unknown>", line -1
 

> EchoConnection: SSL Exception: Failed to receive data in handshake
> EchoConnection: SSL Exception: Failed to receive data in handshake
> EchoConnection: Software caused connection abort

As this looks not related to this patch (zip concern), does anyone can help or relaunch test suite ?

PR #1262, opened 2016-05-08
Update VS2013 projects with a personalized generation step to compile

zosrothko:
Update VS2013 projects with personalized generation steps for shared/static, debug/release build in order to compile TimeHandler.cpsp to TimeHandler.ccp/TimeHandler.h before the cl compilation.


aleks-f:
@zosrothko: editing VS projects directly is a futile job because they will be overwritten by ProGen. Edit .progen config instead and regenerate VS projects. See [Crypto.progen](https://github.com/pocoproject/poco/blob/develop/Crypto/Crypto.progen#L22) for similar functionality.

obiltschnig:
Unfortunately ProGen currently cannot deal with custom build steps like those needed for running PageCompiler. Maybe we should just add the generated file back into Git, like it was before.

aleks-f:
I'm not at all familiar with page compiler, but doesn't it have capability to specify output on the command line? If it does, then pre-build should work.

zosrothko:
The added prebuild step loooks like
..\..\bin\cpspc src/TimeHandler.cpsp
and by default, the page copmpiler generates files into the same directory as the cpsp file, i.e src in this case which is fine for the build.

Issue #1265, opened 2016-05-10
Issue for std::string-typed path / file name containing the 0x00 char


alex-87:
Hello,

  There is an issue concerning the wrong account of the null (not allowed) character in a path, which results in an undefined behavior. The bug is reproducible like following:

```
#include <Poco/Path.h>
#include <Poco/File.h>
#include <iostream>

int main(int argc, char** argv)
{
    std::string fileName = "Text.txt";
    fileName.append(2, 0x0);
    fileName.append(2, 'x');

    Poco::File file( Poco::Path::expand(fileName) );

    if ( file.exists() ) {
        std::cout << "The file "
              << Poco::Path::expand(fileName)
              << " exists"
              << std::endl;
    }

    return 0;
}
```

The result of this sample, **admitting the file** `Text.txt` **really exists**, is:
`The file Text.txtxx exists` (of course, **there is no** `Text.txtxx` **file**).

The issue is caused by the fact that POCO considers the null-terminated string only, but std::string doesn't, by taking into account all of the symbols, and not only the null-terminated char sequence. Consequently, the result of the `file.exists()` isn't correct because of the mal-formed-but-correct string corresponding to the file name.

Hoping it helps,
Have a great day.

micheleselea:
I don't know but usually the concept of "string" is a null terminated sequence of characters. I think is not wrong that the string considered in your case it's Text.txt because if you, for example, for some reason in tour code put the '\0' char to make the string exactly what you need (Text.txt) without resising string the behavior is correct...but it's just my point of view.

alex-87:
I agree with the fact that POCO considers, in this case, `Text.txt` only. My concern is more about processing the file name, for example here:

```
FileImpl::FileImpl(const std::string& path): _path(path)
{
	std::string::size_type n = _path.size();
	if (n > 1 && _path[n - 1] == '/')
		_path.resize(n - 1);
}
```

In [File_UNIX.cpp](https://github.com/pocoproject/poco/blob/develop/Foundation/src/File_UNIX.cpp), for example, either we consider an entier string by using `_path.size()`, or we consider a null-terminated string by using `std::char_traits<char>::length(_path.data())`, otherwise the behaviour could be random.

This is why I opened a pull request which only ensures there is no possible gap between the null-terminated string and a string which contains a null char.

PR #1266, opened 2016-05-10
Fix for issue #1265.

alex-87:
The fix purpose to add the following assertion :
`poco_assert( std::char_traits<char>::length(path.data()) == path.size() );`

The `std::char_traits<char>::length` considers the null-terminated sequence, while the `std::string::size` considers the total number of symbols of the string.

The assertion ensures the sizes are the same.

Hoping it helps,
Thank you for considering my pull-request =)

Have a great day.

### DIFF
--- a/Foundation/src/File.cpp
+++ b/Foundation/src/File.cpp
@@ -46,6 +46,7 @@ File::File()
 
 File::File(const std::string& rPath): FileImpl(rPath)
 {
+	poco_assert( std::char_traits<char>::length(rPath.data()) == rPath.size() );
 }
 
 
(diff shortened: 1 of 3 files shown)


aleks-f:
@alex-87 : Thanks for the contirbution, but this pull randomly breaks [this test](https://github.com/pocoproject/poco/blob/develop/Foundation/testsuite/src/PathTest.cpp#L1447). Typically, pulls should come with tests and/or contributors should make sure that tests pass. It may be a good thing that this test fails, but we have to get it fixed before this is merged.

alex-87:
Yes, thank you for the observation. I just committed the modification which ensures the tests pass.

PR #1273, opened 2016-05-16
[CppUnit/RepeatedTest.h] Add missing return type

tkoeppe:


tkoeppe:
@aleks-f: I don't think the test failure is related to this change. Could you please have a look?

PR #1274, opened 2016-05-16
Added some missing inclusions

tkoeppe:


tkoeppe:
Thanks!

aleks-f:
better late than never :) sorry for the long wait

PR #1276, opened 2016-05-16
Event_POSIX: use monotonic clock if available

radimdvorak:


aleks-f:
@radimdvorak Event_POSIX does not exist in 2.0 anymore (std::condition_variable used),  perhaps consider sending pull to 1.8 branch

PR #1277, opened 2016-05-16
GH #1050 Net tests: fix gcc -Wshadow warnings

vmiklos:


vmiklos:
Hi,

Ping, can I help anything with reviewing this?

Thanks,

Miklos

vmiklos:
Hi,

Second ping, can I do anything to get this reviewed?

Thanks,

Miklos

obiltschnig:
Personally I'm not too happy with those fixes as they do not really improve the readability of the code by decorating parameter names with otherwise useless  (except silencing -Wshadow) prefixes, but since we've already started with this I'll merge it.

zosrothko:
Agreed with @obiltschnig ... This is a feature of C and C++ to allow a local variable to have the same name as a parameter, type, class member and the scope of that name in that configuration is well defined by the standard. 

Moreover, the warning case is incomplete as in C++, the compiler warns if a local variable shadows an explicit   typedef, but not if it shadows a struct/class/enum.

```
-Wshadow
           Warn whenever a local variable or type declaration shadows another variable, parameter, type, class member (in C++), or instance variable (in
           Objective-C) or whenever a built-in function is shadowed. Note that in C++, the compiler warns if a local variable shadows an explicit
           typedef, but not if it shadows a struct/class/enum.
```

vmiklos:
OK, thanks for the feedback.

PR #1279, opened 2016-05-19
Added preprocessor guards for "define"s.

hakan-akan:
Many projects have already defined these flags. Redefining them causes a warning.

### DIFF
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -23,7 +23,9 @@
 // Define to enable Windows Unicode (UTF-8) support
 // NOTE: As of POCO C++ Libraries release 1.6.0, compiling POCO
 // without POCO_WIN32_UTF8 defined on Windows is deprecated.
+#ifndef POCO_WIN32_UTF8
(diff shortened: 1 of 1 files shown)


aleks-f:
@zosrothko : what's up with all these appveyor fails. can you take a look?

hakan-akan:
@aleks-f @zosrothko It seems like Visual Studio 12 does not support default move constructor and default move assignments in PoolableConnectionFactory.h. These are being conditionally compiled with the POCO_ENABLE_CPP11 flag. And this flag is conditionally being #defined at Platform_Win32.h in case it is (#if defined(_MSC_VER) && (_MSC_VER >= 1700) && !defined(POCO_ENABLE_CPP11)). This condition is not sufficient. I think it should be (_MSC_VER >= 1900) which corresponds to Visual Studio 2015. I am not sure if this would be too heavy handed since doing this would remove all other c++11 features until Visual Studio 2015. I submitted the #1286 to address this for the time being.

zosrothko:
This should help: https://msdn.microsoft.com/fr-fr/library/hh567368.aspx#featurelist

PR #1282, opened 2016-05-23
Added cast operator from Poco::LocalDateTime into Poco::DateTime

skident:


obiltschnig:
I'm not really a fan of cast operators, as they tend to cause more troubles than worth it. Explicitly calling utc() makes it much more obvious.

skident:
`DateTime utc() const;
		/// Returns the UTC equivalent for the local date and time.`
I need local datetime with timezone in variable Poco::DateTime.
Cast operator is the simplest way to solve my problem, but I don't have any objections against method.
I just think that we should have some way to convert this two types into each other.

PR #1287, opened 2016-05-30
Add mechanism to start a task from within a task

phirg:
... staying in the same thread.

It does not seem possible to join a task. 
This proposal would allow to make e.g. some kind of task queues, or at least, to execute some tasks after each others, without leaving the thread.

### DIFF
--- a/Foundation/include/Poco/Task.h
+++ b/Foundation/include/Poco/Task.h
@@ -108,6 +108,16 @@ protected:
 		///
 		/// A Task should use this method in favor of Thread::sleep().
 
+	bool yield();
+	    /// Yields cpu to other threads
+		///
+		/// If the task is cancelled while it is suspended,
+		/// yield() will return true. If the tasks resumes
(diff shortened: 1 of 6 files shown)


phirg:
Regarding the task queue, I will try to implement it in a project using the taskFinished notification. I am not pretty sure how I will do this by now. I would need parallel task queues. I'll post something here if I succeed.

phirg:
I am not sure if the new test proposed in this last commit is very relevant since it only provides clues to build a queuing for the tasks. 

Known weaknesses:
 * the SimpleTaskQueue as proposed in the test should never be used "as is", since it is not thread safe (at all). Although it is safe enough in the context of the test. 
 * The task count is decremented before re-incremented if the taskFinishedNotification is used to launch the next thread as proposed in the test. It could drive to strange behavior if any makes the assumption that a task count of zero means that all the tasks are finished. 

If you judge that those proposals are too much case-specific, I would surely understand it very well.

PR #1296, opened 2016-06-16
std:: namespace changes to fix compile errors

Bjoe:
Add or remove std:: namespace to fix compile errors with Dinkum C/C++ Standard Library on QNX/Blackberry 10.
See also commit message.


aleks-f:
@Bjoe sorry for very delayed response. I'm afraid we'll have to turn this down for 1.x - we've had a disaster in the past with this thing, haven't heard of any problems for a while and don't want to go back there. the problem, as far as i can remember, was on android and some older VS versions. If you need it badly, please ifdef for qnx and re-send, I'll take it :)

Bjoe:
@aleks-f I revert some commits and add some C compatibility header, because there the function is used without namespace. This means, we should also use xxx.h header because these headers included the function in the global namspace.

From http://en.cppreference.com/w/cpp/header
Each xxx.h header included in the C++ standard library places in the global namespace each name that the corresponding cxxx header would have placed in the std namespace.
These headers are allowed to also declare the same names in the std namespace, and the corresponding cxxx headers are allowed to also declare the same names in the global namespace: including <cstdlib> definitely provides std::malloc and may also provide ::malloc. Including <stdlib.h> definitely provides ::malloc and may also provide std::malloc

In QNX or Dinkum Standard library http://www.dinkumware.com/ they only included the function in the global namespace with xxx.h header and with cxxx header they only placed the function in the std namespace.

I think, we should add std:: namespace when we use cxxx header files.
At the moment I include for example in NummericString.h 
#include <cctype>
#include <ctype.h>
both variants of headers.

aleks-f:
I'd have to look at it and we should check what does at least windows and android say about the change - unfortunately, we are trying to cater to many platforms and not al of them fully comply with standards. Furthermore, this pull is to 1.7.3, which has been released long time ago. Best to close it and create a clean new one to 1.8.0 branch

Bjoe:
I already merge poco 1.8.0 in, but anyways ... it is a good idea. I cleanup the code/commits and created a new pull request, see https://github.com/pocoproject/poco/pull/1946

PR #1315, opened 2016-07-15
Redefine Poco assertions for static analysis

kimgr:
Static analyzers typically don't do cross-translation unit analysis, so
they don't understand that most of the Poco::Bugcheck methods never
return.

Unfortunately, some of them also don't understand noreturn annotations
(see e.g. http://en.cppreference.com/w/cpp/language/attributes), so
decorating Bugcheck doesn't help.

Instead redefine all the Poco assertion macros to just call abort() iff
we're parsed by a known static analyzer.

### DIFF
--- a/Foundation/include/Poco/Bugcheck.h
+++ b/Foundation/include/Poco/Bugcheck.h
@@ -81,6 +81,25 @@ protected:
 //
 // useful macros (these automatically supply line number and file name)
 //
+#if defined(__KLOCWORK__) || defined(__clang_analyzer__)
+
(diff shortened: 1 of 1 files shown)


kimgr:
Not sure if this is something you could consider for upstream. We're carrying a local patch which reduces false positives from static analyzers quite a lot.

kimgr:
Friendly ping!

obiltschnig:
Was actually stumbling across the same issue lately ;-)

kimgr:
Great, thanks!

PR #1330, opened 2016-07-29
Additional Functions for Dynamic::Struct (clear(), swap(), getVar())

Cymaphore:
- Providing interfaces to std::map features clear() and swap()
- Adding getVar() functions with an interface similar to AbstractConfiguration::getString(), AbstractConfiguration::getInt(). Simplified and more robust usage of Struct in many situations. Just like in AbstractConfiguration, getVar returns by copy to avoid bad reference issues for those usecases.


aleks-f:
417be8d4d1fbe2664669997fca614c8b6a14997a

PR #1332, opened 2016-07-29
Fixes #1331 - Poco::HTMLForm fails to set content length for HTTPMessage::HTTP_1_0 for ENCODING_MULTIPART

### LINKED ISSUE #1331: Poco::HTMLForm fails to set content length for HTTPMessage::HTTP_1_0 for ENCODING_MULTIPART
Currently it seems that for `HTTPMessage::HTTP_1_0` connections (i.e. chunked transfer encoding is disabled) `HTMLForm::ENCODING_MULTIPART` do not get a `Content-Length` header.  Currently to make servers accept my posts, I must manually calculate the content length and set the header in my own code.  Here is a proposed rewrite of `HTMLForm::prepareSubmit` for discussion.  If it looks good I'll issue a pull request.  Having read the specs, I realize encoding forms and setting headers is complex, so I may be missing something obvious.  Also, for reference, I checked the https://github.com/pocoproject/poco/blob/master/Net/testsuite/src/HTMLFormTest.cpp and it doesn't seem that any tests are being done for this situation.  Anyway, here's a take on it:

``` c++
void HTMLForm::prepareSubmit(HTTPRequest& request)
{
    if (request.getMethod() == HTTPRequest::HTTP_POST || request.getMethod() == HTTPRequest::HTTP_PUT)
    {
        // 1. We inspect the encoding to set our content type.
        if (_encoding == ENCODING_MULTIPART)
        {
            _boundary = MultipartWriter::createBoundary();
            std::string ct(_encoding);
            ct.append("; boundary=\"");
            ct.append(_boundary);
            ct.append("\"");
            request.setContentType(ct);
        }
        else
        {
            request.setContentType(_encoding);
        }

        // 2. Enforce conforming keep-alive, but don't overwrite existing.
        if (request.getVersion() == HTTPMessage::HTTP_1_0)
        {
            request.setKeepAlive(false);
        }

        // 3. Enforce chunked transfer encoding, but don't overwrite existing.
        if (request.getVersion() == HTTPMessage::HTTP_1_0 || _encoding == ENCODING_URL)
        {
            request.setChunkedTransferEncoding(false);
        }

        // 4. Write the content length only if necessary (i.e. chunked transfer is false).
        if (!request.getChunkedTransferEncoding())
        {
            request.setContentLength(calculateContentLength());
        }
    }
    else
    {
        std::string uri = request.getURI();
        std::ostringstream ostr;
        writeUrl(ostr);
        uri.append("?");
        uri.append(ostr.str());
        request.setURI(uri);
    }
}
```

Existing for reference: 

https://github.com/pocoproject/poco/blob/master/Net/src/HTMLForm.cpp#L203-L243

bakercp:
This is a proposed fix for #1331.

### DIFF
--- a/Net/src/HTMLForm.cpp
+++ b/Net/src/HTMLForm.cpp
@@ -230,6 +230,10 @@ void HTMLForm::prepareSubmit(HTTPRequest& request)
 		{
 			request.setChunkedTransferEncoding(true);
 		}
+		if (!request.getChunkedTransferEncoding())
+		{
+			request.setContentLength(calculateContentLength());
+		}
 	}
 	else
 	{


obiltschnig:
Had to revert. Please no pull requests against master branch. Please resubmit for develop.

PR #1345, opened 2016-08-11
ThreadPool startWithOSPriority()

gasuketsu:
Add ability to start thread with OS specific priority and policy into ThreadPool


gasuketsu:
Currently there is no way to start thread in `ThreadPool` with OS specific priority, so added ability to do that.
Also, `Poco::Timer` that depends on ThreadPool will be able to start with given specific priority.

PR #1363, opened 2016-08-25
Internet is unavailable on build servers

CosminCremarenco-Murex:
It's more the exception that outside Internet is available from build servers, definitely not the case in our outfit. Since hosts are not available ping tests fail and also ntp tests that expect to find the ntp pool.


CosminCremarenco-Murex:
Strangely the test that failed doesn't seem to have anything to do with my change.

1: N7CppUnit10TestCallerI17LocalDateTimeTestEE.testTimezone

CosminCremarenco-Murex:
Will open a new pull request with a single commit.

PR #1408, opened 2016-09-16
Fix lost precision error

Kirozen:


Kirozen:
Maybe but in my case while an unsigned long is 32b and uintptr_t 64b gcc (6.1) returns an error because (unsigned long)ptr loses precision. I could avoid this error with this cast.

obiltschnig:
Also, you would need to include <cstdint> or <stdint.h> for uintptr_t. Better use Poco::UIntPtr. Your changes fail to build on Linux:

```
** Compiling src/xmlparse.cpp (debug, shared)
src/xmlparse.cpp: In function ‘long unsigned int generate_hash_secret_salt(XML_Parser)’:
src/xmlparse.cpp:739:77: error: expected type-specifier before ‘uintptr_t’
src/xmlparse.cpp:739:77: error: expected ‘>’ before ‘uintptr_t’
src/xmlparse.cpp:739:77: error: expected ‘(’ before ‘uintptr_t’
src/xmlparse.cpp:739:77: error: ‘uintptr_t’ was not declared in this scope
src/xmlparse.cpp:739:95: error: expected ‘)’ before ‘;’ token
```

Kirozen:
Ok, it is better to use Poco defined type.

PR #1410, opened 2016-09-16
Add ccache support with CMake

Kirozen:


obiltschnig:
@bachp could you quickly look over it?

bachp:
I'm not sure we want to enable this by default maybe not everybody who has ccache installed want's to use it.
It think it's is better to either add an **option** for it or just recommend to people to use:
 `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache`

PR #1426, opened 2016-09-26
inttypes.h is available on sun solaris

CosminCremarenco-Murex:
Definitions for integer types are available on Solaris in inttypes.h (much like stdint.h).


Foundation/src/pcre_internal.h
```diff
@@ -231,7 +231,7 @@ by "configure". */
 
 #if defined HAVE_STDINT_H
 #include <stdint.h>
-#elif defined HAVE_INTTYPES_H
```
Nuruzzaman5230: Still Alive

PR #1432, opened 2016-09-28
Fixes #1431 Poco/FIFOBuffer.h copy

### LINKED ISSUE #1431: Poco/FIFOBuffer.h copy issue
void copy(const T* ptr, std::size_t length)
		/// Copies the supplied data to the buffer and adjusts
		/// the used buffer size.
	{
		poco_check_ptr(ptr);
		if (0 == length) return;

		Mutex::ScopedLock lock(_mutex);
		
		if (length > available())
			throw Poco::InvalidAccessException("Cannot extend buffer.");
		
		if (!isWritable())
			throw Poco::InvalidAccessException("Buffer not writable.");

		std::memcpy(&_buffer[_used], ptr, length * sizeof(T));
		std::size_t usedBefore = _used;
		_used += length;
		if (_notify) notify(usedBefore);
	}

In the function, after the memcpy, '_used' should be assigned the value of  'length'.

keenbin7:
using 'std::memcpy(begin() + _used, ptr, length * sizeof(T))' instead of 'std::memcpy(&_buffer[_used], ptr, length * sizeof(T))'

and add some test case.


keenbin7:
In this pull, travis-ci can't pass 'Linux gcc 4.6 (make) bundled'.
However, it can pass in my travis-ci.
[https://travis-ci.org/keenbin7/poco/jobs/163326545/config](url)

PR #1435, opened 2016-09-28
MongoDB BSON type 0x11 support, fixes #1429

### LINKED ISSUE #1429: exception thrown in MongoDB  when using replicaset
I configured mongo to run in a replicaset and can't insert documents because of unsupported BSON type timestamp 0x11. When calling Connection::sendRequest() with insert command an exception is thrown from MongoDB/src/Document.cpp Document::read()

Not implemented: Element ts contains an unsupported type 11

I am using poco 2.0.0 on Ubuntu 16.04  x86_64.

tomaz-beltram:
New SpecialTimestamp support used only internally by mongodb handling replica set and sharded collections.

### DIFF
--- a/MongoDB/include/Poco/MongoDB/Element.h
+++ b/MongoDB/include/Poco/MongoDB/Element.h
@@ -291,7 +291,7 @@ inline void BSONWriter::write<NullValue>(NullValue& from)
 {
 }
 
-struct SpecialTimestamp {
+struct Timestamp {
 	Poco::Timestamp ts;
 	Poco::Int32 inc;
 };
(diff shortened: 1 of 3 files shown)


obiltschnig:
In the [BSON spec](https://docs.mongodb.com/manual/reference/bson-types/) this type is just called "timestamp", so maybe we could drop the "special" from the name here. Since it's in the Poco::MongoDB namespace this should not be an issue.

tomaz-beltram:
It will not be an issue for the compiler but for the human reader. What about InternalTimestamp?

obiltschnig:
BSONTimestamp should make it clear. 

--
Günter Obiltschnig
Applied Informatics Software Engineering GmbH

tomaz-beltram:
BSONTimestamp it is.

PR #1437, opened 2016-09-28
Fix #1241 Rename CMAKE Variables to allow grouping

### LINKED ISSUE #1241: Rename CMAKE Variables to allow grouping
Some context information:

I am currently building Poco Version  1.6.1 as a cmake project.

As I find it easier to work with a GUI tool when configuring cmake, I am using cmake-gui (https://cmake.org/cmake/help/v3.0/manual/cmake-gui.1.html)

Using cmake-gui it is possible to group the cmake variables. 

Problem:

When grouping, some of the variables belong to a group called POCO but many others are located under a group called ENABLE.

Proposed Solution:
Would it be possible to rename those variables (e.g. add POCO_ prefix)
As I am currently building on a linux machine I am doing a sed to rename, as an workaround:
grep -rl 'ENABLE_' ./ | xargs sed -i 's/ENABLE_/POCO_ENABLE/g'

bracketttc:


bachp:
Thanks, if we want to merge this before the 2.0 release we need to have a way to work with the old variables as well, maybe as some kind of alias. Otherwise we will most probably break a lot of builds.

bracketttc:
I can do that. Add a macro or section that looks for the old ENABLE_<COMPONENT> flags, sets the appropriate POCO_ENABLE_<COMPONENT> flags and then unsets the ENABLE_<COMPONENT> variable so that it doesn't show up in CMake configuration UIs.

With that, you could specify options on the CMake command-line using either naming system during the transitional period.

I'll try to get to that tonight.

bachp:
That would be awesome thanks

bracketttc:
Neat little bit of scripting runs through all the cached variables, looks for `ENABLE_*` variables that match up to `POCO_ENABLE_*` variables and copies those variables over to the `POCO_ENABLE_*` variable. It then cleans up after itself and removes the original `ENABLE_*` variable.

The clean-up bit is needed or that variable would become permanently forced and a user would have to delete their cache entirely, or `cmake -U ENABLE_*` or something.

aleks-f:
@bracketttc sorry for the neglect. I'd like very much to merge this for 2.0, but there's of course conflict now. If you don't mind syncing everything up, and if @bachp is still ok with this pull then please refresh it. @zosrothko can advise on anything regarding travis/appveyor CI

Issue #1439, opened 2016-09-29
Make Config.h a CMake configure_file


bracketttc:
[Foundation/include/Poco/Config.h](https://github.com/pocoproject/poco/blob/master/Foundation/include/Poco/Config.h) is a prime candidate to use CMake's [configure_file](https://cmake.org/cmake/help/v3.0/command/configure_file.html) feature. Then all of those commented out preprocessor defines become CMake options ( advanced or regular ).

Converting the [`POCO_ENABLE_CPP11`](https://github.com/pocoproject/poco/blob/master/Foundation/include/Poco/Config.h#L31-L32), [`POCO_DISABLE_CPP11` and `POCO_ENABLE_CPP14`](https://github.com/pocoproject/poco/blob/master/Foundation/include/Poco/Config.h#L35-L40) to CMake options ( and upping the minimum required CMake version to 3.1 ) would allow the use of [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD.html#variable:CMAKE_CXX_STANDARD) and [`CMAKE_CXX_STANDARD_REQUIRED`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD_REQUIRED.html#variable:CMAKE_CXX_STANDARD_REQUIRED) to get the C++ standard specified in compiler flags without messy conditional blocks.

It's possible to create the conditionals necessary for CMake < 3.1 to pass appropriate flags for C++ standards, but not recommended.

bracketttc:
As noted in commit comments, this breaks committed Makefile builds which are considered over the CMake-generated builds.

Closing PR and issue.

PR #1440, opened 2016-09-29
1439 make config h a cmake configure file

bracketttc:
Does not apply C++ standards using CMAKE_CXX_STANDARDS as in #1439 as that requires incrementally increasing the minimum required CMake version for the project.


bracketttc:
As noted in commit comments, this breaks committed Makefile builds which are considered over the CMake-generated builds.

Closing PR and issue.

PR #1445, opened 2016-10-04
Use stable_sort to preserve order of IP addresses from DNS

cryptoknight:
SocketAddress uses a sort operation on IP addresses returned from DNS to ensure that IPv4 addresses are preferred over IPv6 addresses. Unfortunately, std::sort does not guarantee that the order of "equal" elements will be preserved, in which case SocketAddress can wind up with an IP other than the first one when multiple addresses are returned. Since DNS round robin-ing (https://en.wikipedia.org/wiki/Round-robin_DNS) assumes that a client will typically use the first address, it would be beneficial to have the order preserved by the sort. Switching to std::stable_sort achieves this.


bracketttc:
Nice catch.

PR #1473, opened 2016-10-26
-change shared pointer access private members to be protected

RobertAcksel:
… to create an extensibility point. we need to have access to the release policy instance of the shared pointer. we will use it to provide a weak pointer implementation.

### DIFF
--- a/Foundation/include/Poco/SharedPtr.h
+++ b/Foundation/include/Poco/SharedPtr.h
@@ -378,7 +378,7 @@ public:
 		return _pCounter->referenceCount();
 	}
 
-private:
+protected:
 	C* deref() const
 	{
 		if (!_ptr)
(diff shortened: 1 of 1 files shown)


RobertAcksel:
Since 'Pocos ci.appveyor.com is broken #1481' is fixed now, it might be time to review this pull request?!

aleks-f:
@obiltschnig ?

RobertAcksel:
Yes it would be better not to make the fields direktly acsessable but instead to provide protected getter and setter. This is a general rule which schould be followed always but wich is not respected by my commit. 

However I dont think thad friend declarations are a candidate here since that would miss the whole point of beeing able to derive and to do something with shared pointer.
And I read somewhere here that implementations of weak ptr would never go live into poco because thy would just mimic what is already in c++11 (which poco is moving towards)

PR #1480, opened 2016-11-07
ReuseAddress and reusePort options should be separeted.

skypexu:
The reuseAddress option is to reuse a socket in TIME-WAIT state, bind still fails
if the socket is in-use, server applications often use the option.

But reusePort can reuse a port which is already in-use, I don't think you want
to share a port with another irrelevant application, unless you know what you
are doing, this is a more error prone option only is used in rare cases.

### DIFF
--- a/Net/include/Poco/Net/SocketImpl.h
+++ b/Net/include/Poco/Net/SocketImpl.h
@@ -76,7 +76,7 @@ public:
 		/// the TCP server at the given address. Prior to opening the
 		/// connection the socket is set to nonblocking mode.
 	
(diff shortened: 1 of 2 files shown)


zosrothko:
All failed run stop after 2 hours of elapse time which is the upper limit allowed by AppVeyor to pocoproject. Thus the point is to find out why the Poco/Net takes so mush time or if it goes in loop since all failed build end with
```testOptions: ERROR
testSelect: ERROR
testSelect2: ERROR
testSelect3: 
testEchoUnixLocal: 
testStreamEcho: ERROR
testLargeStreamEcho: ERROR
```

PR #1491, opened 2016-11-18
enable CppUnit with ENABLE_CPPUNIT

Ubiquite:
I wasted some time because ENABLE_CPPUNIT seems not to enable anything,

Also we could enable CppUnit component without enabling the poco unit tests.


bachp:
Makes sense. Thanks

PR #1493, opened 2016-11-21
* fix logger message current thread id

vm2mv:


obiltschnig:
This will change behavior to use the system thread ID instead of the POCO-assigned thread ID. Note that current develop branch has both POCO thread ID and system ID, which is probably what you're looking for.

PR #1494, opened 2016-11-21
Net: fixed parsing of media type in message header

vm2mv:


obiltschnig:
What issue does this solve? According to the [RFC](https://tools.ietf.org/html/rfc7231#section-3.1.1.1), a semicolon is the only allowed separator.

vm2mv:
Our software processes lots of data received from different sources. It processes requests from many programs web browsers as well as responses from different servers. Unfortunately we often see data where comma (',') is used as separator

obiltschnig:
OK, I understand your point. Will have to think about it. My issue with such fixes is that the bad code sending badly formatted messages never gets fixed if everyone else changes their code to accept these badly formatted messages. It may be the right thing to do for your application, but the question is whether it's the right thing to do for a general purpose library like POCO.

PR #1495, opened 2016-11-21
* Net: fixed splitting parameters in message header

vm2mv:


obiltschnig:
What is the rationale behind the proposed change? This does not seem to correspond to any specification I am aware of.

vm2mv:
We often encounter with data which format doesn't match the official specification (as I mentioned before). In any case our software has to process such data. Here we fixed data parsing where the comma (,) symbol is used as separator instead of (;) and single quote symbol (') instead of double (")

PR #1497, opened 2016-11-21
* Net: fixed adding content part to mail message with empty content-d…

vm2mv:
…isposition


vm2mv:
oh, fiexed

obiltschnig:
ok, thanks.

PR #1499, opened 2016-11-21
Foundation: fixed exceptions handling in zlib-output-streams construc…

vm2mv:
…tors

### DIFF
--- a/Foundation/include/Poco/DeflatingStream.h
+++ b/Foundation/include/Poco/DeflatingStream.h
@@ -139,7 +139,7 @@ protected:
 };
 
 
-class Foundation_API DeflatingOutputStream: public DeflatingIOS, public std::ostream
+class Foundation_API DeflatingOutputStream: public std::ostream, public DeflatingIOS
 	/// This stream compresses all data passing through it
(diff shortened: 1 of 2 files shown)


obiltschnig:
How does [changing the order of inheritance](http://stackoverflow.com/questions/17328921/c-multiple-inheritance-order) change exception handling?

vm2mv:
If the **deflateInit2** function returns error in **DeflatingStreamBuf constructor**, then it (constructor) throws **Poco::IOException** exception. It causes invoking of **DeflatingOutputStream destructor** where **basic_ios** class destructor is called. After that "_0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF_" exception is throws in **ios_base::_Ios_base_dtor**.
CRT implies that **std::ostream** object must be destroyed as it was specified as the first in the inheritance order. In fact it's not yet created.
Changing the inheritance order can solve the problem - CRT doesn't try to destroy **std::ostream** because it (CRT) doens't consider the object as created.
Can be reproduced in Visual Studio 2015 Update 3

default

obiltschnig:
ok, thanks, makes sense now.

PR #1527, opened 2016-12-21
Update Timer.h

ljx0305:
The compiler hint warning must be explicit calls base class copy constructed function


ljx0305:
update error

PR #1542, opened 2017-01-04
Let the build system honor flags

jspricke:
Try to append flags where possible. Also remove the -Ox flags set by upstream

### DIFF
--- a/build/config/FreeBSD
+++ b/build/config/FreeBSD
@@ -20,7 +20,7 @@ CXX     = g++
 LINK    = $(CXX)
 LIB     = ar -cr
 RANLIB  = ranlib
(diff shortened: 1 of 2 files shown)


jspricke:
This has been part of the Debian patchset for quite some time and even seems to be proposed on the old sourceforge site.

obiltschnig:
Removing the -O2 flags from the default Linux and FreeBSD build configurations is not a good idea, as people building POCO from source will end up with unoptimized release builds by default.

obiltschnig:
Maybe having a separate build configuration for these build systems is a better idea?

jspricke:
Good point cmake has type None for that. Maybe I will simply switch the Debian build to it.

jspricke:
Would the rest of the patch be accaptable? Should I update it?

PR #1555, opened 2017-01-11
Net::HTTPSession: split timeout to connection, receive, send

proller:
Useful for setting 1s connect timeout.

tested and used in https://github.com/yandex/ClickHouse


proller:
Like this ?

obiltschnig:
Exactly. Thanks.

PR #1560, opened 2017-01-12
Allow link application with double_conversion

proller:
Without this linking fails with duplicate function names errors


dry013:
Please include in future releases, 1.9.0 doesn't contain it yet and it took some time to track down this fix.

adrianmorente:
Will this fix be included any time soon? Thanks!

PR #1563, opened 2017-01-17
Add support for std::tuple to Data/TypeHandler.h.

frwilckens:
This allows to use std::tuple instead of Poco::Tuple for SQL statements. 

I use this code in an in-house project. It was written mainly to support tables that have more than 20 columns (and to make use of C++ 11 features wherever it makes sense). It is tested on MacOS and Linux.

The code is visible only if compiled in C++ 11 mode.


frwilckens:
Looks like the Travis CI check failed because of problems unrelated to the modifications. A Linux package could not be installed. An earlier commit that differed only in source code formatting passed.

PR #1574, opened 2017-01-29
HTTPServerResponse: Allow to send more HTTP headers manually.

alexey-milovidov:
We need to send HTTP headers as indication of progress:
- at each progress notification, if we has not started to send response body, we add a header and flush it immediately.

It looks like this:

```
$ curl -sSv 'http://localhost:8123/' -d "SELECT URL, uniq(UserID) AS u FROM test.hits GROUP BY URL ORDER BY u DESC LIMIT 10"
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8123 (#0)
> POST / HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8123
> Accept: */*
> Content-Length: 82
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 82 out of 82 bytes
< HTTP/1.1 200 OK
< Date: Sun, 22 Jan 2017 16:51:32 GMT
< Connection: Keep-Alive
< Content-Type: text/tab-separated-values; charset=UTF-8
< Transfer-Encoding: chunked
< Keep-Alive: timeout=3
< X-ClickHouse-Progress: {"read_rows":"2752512","read_bytes":"240570816","total_rows":"8880128"}
< X-ClickHouse-Progress: {"read_rows":"5439488","read_bytes":"482285394","total_rows":"8880128"}
< X-ClickHouse-Progress: {"read_rows":"8783786","read_bytes":"819092887","total_rows":"8880128"}
< 
http://yandex.ru/       16748
```

We already use this patch in our product. I am not sure, that you will like the way it was implemented. Maybe you will suggest another solution.

Usage example:
https://github.com/yandex/ClickHouse/blob/3552de49ead229bb8ced77c87d4530e3c3e3eb03/dbms/include/DB/IO/WriteBufferFromHTTPServerResponse.h


alexey-milovidov:
Results from appveyor are probably incorrect:
https://ci.appveyor.com/project/obiltschnig/poco/build/build%20develop-1539/job/k1hjfw4bbsf2shsg

aleks-f:
@alexey-milovidov could you please add a test case for this?

PR #1594, opened 2017-02-15
update to sqlite 3.17.0

mkrivos:


mkrivos:
Approximately 25% better performance from the R-Tree extension.
        Uses compiler built-ins (ex: __builtin_bswap32() or _byteswap_ulong()) for byteswapping when available.
        Uses the sqlite3_blob key/value access object instead of SQL for pulling content out of R-Tree nodes
        Other miscellaneous enhancements such as loop unrolling. 
    Add the SQLITE_DEFAULT_LOOKASIDE compile-time option.
    Increase the default lookaside size from 512,125 to 1200,100 as this provides better performance while only adding 56KB of extra memory per connection. Memory-sensitive applications can restore the old default at compile-time, start-time, or run-time.
    Use compiler built-ins __builtin_sub_overflow(), __builtin_add_overflow(), and __builtin_mul_overflow() when available. (All compiler built-ins can be omitted with the SQLITE_DISABLE_INTRINSIC compile-time option.)
    Added the SQLITE_ENABLE_NULL_TRIM compile-time option, which can result in significantly smaller database files for some applications, at the risk of being incompatible with older versions of SQLite.
    Change SQLITE_DEFAULT_PCACHE_INITSZ from 100 to 20, for improved performance.
    Added the SQLITE_UINT64_TYPE compile-time option as an analog to SQLITE_INT64_TYPE.
    Perform some UPDATE operations in a single pass instead of in two passes.
    Enhance the session extension to support WITHOUT ROWID tables.
    Fixed performance problems and potential stack overflows when creating views from multi-row VALUES clauses with hundreds of thousands of rows.
    Added the sha1.c extension.
    In the command-line shell, enhance the ".mode" command so that it restores the default column and row separators for modes "line", "list", "column", and "tcl".
    Enhance the SQLITE_DIRECT_OVERFLOW_READ option so that it works in WAL mode as long as the pages being read are not in the WAL file.
    Enhance the LEMON parser generator so that it can store the parser object as a stack variable rather than allocating space from the heap and make use of that enhancement in the amalgamation.
    Other performance improvements. Uses about 6.5% fewer CPU cycles.

    Bug Fixes:
    Throw an error if the ON clause of a LEFT JOIN references tables to the right of the ON clause. This is the same behavior as PostgreSQL. Formerly, SQLite silently converted the LEFT JOIN into an INNER JOIN. Fix for ticket 25e335f802dd.
    Use the correct affinity for columns of automatic indexes. Ticket 7ffd1ca1d2ad4ec.
    Ensure that the sqlite3_blob_reopen() interface can correctly handle short rows. Fix for ticket e6e962d6b0f06f46e.

    Hashes:
    SQLITE_SOURCE_ID: "2017-02-13 16:02:40 ada05cfa86ad7f5645450ac7a2a21c9aa6e57d2c"
    SHA1 for sqlite3.c: cc7d708bb073c44102a59ed63ce6142da1f174d1

obiltschnig:
A bit too risky for the upcoming 1.7.8 release though ;-)

mkrivos:
I will watch for 3.17.1, any time for 1.7.8 release?

obiltschnig:
In the next few days. Mostly done with it, but like to do some more testing.

PR #1632, opened 2017-03-13
Update File.h

soyoo:
description input error?


aleks-f:
[if and only if](https://en.wikipedia.org/wiki/If_and_only_if)

PR #1637, opened 2017-03-16
Rollback fix #1605 since it breaks the unit test

### LINKED ISSUE #1605: SecureSocketImpl shutdown()
Hi all, I'm having a strange TCP RESET packet when a secure socket close connection with a server.
The point is this:

1)have an HTTPS server made with Poco libs
2)I have an HTTPSClientSession that make request
3)Connection start, data are transferred ok
4)when it's time to close both server (a little bit earlier) and client close connection
5)Closing a SecureSocketImpl make SecureSocketImpl::shutdown() and than a close() socket
6)I think that both server and client do the same
7)the "strange think" happen because when the client do shutdown the FIN connection is already sent from server and received from client so when client do shutdown() it sends some packet that will receive a TCP RST packet

I don't know if the explanation is clear or not, but I found it because some firewall interprets that RST packet like an attempt to port scan. The comment in shutdown method make me think that we already have some "experience" in shutting down ssl connection

zosrothko:
Hi
The fix done for solving the issue #1605 is breaking the unit test `HTTPSClientSessionTest ::testCachedSession()` that makes all CI AppVeyor and Travis jobs failing. Removing this fix makes all jobs from AppVeyor and Travis be successful..

### DIFF
--- a/NetSSL_OpenSSL/src/SecureSocketImpl.cpp
+++ b/NetSSL_OpenSSL/src/SecureSocketImpl.cpp
@@ -229,10 +229,15 @@ void SecureSocketImpl::shutdown()
 			// OpenSSL and/or browser versions, and things
 			// seem to work better now.
 			int rc = SSL_shutdown(_pSSL);
+#if FIXME
+			#1605: try to do a proper SSL_shutdown()
+			This fix for the issue above breaks the 
+			HTTPSClientSessionTest::testCachedSession() Unit test
 			if (rc == 0 && _pSocket->getBlocking())
 			{
 				rc = SSL_shutdown(_pSSL);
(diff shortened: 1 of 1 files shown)


zosrothko:
Additional information from the Travis jobs
```
testCachedSession: HTTPSTestServer: SSL connection unexpectedly closed
ERROR
!!!FAILURES!!!
Runs: 34   Failures: 0   Errors: 1

There was 1 error: 
 1: N7CppUnit10TestCallerI22HTTPSClientSessionTestEE.testCachedSession
    "N4Poco16TimeoutExceptionE: Timeout"
    in "<unknown>", line -1
```

PR #1645, opened 2017-03-21
Add SHA2 algorithm

OgreTransporter:
Add SHA2 algorithm because MD5 and SHA1 are not secure anymore.

### DIFF
--- a/Foundation/Foundation.vxbuild
+++ b/Foundation/Foundation.vxbuild
@@ -91,6 +91,7 @@ SOURCES="
 	RotateStrategy.cpp
 	Runnable.cpp
 	SHA1Engine.cpp
+	SHA2Engine.cpp
 	Semaphore.cpp
 	SharedLibrary.cpp
 	SimpleFileChannel.cpp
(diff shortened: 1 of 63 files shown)


obiltschnig:
Has this code been written entirely by you, or is it based on some other implementation. If so, we'll need to be careful about the license and add some acknowledgement.

OgreTransporter:
I've added license information to the files. The code is based on the implementation of mbed TLS (Apache 2.0, http://www.apache.org/licenses/LICENSE-2.0)

obiltschnig:
Aand, would you mind adding a test case. Just like [SHA1EngineTest.cpp](https://github.com/pocoproject/poco/blob/develop/Foundation/testsuite/src/SHA1EngineTest.cpp), but with the correct test vectors.

OgreTransporter:
Give me a few minutes ;-)

OgreTransporter:
Done ;-)

obiltschnig:
Thank you!

OgreTransporter:
You're welcome! I'm working on an implementation of SHA3, too.

OgreTransporter:
https://github.com/pocoproject/poco/pull/1664

PR #1647, opened 2017-03-22
Error: simplify strerror_r() handling

andred:
This is a simpler version of the existing strerror_r()
handling.

Idea borrowed from boost.asio:
  https://github.com/chriskohlhoff/asio/commit/443bc17d13eb5e37de780ea6e23157493cf7b3b9


andred:
I had this lying around from an older version, please feel free to ignore.

I do think it is simpler and easier to read than what is currently implemented, though.

PR #1664, opened 2017-04-01
Add SHA3 and BLAKE2 hashing algorithm

OgreTransporter:
This adds the SHA3 algorithm to complete the SHA group. Additional this adds the BLAKE2 algorithm too. BLAKE is an other SHA3 finalist (http://csrc.nist.gov/groups/ST/hash/sha-3/) algorithm which is extremly fast.


OgreTransporter:
Anything wrong with that code?

PR #1668, opened 2017-04-05
use double check for SingletonHolder

richardor:
use double check for SingletonHolder, avoid the lock competition whe calling the 'get' function every time

### DIFF
--- a/Foundation/include/Poco/SingletonHolder.h
+++ b/Foundation/include/Poco/SingletonHolder.h
@@ -54,8 +54,11 @@ public:
 		/// hold by the SingletonHolder. The first call
 		/// to get will create the singleton.
 	{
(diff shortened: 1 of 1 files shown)


orlinhristov:
Hi Richardor,
The Double-checked locking pattern is broken because of compiler optimizations, CPU caches and shared memory used on multi processor/core systems. There are many articles that explain the problems of this pattern. One of them was written by Scott Meyers and Andrei Alexandrescu back in 2004  (http://www.aristeia.com/Papers/DDJ_Jul_Aug_2004_revised.pdf). C++11 gives 3 different ways to implement thread-safe singleton initialization, but unfortunately I don't see any utilities in the Poco library that could help to achieve it.

PR #1672, opened 2017-04-06
CMake properly detects PostgreSQL, fixes #1671

### LINKED ISSUE #1671: PostgreSQL not found by CMake
I cloned the develop branch today to try working with the PostgreSQL connector but after some frustration I realized that CMake is not finding the required libraries, therefore not compiling the module.

I can run find_package(PostgreSQL) in a new CMake project and ${POSTGRESQL_FOUND} is set to TRUE. I think I have located the problem to the top-level CMakeLists.txt but I'm not certain since I'm not that experienced with CMake and C++.

To the point: If I try the following code
```
find_package(PostgreSQL)
message(FOUND: ${POSTGRESQL_FOUND} ";" ${PostgreSQL_INCLUDE_DIRS} ";" ${PostgreSQL_LIBRARIES} ";" ${PostgreSQL_VERSION_STRING})
```
before and after line 37:
```
# Append our module directory to CMake
set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
```

the console prints 
```
FOUND:TRUE;/usr/local/include/usr/local/include/server;/usr/local/lib/libpq.dylib;9.6.2
```
when the code is tried before and just `FOUND:;;;` when tried after.

Using:
psql 9.6.2 on macOS 10.11.6
CMake 3.7.2

Any ideas?
Thanks in advance

hampustagerud:
find_package(PostgreSQL) is built into CMake, at least since v2.8.8 (poco requires 3.0.0). Removing cmake/FindPostgreSQL.cmake and fixing a typo is enough for CMake to find PostgreSQL properly.


hampustagerud:
The variable is set by the FindPostgreSQL module, according to the documentation: https://cmake.org/cmake/help/v3.0/module/FindPostgreSQL.html this is the way it is spelled.

bachp:
Sorry I think my comment was not clear. If I look at the [source](https://github.com/Kitware/CMake/blob/master/Modules/FindPostgreSQL.cmake#L46) the found variable is spalled like this `PostgreSQL_FOUND `. I think it would be cleaner that way, even tough the old spelling still works.

hampustagerud:
I'm afraid that I still don't understand your question, the spelling matches the documentation now (it didn't earlier) so since CMake variables are case sensitive this is the way it should be?

Data/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ endif(ENABLE_DATA_MYSQL)
 if(ENABLE_DATA_POSTGRESQL)
     find_package(PostgreSQL)
     if(POSTGRESQL_FOUND)
```
bachp: Sorry, I'm talking about the `PostgreSQL_FOUND`. As you correctly pointed out the variables are case sensitive. And according to the [documentation](https://github.com/Kitware/CMake/blob/master/Modules/FindPostgreSQL.cmake#L46) this should be `PostgreSQL_FOUND` not `POSTGRESQL_FOUND`.

hampustagerud: I'm with you now (finally hehe) and yes,  it could probably be changed to adhere to the documentation but since it didn't break any functionality I didn't even notice it. The do however set PostgreSQL_FOUND first on line 178 and the set it to the value of POSTGRESQL_FOUND so functionalitywise it doesn't matter really. It doesn't follow the documentation exactly but since all other variables in Data/CMakeLists.txt are using uppercase only I think it looks more coherent using the "inofficial" POSGRESQL_FOUND?

bachp:
LGTM 👍

astillich:
I just stumbled over this, could you please merge the PR?

OgreTransporter:
I agree it's not necessary to have a script for a working build in function.

PR #1679, opened 2017-04-08
Add POCO_VERBOSE_MESSAGES option

audiofanatic:
Projects may want to bring poco into their build with `add_subdirectory()`,
`ExternalProject_Add()` or any number of other methods.
The existing behaviour causes quite a lot of output at the CMake stage
and this can be undesirable. This merge request provides a new option
POCO_VERBOSE_MESSAGES which can be used to control this. The
option is ON by default to preserve existing behaviour. All find-related
messages, warnings, errors, etc. are still always logged.


zosrothko:
This seems reasonable to merge since the existing behavior is kept while bringing a new usage of poco.

audiofanatic:
Sorry to comment on a closed pull request, but I'm wondering why this doesn't seem to have made it into the past few releases even though it was merged?

PR #1680, opened 2017-04-10
Enable parsing absolute Windows paths even if it starts with a slash

csabakeszegh:
In file URI when using absolute path on windows one has to write additional
'/' before the drive letter (empty host part). Eg: `file:///c:/windows/system32/`
`Path::parseWindows` throws an exception when using `URI::getPath()` to feed
it directly. The whole point of the commit to enable `/<DRIVE_LETTER>:` style
path as absolute path.
Test case added.


csabakeszegh:
Seems like the windows node in the CI system, can't download some prerequisites.

PR #1690, opened 2017-04-17
Publish poco_trace & poco_debug macro even in release mode

zosrothko:
A user of a prepackaged release versio of poco would have to add the constant POCO_DEBUG_LOG in order to get non empty  poco_trace and poco_debug macros. There is nothing removed or added to poco itself if those macros are published even in a release build. Publishing those macros by default make life easier for the poco's users.

### DIFF
--- a/Foundation/include/Poco/Logger.h
+++ b/Foundation/include/Poco/Logger.h
@@ -647,48 +647,35 @@ private:
 #define poco_information_f4(logger, fmt, arg1, arg2, arg3, arg4) \
 	if ((logger).information()) (logger).information(Poco::format((fmt), (arg1), (arg2), (arg3), (arg4)), __FILE__, __LINE__); else (void) 0
 
-#if defined(_DEBUG) || defined(POCO_LOG_DEBUG)
-	#define poco_debug(logger, msg) \
-		if ((logger).debug()) (logger).debug(msg, __FILE__, __LINE__); else (void) 0
+#define poco_debug(logger, msg) \
+	if ((logger).debug()) (logger).debug(msg, __FILE__, __LINE__); else (void) 0
 
-	#define poco_debug_f1(logger, fmt, arg1) \
-		if ((logger).debug()) (logger).debug(Poco::format((fmt), (arg1)), __FILE__, __LINE__); else (void) 0
+#define poco_debug_f1(logger, fmt, arg1) \
+	if ((logger).debug()) (logger).debug(Poco::format((fmt), (arg1)), __FILE__, __LINE__); else (void) 0
 
-	#define poco_debug_f2(logger, fmt, arg1, arg2) \
-		if ((logger).debug()) (logger).debug(Poco::format((fmt), (arg1), (arg2)), __FILE__, __LINE__); else (void) 0
+#define poco_debug_f2(logger, fmt, arg1, arg2) \
+	if ((logger).debug()) (logger).debug(Poco::format((fmt), (arg1), (arg2)), __FILE__, __LINE__); else (void) 0
 
-	#define poco_debug_f3(logger, fmt, arg1, arg2, arg3) \
-		if ((logger).debug()) (logger).debug(Poco::format((fmt), (arg1), (arg2), (arg3)), __FILE__, __LINE__); else (void) 0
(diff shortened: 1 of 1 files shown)


cristiantm:
This changes a contract on the usage of Poco Library. Our code relyes on this to make sure some logs will be available only on development, but will not even exist on production code.

We think this should be reverted because probably many other Poco users already expect that poco_trace and poco_debug won´t produce code on release, and this logs may contain sensitive data that should not be logged on production code.

aleks-f:
@zosrothko I'm leaning with @cristiantm on this one - why couldn't we just remove the (redundant) `POCO_LOG_DEBUG`? that way, there would be nothing a package user should do and macros would not exist in debug builds

zosrothko:
@aleks-f Removing only the `POCO_LOG_DEBUG` does not solve the original issue, which is for a `packaged` Poco, i.e. compiled poco. Even in this case, the poco_debug & poco_trace will not be available in the `release` version, wich is the target of this issue.

aleks-f:
> Even in this case, the poco_debug & poco_trace will not be available in the release version of the application

That is what @cristiantm objected to - enabling macros in release builds, that were previously disabled. 

The reason I'm agreeing is because of following risks:

- as mentioned - there may be sensitive data in debug/trace logs, which people do not ever want in release/production anywhere, log files of binary code; macros solve that that problem
- performance problem 1: debug/trace logs may be massive and accidentally enabling them through configuration could seriously affect production code performance, clog up disk etc.
- performance problem 2: logging framework imposes performance penalty of checking the log level, macros doubling that by first checking if they at the appropriate log level and should log, then logging framework checking again

We can help users address the above concerns by separating the modes of operation - Logging framework code itself is always completely present and if user chooses to use it directly, then he should be aware of the risks listed above. On the other hand, if user has something they never-ever want in production, they can use macros and be 100% safe.

zosrothko:
`POCO_LOG_DEBUG` is needed for the users that are building Poco from the souce code and want `poco_debug() & poco_trace()` in their release build.

zosrothko:
@cristiantm  Are you building Poco from the source code or do you pickup a packaged Poco?

zosrothko:
While I understand some arguments for restoring the previous version, it is annoying point from my perpective and from the software engineering point of view. When one develops an application that is to be delivered as a binary, it is a `must have` feature to be able to activate the embedded `poco_debug` & `poco_trace` logs on a production system without having to switch to the debug packaged binary....

obiltschnig:
I'm with Alex here. The `poco_debug` and `poco_trace` macros had it clearly stated in their documentation that they will be only armed for debug builds. Now suddenly also arming them in release builds is a severly breaking change.

aleks-f:
>  it is a must have feature to be able to activate the embedded poco_debug & poco_trace logs

@zosrothko I understand what you mean and agree with the principle, but then the reality kicks in;  using the logging macros is not mandatory and for that purpose we have the logging system - use it directly, then you can enable/disable logs at will, in any build.

Use macros for excessive or sensitive logging that you will never want in production. Like eg. that trace log dumping 10 bytes every second during debugging session, because you were developing with one device, but becomes 10MB every second in production because your customer has 1000 devices, you enable trace logging on customer's system to troubleshoot and then troubleshoot the troubleshooting :). Silly example, but I've experienced similar things.

If the purpose and extent of macros is communicated and documented clearly, then there's no confusion or violation of engineering principles.

zosrothko:
Fixed in develop with 4ef29eb0410a457fe4829a791c9cc362eb9437a6

cristiantm:
Thanks. I think this was the correct decision, since there is an expected behaviour that should not be broken. 

However, I can understand @zosrothko. It is not that clear on the docs (https://pocoproject.org/docs/Poco.Logger.html) that there is such behaviour, and usage of macros in all cases is somehow encouraged for performance reasons. So one that do wants debug logs on production code would use those and be surprised it does not work until he reads the header file. 

Suggestions:
1) Make it clearer on docs (https://pocoproject.org/docs/Poco.Logger.html) how the poco_debug/poco_trace macros work.
2) Maybe offer "release mode" macros for debug and trace. I would like to have better name ideas, but something like poco_debug_release and poco_trace_release).

PR #1691, opened 2017-04-17
Add SQLite Notifier table name getter

petko:
The SQLite callback table name is available anyway, so we just expose it in the Notifier..

### DIFF
--- a/Data/SQLite/include/Poco/Data/SQLite/Notifier.h
+++ b/Data/SQLite/include/Poco/Data/SQLite/Notifier.h
@@ -137,6 +137,9 @@ public:
 	void setRow(Poco::Int64 row);
 		/// Sets the row number.
 
+    std::string getTable() const;
+        /// Returns the table name.
+
 	const Poco::Dynamic::Var& getValue() const;
 		/// Returns the value.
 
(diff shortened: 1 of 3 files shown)


Data/SQLite/include/Poco/Data/SQLite/Notifier.h
```diff
@@ -178,6 +182,12 @@ inline Poco::Int64 Notifier::getRow() const
 }
 
 
+inline std::string Notifier::getTable() const
```
aleks-f: let's return `const std::string&` here

Data/SQLite/src/Notifier.cpp
```diff
@@ -27,6 +27,7 @@ namespace SQLite {
 Notifier::Notifier(const Session& session, EnabledEventType enabled):
 	_session(session),
 	_row(),
+    _table(),
```
aleks-f: not needed

Data/SQLite/include/Poco/Data/SQLite/Notifier.h
petko: OK

Data/SQLite/src/Notifier.cpp
petko: I've corrected this one in the previous commit

PR #1741, opened 2017-06-01
moved #endif to the correct place

wegylexy:
Before
```
#if...
namespace ...{
    ...
#endif
}
```

After
```
#if...
namespace ...{
    ...
}
#endif
```


wegylexy:
looking forward to a new stable release with better MongoDB support

obiltschnig:
Any specific requests regarding MongoDB support?

wegylexy:
GridFS; WriteResult for update request; connection URI, including replica set name constraint in URI

PR #1743, opened 2017-06-02
[Format] add variable width support

was4444:


obiltschnig:
Thanks! Could you please also update the documentation in the header file?

was4444:
@obiltschnig please feel free to correct my pool English.

PR #1744, opened 2017-06-02
[Format] support format characters pointer as %s

wasphin:
There is a case that I need to use both ``poco`` and another c++ library which has it's own string class. And its annoying to write code like:
```
Poco::format("%s", std::string(otheStr.c_str()));
```


aleks-f:
@obiltschnig this is a sensible request, but I vaguely remember we rejected it long time ago, not exactly sure why anymore

morris-stock:
I previously thought we cannot pass a char array to ``Any`` easily, and ``const char *`` is good enough(as least for me). Support ``char *`` to stay impartial.

morris-stock:
ping

PR #1762, opened 2017-06-21
Propertylist configuration aka plist on OS X

DraconPern:
I have created a class for managing plist style configuration files.  It's used mainly on OS X.  This is for the XML formatted version that's used by most software and editable by X Code.  It passes all AbstractConfiguration tests.  An additional test for reading plist.  It currently doesn't support arrays.  It has full read write support for string, integer, and Boolean type.  

Just a note that setInt and setBool seem to be a kludge in regards to events.
A save/load test should be added, but I am not sure where the temp file should be stored.

Please review, thanks!

### DIFF
--- a/PDF/include/Poco/PDF/XMLTemplate.h
+++ b/PDF/include/Poco/PDF/XMLTemplate.h
@@ -1,6 +1,17 @@
 //
 // XMLTemplate.h
 //
+// Library: PDF
+// Package: PDFCore
+// Module:  XMLTemplate
+//
+// Definition of the XMLTemplate class.
+//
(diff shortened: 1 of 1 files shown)


DraconPern:
Not sure what file needs to changed to fix linking.

aleks-f:
Use [Path::temp()](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Path.h#L313). As for linking, you should add any new cpp file(s) to the [Makefile](https://github.com/pocoproject/poco/blob/develop/Util/Makefile)

PR #1769, opened 2017-06-27
Change sudo: required to sudo: enabled as per Travis request

zosrothko:


.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get install -qq -y sloccount cppcheck; fi
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test; fi
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get update -qq; fi
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get install -qq -y g++-4.8; fi
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sudo apt-get install -qq -y g++-4.6; fi
```
aleks-f: actually, we should upgrade this to 4.9, that will be the min version for 2.0. @obiltschnig ?

zosrothko:
@aleks-f g++ 4.9 is not available on Ubuntu trusty 14.04LTS. The latest g++ version is 4.8. See [g++ on trusty](https://packages.ubuntu.com/search?suite=trusty&searchon=names&keywords=g%2B%2B)

aleks-f:
isn't there a newer OS available? 2.0 release is coming, it's way past time for c++11

zosrothko:
The point is that for now, Travis proposes only the Ubuntu trusty not the one that supports g++ std=c++11 for open source projet. I will follow up with travis and let you know.

zosrothko:
At Travis, Ubuntu trusty is yet in beta status and there is no plan (right now) for upgrading to new releases of Ubuntu. Thus the only way I see for testing poco with c++1 features is to use a backport gcc 5.2.1 on Ubuntu trusty found [here ](https://askubuntu.com/questions/618474/how-to-install-the-latest-gcurrently-5-1-in-ubuntucurrently-14-04). If you agree with this option, I can manage it once the switch to Poco 2.0 is acted.

aleks-f:
@zosrothko we are good with c++11 in develop branch immediately; 1.8, which will be based on master/1.7 will remain pre-c++11
/cc @obiltschnig

zosrothko:
Useless for c++11 branch

Issue #1776, opened 2017-06-29
Throwing exception inside of ~BasicBufferedStreamBuf()


sapharow:
Hello,

The destructor of BasicBufferedStreamBuf() calls Allocator::deallocate which can throw exceptions. In particular it throws Poco::SystemException when can't lock mutex. 

I've encountered such case in rare cases when application is about to finish and when ~BasicBufferedStreamBuf() is being called from detached thread whilst main thread is already finished and static variables such as Allocator's pool are destroyed. Enclosing body of ~BasicBufferedStreamBuf() into try/catch solves the issue.

References:
https://isocpp.org/wiki/faq/exceptions#ctor-exceptions
http://www.codingstandard.com/rule/15-2-1-do-not-throw-an-exception-from-a-destructor/

niparx:
I wonder, why this fix is not released yet? I mean, you have a fix, you had several releases already. But it still not there. Why?

PR #1777, opened 2017-06-30
Exception in destructor

sapharow:
The destructor of BasicBufferedStreamBuf() calls Allocator::deallocate which can throw exceptions. In particular it throws Poco::SystemException when can't lock mutex.
#1776

### DIFF
--- a/Foundation/include/Poco/BufferedStreamBuf.h
+++ b/Foundation/include/Poco/BufferedStreamBuf.h
@@ -67,7 +67,13 @@ public:
 
 	~BasicBufferedStreamBuf()
 	{
-		Allocator::deallocate(_pBuffer, _bufsize);
+		try 
+		{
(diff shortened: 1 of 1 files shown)


Foundation/include/Poco/BufferedStreamBuf.h
```diff
@@ -67,7 +67,13 @@ class BasicBufferedStreamBuf: public std::basic_streambuf<ch, tr>
 
 	~BasicBufferedStreamBuf()
 	{
-		Allocator::deallocate(_pBuffer, _bufsize);
+		try 
+		{
+			Allocator::deallocate(_pBuffer, _bufsize);
+		} catch (...)
+		{
+			// Do nothing
```
aleks-f: please put `poco_unexpected();` here

Issue #1793, opened 2017-07-05
C++11 support


aleks-f:
It's time. g++4.9, clang3.3 and VS2015 will become minimum compiler versions.

ibroheem:
With g++-4.9, a lot of C++14 features can also creep in.

aleks-f:
@ibroheem there are [defines](https://github.com/pocoproject/poco/blob/c%2B%2B11/Foundation/include/Poco/Config.h#L43) for that

ibroheem:
OK, seen!
So it is C++11/C++14 in reality!

Time to have true asyn lib using lambdas as callbacks!

whidbey:
I still like vs2010.... 
vs2015 is too bigger

micheleselea:
yes! great job...I'll will soon migrate from vs2015 to vs2017

micheleselea:
@whidbey I think vs2010 is the IDE with more bugs MS ever done :D

zosrothko:
@micheleselea Can you have a look to the review [here](https://github.com/pocoproject/poco/pull/1825)

micheleselea:
sorry @zosrothko  do not understand what I should do

zosrothko:
@micheleselea comment the change you made to the source Socket.cpp (why, reference, ...)

PR #1795, opened 2017-07-05
C++11 support #1793

aleks-f:
this goes into develop, then 2.0 release.
1.8.x will be last major non-c++11 release
	
- [x] All changes are protected by defines POCO_ENABLE_CPP11 and POCO_ENABLE_CPP14		
- [x] Replace std::auto_ptr with std::unique_ptr		
- [x] SharedPtr inherits from std::shared_ptr		
- [x] AtomicCounter inherits from std::atomic_int		
- [x] Array inherits from std::array		
- [x] Move semantics for SharedPtr, AutoPtr, Buffer, Message, JSON::Object and JSON::Array		
- [x] Use <cstdint> for int types		
- [x] Clock and Timestamp uses <chrono>		
- [x] Use std::mutex* for Mutex instead of platform specific		
- [x] Use std::condition_variable for Event instead of platform specific		
- [x] Use std::thread for Thread instead of platform specific		
- [x] Use template variadic for Poco::format and Logger::* instead of lots of overloads		
- [x] StdFunctionDelegate for basic events		
- [x] std::thread doesn`t support priority setting
- ~~[ ] std::tuple instead of Tuple: std::tuple doesn't accept partial initialization~~
- ~~[ ] Semaphore: simulated with mutex and condition variable, but platform specific should be better~~
- ~~[ ] RWLock: std::shared_timed_mutex has separate read and write unlock, but RWLock uses unlock for both~~
- ~~[ ] Use <random> for Poco/Random, but it supports 4 different modes that must be mapped~~
- [x] replace JSON parser
- [ ] Crypto PKCS12 and ECC support


aleks-f:
@obiltschnig this is initial attempt, probably there will be changes before merging; in any case, I propose to squash the merge into develop

RangelReale:
There is a cpp11_changes.txt
<https://github.com/pocoproject/poco/pull/1795/files#diff-fb19503442f6a4845175edbd07e14287>
file
on the root folder which lists all the c++11 changes I made, and also the
potential TODO and problems.

aleks-f:
@obiltschnig I'm trying to translate the list mentioned above to the project cards but for some reason it only allowed me to do one, now it says "awaiting triage", whatever that means, not sure if it has to do with project ownership or something

another thing, I think we could drop the c++11 macro and only carry on c++14/17 for 2.0?

aleks-f:
worked now, [project](https://github.com/pocoproject/poco/projects/1) updated

aleks-f:
this is going to be merged into develop soon, the outstanding items on the list will likely be turned into issues for post-2.0 release

aleks-f:
@zosrothko do you know why this is happening on travis?
```
E: Failed to fetch http://apt.llvm.org/trusty/dists/llvm-toolchain-trusty/main/source/Sources  404  Not Found [IP: 52.11.249.247 80]
```

zosrothko:
@aleks-f It seems that LLVM decided to remove all pre 4.0 version of clang from thair distribution web site
```
gcc backport (ppa) is necessary on Trusty (for libstdc++).
Precise, Quantal, Raring, Saucy and Utopic are no longer supported by Ubuntu.

Trusty (14.04) - Last update : Mon, 04 Sep 2017 06:34:06 UTC / Revision: 312462

deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty main
deb-src http://apt.llvm.org/trusty/ llvm-toolchain-trusty main
# 4.0 
deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-4.0 main
deb-src http://apt.llvm.org/trusty/ llvm-toolchain-trusty-4.0 main
# 5.0 
deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-5.0 main
deb-src http://apt.llvm.org/trusty/ llvm-toolchain-trusty-5.0 main

# Also add the following for the appropriate libstdc++
deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main, so that only the 4.0, 5.0 and upper version remain.
```

Issue #1373, opened 2016-08-30
SessionImpl::close() does not check return code of close handle specific function 


micheleselea:
When in `void SessionPool::closeAll(SessionList& sessionList)` we call `(*it)->session()->close();` the correspond function of SessionImpl is called.
About SQLite::SessionImpl we have this code

```
void SessionImpl::close()
{
    if (_pDB)
    {
        sqlite3_close(_pDB);
        _pDB = 0;
    }
        _connected = false;
}
```

accorling to this docs
https://www.sqlite.org/c3ref/close.html
the function `sqlite3_close(_pDB);` can fail with SQLITE_BUSY. So we set NULL to a not closed HANDLE.
I think we should check if `sqlite3_close(_pDB);` is working as we expect, because if you have for example an in memory db, you will not obtain the "memory clear" you expect when you close session. What do you think?

I do not know in other backend (like MySql) how it works

obiltschnig:
Maybe sqlite3_close() should be wrapped in a retry loop with a Poco::Thread::sleep(100)

micheleselea:
Yes I think could be an acceptable solution. I do not know if mysql backend or ODBS have same issue

PR #1800, opened 2017-07-10
Poco Path and simbol $

bas524:
bug : into current implementation od path_unix.cpp I can't use simbol $ like a simbol, not a variable prefix, for example,
I need to use path /home/user/O1$$/folder. I propose to use escape sequence for this, for ex., if I want to use Poco::Glob, I can do 
```
std::set<std::string> files;
std::string path = "/home/user/O1$$/folder"
Poco::replaceInPlace(path, "$", "\\$");
Poco::Glob::glob(path, files);
```

### DIFF
--- a/Data/SQLite/src/SessionImpl.cpp
+++ b/Data/SQLite/src/SessionImpl.cpp
@@ -205,9 +205,14 @@ void SessionImpl::open(const std::string& connect)
 
 void SessionImpl::close()
 {
-	if (_pDB)
-	{
-		sqlite3_close(_pDB);
+	if (_pDB) {
+		int result = 0;
+		do {
+			result = sqlite3_close(_pDB);
+			if (result == SQLITE_BUSY) {
+				Poco::Thread::sleep(100);
+			}
+		} while (SQLITE_BUSY == result);
 		_pDB = 0;
 	}
 
--- a/Data/SQLite/src/SessionImpl.cpp
+++ b/Data/SQLite/src/SessionImpl.cpp
@@ -214,21 +214,24 @@ void SessionImpl::close()
 			result = sqlite3_close(_pDB);
 			if (result == SQLITE_BUSY) 
 			{
-				sqlite3_stmt *pStmt = NULL;
-				do 
-				{
-					pStmt = sqlite3_next_stmt(_pDB, NULL);
-					if (pStmt && sqlite3_stmt_busy(pStmt)) 
-					{
-						sqlite3_finalize(pStmt);
-					}
-				}
-				while (pStmt != NULL);
-				Poco::Thread::sleep(100);
 				--times;
 			}			
 		} 
 		while (SQLITE_BUSY == result && times > 0);
+		if (SQLITE_BUSY == result && times == 0) 
+		{
+			sqlite3_stmt *pStmt = NULL;
+			do 
+			{
+				pStmt = sqlite3_next_stmt(_pDB, NULL);
+				if (pStmt && sqlite3_stmt_busy(pStmt)) 
+				{
+					sqlite3_finalize(pStmt);
+				}
+			}
+			while (pStmt != NULL);
+			sqlite3_close(_pDB);		
+		}
 		_pDB = 0;
 	}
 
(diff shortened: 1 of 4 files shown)


bas524:
Hello, please look at the AppVeyor build, it contain errors with odbc test building, not with my commit

zosrothko:
The CI AppVeyor & Travis are in a unstable state because there is a switch to the c++11 standard for the develop branch on the way. This should be fixed sooner or later.

Data/SQLite/src/SessionImpl.cpp
```diff
@@ -207,12 +207,14 @@ void SessionImpl::close()
 {
 	if (_pDB) {
 		int result = 0;
+		int times = 50;
 		do {
```
aleks-f: please check [Coding Style Guide](https://www.appinf.com/download/CppCodingStyleGuide.pdf)

Foundation/src/Path_UNIX.cpp
```diff
@@ -120,7 +120,15 @@ std::string PathImpl::expandImpl(const std::string& path)
...
 	{
-		if (*it == '$')
+		if (*it == '\\')
+		{
+			++it;
+			if (*it == '$')
+			{
+				result += *it++;
+			}
+		}
```
aleks-f: please add a test case for this feature

bas524:
Ok, I fixed codestyle and add testcase for Path::expand

Foundation/src/Path_UNIX.cpp
bas524: PLease, review my changes

Data/SQLite/src/SessionImpl.cpp
```diff
@@ -205,16 +205,20 @@ void SessionImpl::open(const std::string& connect)
...
+	if (_pDB) 
+	{
 		int result = 0;
 		int times = 50;
-		do {
+		do 
+		{
 			result = sqlite3_close(_pDB);
-			if (result == SQLITE_BUSY) {
+			if (result == SQLITE_BUSY)
```
aleks-f: Looping limited number of times prevents endless loop but does not necessarily fix the root cause for the busy state. If this is happening, something is [still opened](http://www.sqlite.org/c3ref/close.html) - likely a prepared statement, which should be finalized (or should not be lingering to start with).

bas524: So, if I want to close session correctly, then I must to have list of all statements, and into close method all statements must be destroyed before sqlite3_close. And every statement  into destuctor must erase itself from session which owned statements-list. Am I right?

aleks-f: you can loop through all prepared statements using
```
sqlite3_next_stmt(_pDb, 0);
```
and then finalize each.
Note that there may be other `SQLITE_BUSY` on close reasons, like opened blob handles or backups. a proper solution would be to simulate such conditions and make sure they are handled correctly; preferably, in a separate pull.

bas524:
As for sessionimpl->close... I add finalize for every busy-stmt, but I can't get opened blobs and backups now ...

aleks-f:
includes fix for #1373

zosrothko:
@bas524 The test below is failing on Windows. can youy fix it?
```
void PathTest::testExpandUnix()
{
	std::string pathWithoutVar = "/usr/share/O1\\$\\$/folder";
	std::string pathWithVar = "${HOME}/folder";
	Path p;
	std::string s = p.expand(pathWithoutVar);
	assert (s == "/usr/share/O1$$/folder");
	s = p.expand(pathWithVar);
	Path tmpPath = Path::home();
	tmpPath.append("folder");
	assert (s == tmpPath.toString());
}
```
```
C:\Users\fandre\Documents\git\poco-develop>build\script\runtests2.cmd Foundation testExpandUnix


*** Foundation
testExpandUnix: ===================
p=/usr/share/O1$$/folder
s=/usr/share/O1\$\$/folder
FAILURE
!!!FAILURES!!!
Runs: 1   Failures: 1   Errors: 0

There was 1 failure:
 1: class CppUnit::TestCaller<class PathTest>.testExpandUnix
    "s == "/usr/share/O1$$/folder""
    in "src\PathTest.cpp", line 388
1 runs, 1 failed.
```

bas524:
Yes, shure.

bas524:
Complete. Request #1850, please review it

PR #1809, opened 2017-07-12
make RecordSet::reset(stmt) return a reference to itself

alterratz:
Statement::reset returning a self-reference is very convenient.
Lets make RecordSet::reset do the same  :-)


obiltschnig:
There are possibly other methods where this could make sense too.

PR #1825, opened 2017-07-25
SQlite 3.19.3 and Socket::select fix

micheleselea:
changelog
http://www.sqlite.org/releaselog/3_19_3.html

### DIFF
--- a/Net/src/Socket.cpp
+++ b/Net/src/Socket.cpp
@@ -80,8 +80,7 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 
 	int epollfd = -1;
 	{
-		struct epoll_event eventsIn[epollSize];
-		memset(eventsIn, 0, sizeof(eventsIn));
+		struct epoll_event eventsIn[epollSize] = { 0 };		
 		struct epoll_event* eventLast = eventsIn;
 		for (SocketList::iterator it = readList.begin(); it != readList.end(); ++it)
 		{
@@ -166,8 +165,7 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 		}
 	}
 
-	struct epoll_event eventsOut[epollSize];
-	memset(eventsOut, 0, sizeof(eventsOut));
+	struct epoll_event eventsOut[epollSize] = { 0 };	
 
 	Poco::Timespan remainingTime(timeout);
 	int rc;
--- a/Net/src/Socket.cpp
+++ b/Net/src/Socket.cpp
@@ -79,8 +79,10 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 	if (epollSize == 0) return 0;
 
 	int epollfd = -1;
-	{
-		struct epoll_event eventsIn[epollSize] = { 0 };		
+	{		
+		struct epoll_event eventsIn[epollSize];
+		memset(eventsIn, 0, sizeof(epoll_event) * epollSize );
+		
 		struct epoll_event* eventLast = eventsIn;
 		for (SocketList::iterator it = readList.begin(); it != readList.end(); ++it)
 		{
@@ -165,7 +167,8 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 		}
 	}
 
-	struct epoll_event eventsOut[epollSize] = { 0 };	
+	struct epoll_event eventsOut[epollSize];
+	memset(eventsOut, 0, sizeof(epoll_event) * epollSize );	
 
 	Poco::Timespan remainingTime(timeout);
 	int rc;
--- a/Data/SQLite/src/sqlite3.h
+++ b/Data/SQLite/src/sqlite3.h
@@ -114,16 +114,16 @@ extern "C" {
 ** system</a>.  ^The SQLITE_SOURCE_ID macro evaluates to
 ** a string which identifies a particular check-in of SQLite
 ** within its configuration management system.  ^The SQLITE_SOURCE_ID
-** string contains the date and time of the check-in (UTC) and an SHA1
-** hash of the entire source tree.
+** string contains the date and time of the check-in (UTC) and a SHA1
+** or SHA3-256 hash of the entire source tree.
 **
 ** See also: [sqlite3_libversion()],
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.17.0"
-#define SQLITE_VERSION_NUMBER 3017000
-#define SQLITE_SOURCE_ID      "2017-02-13 16:02:40 ada05cfa86ad7f5645450ac7a2a21c9aa6e57d2c"
+#define SQLITE_VERSION        "3.19.3"
+#define SQLITE_VERSION_NUMBER 3019003
+#define SQLITE_SOURCE_ID      "2017-06-08 14:26:16 0ee482a1e0eae22e08edc8978c9733a96603d4509645f348ebf55b579e89636b"
 
 /*
 ** CAPI3REF: Run-Time Library Version Numbers
@@ -857,7 +857,7 @@ struct sqlite3_io_methods {
 ** opcode allows these two values (10 retries and 25 milliseconds of delay)
 ** to be adjusted.  The values are changed for all database connections
 ** within the same process.  The argument is a pointer to an array of two
-** integers where the first integer i the new retry count and the second
+** integers where the first integer is the new retry count and the second
 ** integer is the delay.  If either integer is negative, then the setting
 ** is not changed but instead the prior value of that setting is written
 ** into the array entry, allowing the current retry settings to be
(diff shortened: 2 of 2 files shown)


micheleselea:
Working with this SQlite version since June, and no problem in production environment

micheleselea:
about epoll_event memset fix: with -std=c++14 sizeof(eventsOut) gives wrong result, so memset fail. This issue affect only Linux compiled in c++14 with epoll.
The function is used only by socketreactor, that can cause a while(1) in the thread that does select request

aleks-f:
it fails to compile on 4.8:
```
[0m/home/travis/build/pocoproject/poco/Net/src/Socket.cpp: In static member function â€˜static int Poco::Net::Socket::select(Poco::Net::Socket::SocketList&, Poco::Net::Socket::SocketList&, Poco::Net::Socket::SocketList&, const Poco::Timespan&)â€™:
/home/travis/build/pocoproject/poco/Net/src/Socket.cpp:83:48: error: variable-sized object â€˜eventsInâ€™ may not be initialized
   struct epoll_event eventsIn[epollSize] = { 0 };  
                                                ^
/home/travis/build/pocoproject/poco/Net/src/Socket.cpp:168:48: error: variable-sized object â€˜eventsOutâ€™ may not be initialized
  struct epoll_event eventsOut[epollSize] = { 0 }; 
                                                ^
make[2]: *** [Net/CMakeFiles/Net.dir/src/Socket.cpp.o] Error 1
make[1]: *** [Net/CMakeFiles/Net.dir/all] Error 2
```
I am surprised that it fails to calculate sizeof on c++14 and that the fix compiles at all because it's a dynamic sized array.

Anyway, could this be a fix:
```
struct epoll_event eventsIn[epollSize];
memset(eventsIn, 0, sizeof(epoll_event) * epollSize );
```

micheleselea:
Yes I'll try to put a check on c++xx version used...

I manage to remove the pull request



*Michele Pradella*
/R&D Software Engineer
/ michele.pradella@selea.com <mailto:michele.pradella@selea.com>
Office: +39 0375 889091

------------------------------------------------------------------------
	
------------------------------------------------------------------------
Selea s.r.l

V. Aldo Moro, 69
46019 Cicognara MN - Italy
Phone: +390375889091
Fax: +390375889080
http://www.selea.com

	

Note: The information contained in this message may be privileged and 
confidential and protected from disclosure. If the reader of this 
message is not the intended recipient, or an employee or agent 
responsible for delivering this message to the intended recipient, you 
are hereby notified that any dissemination, distribution or copying of 
this communication is strictly prohibited. If you have received this 
communication in error, please notify us immediately by replying to the 
message and deleting it from your computer. Thank you.

Il 25/07/2017 11.19, Aleksandar Fabijanic ha scritto:

micheleselea:
If you try you will see that sizeof(eventsIn)will give you bytes count 
with c++11 and element number in c++14

Someone find the same problem
http://wm.ite.pl/notesen/2015-03-22-compiler-warnings.html

*Michele Pradella*
/R&D Software Engineer
/ michele.pradella@selea.com <mailto:michele.pradella@selea.com>
Office: +39 0375 889091

------------------------------------------------------------------------
	
------------------------------------------------------------------------
Selea s.r.l

V. Aldo Moro, 69
46019 Cicognara MN - Italy
Phone: +390375889091
Fax: +390375889080
http://www.selea.com

	

Note: The information contained in this message may be privileged and 
confidential and protected from disclosure. If the reader of this 
message is not the intended recipient, or an employee or agent 
responsible for delivering this message to the intended recipient, you 
are hereby notified that any dissemination, distribution or copying of 
this communication is strictly prohibited. If you have received this 
communication in error, please notify us immediately by replying to the 
message and deleting it from your computer. Thank you.

Il 25/07/2017 11.19, Aleksandar Fabijanic ha scritto:

micheleselea:
@aleks-f  can I need to know a thing about pullrequest. If I have more changes to commit, should I continue commit in my current fork and create another pullrequest? or should I re fork it?

aleks-f:
this one is merged, create another fork from develop branch and send new pull

aleks-f:
generally speaking it is best to keep pulls focused on one specific thing (also create an issue if it's a bug so we can track it)

micheleselea:
ok thanks, sorry for this question but was the first time of pull request and I want to make it in the correct way. I suppose that the correct way was to split the modification one for "kind" (one for SQLite, and one for socket) but I misunderstood the good way to do it.
Tanks for informations

zosrothko:
@micheleselea I was asking to comment the Socket.cpp code itself, not this pull request! ;)

micheleselea:
ah ok understand I'll do it as soon as I can.  

sorride but was my first pullrequest since long time  

Il f2017-08-02 08:27 zosrothko ha scritto:

> @micheleselea [1] I was asking to comment the Socket.cpp code itself, not this pull request! ;) 
> 
> --
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].

 

Links:
------
[1] https://github.com/micheleselea
[2] https://github.com/pocoproject/poco/pull/1825#issuecomment-319580139
[3]

Net/src/Socket.cpp
```diff
@@ -79,9 +79,10 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 	if (epollSize == 0) return 0;
 
 	int epollfd = -1;
-	{
+	{		
 		struct epoll_event eventsIn[epollSize];
-		memset(eventsIn, 0, sizeof(eventsIn));
+		memset(eventsIn, 0, sizeof(epoll_event) * epollSize );
```
micheleselea: gcc with -std=c++11 sizeof(eventsIn) gives bytes number
gcc with -std=c++14 sizeof(eventsIn) gives elements number
changed memset to keep it save in both condition

Net/src/Socket.cpp
```diff
@@ -167,7 +168,7 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 	}
 
 	struct epoll_event eventsOut[epollSize];
-	memset(eventsOut, 0, sizeof(eventsOut));
+	memset(eventsOut, 0, sizeof(epoll_event) * epollSize );
```
micheleselea: gcc with -std=c++11 sizeof(eventsOut) gives bytes number
gcc with -std=c++14 sizeof(eventsOut) gives elements number
changed memset to keep it safe

aleks-f:
@micheleselea : the array size fix does not hurt, but it must be a g++ version oddity/bug or something like that because I can not reproduce it, eg.:

```
$ clang++ --version
Apple LLVM version 8.1.0 (clang-802.0.42)

$ cat array.cpp
#include <iostream>

int main()
{
    int sz = 10;
    int a[10], b[sz];
    std::cout << "size of array of 10 int: " << sizeof(int[10])   << '\n'
              << "size of array a of 10 int: " << sizeof a          << '\n'
              << "size of array b of 10 int: " << sizeof b << '\n';
}

$ clang++ -std=c++14 -o array14 array.cpp
$ clang++ -std=c++11 -o array11 array.cpp

$ ./array11
size of array of 10 int: 40
size of array a of 10 int: 40
size of array b of 10 int: 40

$ ./array14
size of array of 10 int: 40
size of array a of 10 int: 40
size of array b of 10 int: 40
```

```
$ g++ --version
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

$ g++ -std=c++14 -o array14 array.cpp
$ g++ -std=c++11 -o array11 array.cpp

$ ./array11
size of array of 10 int: 40
size of array a of 10 int: 40
size of array b of 10 int: 40

$ ./array14
size of array of 10 int: 40
size of array a of 10 int: 40
size of array b of 10 int: 40
```

micheleselea:
I don't know I was able to reproduce on a Debian jessie (3.16.0-4amd64) with gcc version 4.9.2-10 and on a Raspberry with DietPi 4.1.17-v7+ distribution and gcc version 4.9.2-10
I use -std=c++11 or 14 in the config.make file with POCO_CFLAGS (I'm not in office so I'm try remember best as I can).
Could be a problem of that gcc version?

zosrothko:
@aleks-f Your test is not a good one, neither the compiler. Below a relevant test
```
fandre@ubuntu-trusty-64:~/gcc$ cat array.cpp
#include <iostream>
using namespace std;

int size(int n) {
    char array[n];
    return sizeof(array);
}

int main() {
        cout << "size(5)=" << size(5) <<endl;
        return 0;
}
fandre@ubuntu-trusty-64:~/gcc$ g++-4.9 -std=c++11 array.cpp -o array11
fandre@ubuntu-trusty-64:~/gcc$ g++-4.9 -std=c++14 array.cpp -o array14
array.cpp: In function ‘int size(int)’:
array.cpp:6:24: warning: taking sizeof array of runtime bound [-Wvla]
     return sizeof(array);
                        ^
fandre@ubuntu-trusty-64:~/gcc$ ./array11
size(5)=5
fandre@ubuntu-trusty-64:~/gcc$ ./array14
size(5)=1
fandre@ubuntu-trusty-64:~/gcc$
```

zosrothko:
@micheleselea Can you reference this [url ](http://wm.ite.pl/notesen/2015-03-22-compiler-warnings.html) in your comment; Thanks

aleks-f:
@zosrothko: yes and no - I understand why compiler warns in your example - it can't see in the function that in this particular case it's dealing with a constant known at compile time. but, what is confusing to me is that a compiler would return number of elements as sizeof and array - that sounds like something that would break lots of code.

zosrothko:
@aleks-f Right and as this [paper](http://wm.ite.pl/notesen/2015-03-22-compiler-warnings.html) states, this feature has been removed from the standard, but it is kept in g++-4.9. Thus one have to deal with it since Poco should support g++ 4.9.
```
Weird? Yes and no. GCC implemented proposal N3639 for C++14 ("Runtime-sized arrays with automatic storage duration") where sizeof(VLA) is an error ("ill-formed"), but as we see authors have chosen a soft warning rather a compile error. BTW, N3639 was rejected and GCC also removed support for this proposal.
```

micheleselea:
@zosrothko ok I'll put in the code as soon I came back from holidays.
@aleks-f I think that the point here is not when or when not the size of give correct or wrong results. Probably, generally speaking, is not a good idea do sizeof(vector pointer) because the behavior could be unpredictable. I mean for example that the compiler could give you the size of the pointer (just an example) instead what we need here in memset. Doing it explicitly I think it's the best way to reach the result without need to consider compilers versions or other variables.
I will never use sizeof that way because for me it's not clear. Anyway I think the solution proposed should solve ambiguity anyway

PR #1827, opened 2017-07-26
HttpClientSession set specific proxysettings for attached socket

micheleselea:
If we have a global proxy settings and we attach an external socket to the HTTPClientSession there's no way to use different proxy settings for that connection. For example if you do not need for that connection httpproxy because is already attached to the correct and point


micheleselea:
I prefer to add extra constructor to leave the old behavior unchanged for users that may be use it that way.
Le me know what do you think about. The case is when you (for some reason) create your own tcp connection directly attached to the correct and point: so no need to use http proxy. With the global proxy settings you can't disable the proxy of an already connected socket.

micheleselea:
@aleks-f  @obiltschnig 
Next pullrequest I want to insert code for FTPSClientSession. I'm using since 1 year without problem. Which is the best way to fork project and add that function? is it better to make a new branch?

PR #1832, opened 2017-07-28
Remove CMake jobs since it is already tested on Travis

zosrothko:
AppVeyor CI jobs takes 8 hours which is quite long. Thus removing the CMake will reduce the turnaround to 6 hours. That's ok because the CMake build is already build by the Travis CI


aleks-f:
@zosrothko: it looks like the compiler version is not accurate in failing builds:
```
$ export TEST_NAME="clang 3.8 (make) bundled"

$ export CXX=clang++
$ export CC=clang
$ clang --version
clang version 3.5.0 (tags/RELEASE_350/final)
```

zosrothko:
@aleks-f Your remark about the snipet below is correct but it is a bug from Travis that is merging the first lines of the osx test config to the linux one. If you look downward the log, there is a last $CXX --version command that is correct related to the TEST_NAME

```$ export TEST_NAME="clang 3.8 (make) bundled"

$ export CXX=clang++
$ export CC=clang
$ clang --version
clang version 3.5.0 (tags/RELEASE_350/final)
```

PR #1837, opened 2017-08-05
Fix issues #1373 for ODBC

komainu8:
Dear, developers.

I tried fix issues #1373 for ODBC.
The function `SQLDisconnect(_db);` can fail with SQL_ERROR or SQL_INVALID_HANDLE.
So, I modified check return value of `SQLDisconnect(_db);`.

When SQLState is "25000" or "25501", retried `SQLDisconnect(_db);`.
The reason why retrying is done is that if SQLState is "25000" or "25501", there is a possibility that `SQLDisconnect(_db);` will succeed by retrying.

For Poco::Data::MySQL::SessionHandle::close() and  Poco::Data::PostgreSQL::SessionHandle::disconnect(), the same problem does not occur.
Because mysql_close and PQfinish don't have return value.

### DIFF
--- a/Data/ODBC/src/SessionImpl.cpp
+++ b/Data/ODBC/src/SessionImpl.cpp
@@ -406,7 +406,8 @@ void SessionImpl::close()
 	SQLINTEGER nativeErrorCode;
 	SQLSMALLINT sqlErrorMessageLength;
 	unsigned int retryCount = 10;
-	do {
+	do
+	{
 		SQLRETURN rc = SQLDisconnect(_db);
 		if (SQL_SUCCESS != rc && SQL_SUCCESS_WITH_INFO != rc)
 		{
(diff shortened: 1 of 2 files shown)


Data/ODBC/src/SessionImpl.cpp
```diff
@@ -401,7 +401,39 @@ void SessionImpl::close()
...
+			{
+				break;
+			}
+		}
+		else
+		{
+			return;
+		}
+	} while(retryCount > 0);
+
```
aleks-f: something not quite right here:

Data/ODBC/src/SessionImpl.cpp
```diff
@@ -401,7 +401,39 @@ void SessionImpl::close()
 	{
 	}
 
-	SQLDisconnect(_db);
+	SQLCHAR sqlState[5+1];
+	SQLCHAR sqlErrorMessage[SQL_MAX_MESSAGE_LENGTH];
+	SQLINTEGER nativeErrorCode;
+	SQLSMALLINT sqlErrorMessageLength;
+	unsigned int retryCount = 10;
+	do {
```
aleks-f: style

Data/ODBC/src/SessionImpl.cpp
aleks-f: the comment refers to the exception below, not the code above

komainu8:
Thank you for your comments.
I've fixed them.

Issue #1824, opened 2017-07-25
Connect Redis server with password field


kapcino:
@fbraem Can I specify a password to connect redis server?

I only find the host and port fields in Client class, how should I do when I connect a server requires password?

Thanks!

Update: Just find AUTH command, seems we can send AUTH with password after connect? I will try this method.

fbraem:
The AUTH command is indeed the solution for this.

kapcino:
Yeah, it works, thanks again @fbraem!

PR #1838, opened 2017-08-07
Add password field for Redis Client

kapcino:
@fbraem For issue #1824 I create a pull request for adding password field in Redis Client.
Please have a review, thanks.

### DIFF
--- a/Redis/include/Poco/Redis/Client.h
+++ b/Redis/include/Poco/Redis/Client.h
@@ -83,11 +83,21 @@ public:
 		/// Constructor which connects to the given Redis host/port.
 		/// The host and port must be separated with a colon.
 
+    Client(const std::string& hostAndPort, const std::string& password);
+		/// Constructor which connects to the given Redis host/port/password.
+		/// The host and port must be separated with a colon.
+
 	Client(const std::string& host, int port);
 		/// Constructor which connects to the given Redis host/port.
 
+	Client(const std::string& host, int port, const std::string& password);
+		/// Constructor which connects to the given Redis host/port/password.
+
 	Client(const Net::SocketAddress& addrs);
 		/// Constructor which connects to the given Redis host/port.
+
+    Client(const Net::SocketAddress& addrs, const std::string& password);
+		/// Constructor which connects to the given Redis host/port/password.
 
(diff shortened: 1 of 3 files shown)


kapcino:
@fbraem You mean Iike this?

```
Poco::Redis::Client redis;
redis.connect(host, port);
redis.sendAuth(password);
```

It OK for single connection, but for the connection pool, it is not convenient as above.

```
Poco::Net::SocketAddress sa(_host, _port);
Poco::PoolableObjectFactory<Client, Client::Ptr> factory(sa);
Poco::ObjectPool<Client, Client::Ptr> pool(factory, 10, 15);
PooledConnection pclient1(pool);
```

If I want to add the password for the test code above, I need to add password into class PoolableObjectFactory, the current implement for createdObject() is:

```
Redis::Client::Ptr createObject()
{
    return new Redis::Client(_address);
}
```

Like the single connection, we should change like this:

```
Redis::Client::Ptr createObject()
{
    Redis::Client::Ptr ptr =  new Redis::Client(_address);
    ptr->sendAuth(_password);
    return ptr;
}
```

But it is not as simple as current one, so I change the constructor of Client and call sendAuth.

For the password problem, I will remove it from class.

fbraem:
Maybe add a property to Client that indicates if sendAuth was called succesfully or not. That way we can check on a client that is returned from the pool if it is authenticated or not ...

kapcino:
@fbraem Changed based on your comments, please review again, thanks!

fbraem:
Sorry for my late answer, but I'm on a holiday for some weeks ...

Do we need the password in the pool class? Let the code that gets a client from the pool check if it is authenticated or not and call sendAuth if needed ... That way the password member is also not needed in the Pool class ...

kapcino:
@fbraem Updated.

fbraem:
This seems ok for me now. One question: is there a reason why you return true from sendAuth when the password is empty?

kapcino:
@fbraem it is a bug, I've changed, thanks.

fbraem:
@aleks-f : For me this enhancement is ok.
@kapcino : Thanks for your work.

PR #1849, opened 2017-08-15
Restore Mutex_STD for Cygwin since it does not support std::timed_mutex

zosrothko:


aleks-f:
@zosrothko this should go into c++11 branch, develop does not yet have Mutex_STD.h

Issue #1863, opened 2017-08-28
Unable to extract value to Poco::Dynamic::Var for MySQL database


Erbengi:
If use `Poco::Dynamic::Var` to extract value from MySQL statement then variable is cleared.

Example:
```c++
using namespace Poco::Data;
using namespace Poco::Data::Keywords;

Poco::Data::MySQL::Connector::registerConnector();
Poco::SharedPtr<Session> session = new Session("MySQL", "host=127.0.0.1;port=3306;db=db_name;user=user_name;password=pswd;");
Poco::SharedPtr<Statement> stm = new Statement(*session);
Poco::Dynamic::Var count = 0;
std::string sql("SELECT COUNT(*) FROM table_name");

*stm << sql, into(count);
stm->execute();

if(count.isEmpty())
  std::cerr << "Error: count is empty!" << std::endl;
```
Output:
```
Error: count is empty!
```

**Reason is** that implementation of method `extract` in `Poco::Data::MySQL::Extractor` is:
```c++
bool Extractor::extract(std::size_t pos, Dynamic::Var& val)
{
	return false;
}
```

**Consequence** is that class `Poco::Data::ArchiveByAgeStrategy` never archive any row in database because it uses `Poco::Dynamic::Var` to store number of rows older than given age.

aleks-f:
Yeah, MySQL was contributed to an already existing base of Poco::Data and not all features were entirely implemented. In this particular case, it involves discovering what is the underlying value native type, then call the appropriate extract and assign the value to the `Dynamic::Var`. To discover the type see [bool Extractor::extract(std::size_t, std::string&)](https://github.com/pocoproject/poco/blob/develop/Data/MySQL/src/Extractor.cpp#L132)

If you have time to contribute the fix and send a pull (fix and unit test, please), I'll gladly review and merge it. Otherwise, it will have to wait until we have some time to implement this.

Erbengi:
I implemented extraction of `Dynamic::Var` and also `Any`. I also fixed CI test on AppVeyor with cygwin.  I made pull request #1872. CI Travis failed but It doesn't seems that it was caused by my code.

PR #1872, opened 2017-09-02
Implement extraction of Dynamic::Var and Any in MySQL extractor (#1863)

Erbengi:
Implemented extraction of Dynamic::Var and Any in MySQL extractor according to issue #1863


aleks-f:
@Erbengi thank you for the contribution1

Issue #1147, opened 2016-02-05
Move to standard integer types


Burgch:
On Mac, compiling with Clang (from Xcode 6.2), `Poco::Int64` resolves to `long`, whereas `boost::int64_t` and the <cstdint> `int64_t` resolve to `long long`, which then causes issues when setting an int64_t in a DynamicVar and converting it back to int64_t, unless everything is explicitly cast to Poco::Int64 and back.

obiltschnig:
Changing the definition of Poco::Int64 and Poco::UInt64 will require a lot of changes (in Dynamic::Var and other classes), so not sure I will do this, as it could potentially break existing code.

obiltschnig:
Will handle that in 2.0.

mltucker:
Ran into this on iOS for the arm64 architecture.  And it's a big problem for us since Apple is dropping armv7 support in iOS 11!

Burgch:
I believe I have a workaround where, if `POCO_LONG_IS_64_BIT ` is defined, template specialisations for `long long` and `unsigned long long` (which `int64_t` and `uint64_t` map to on macOS) are added to `Poco::Dynamic::VarHolderImpl`, along with all of the necessary conversion functions. This seems to be passing the tests I've added and prevents the `BadCastException`s that you would see if you optimistically used `(u)int64_t` with e.g. `Poco::JSON::Object`. Would this be an acceptable addition for Poco 1.8.0?

aleks-f:
@Burgch for 1.x line of releases, yes - coordinate contribution with @obiltschnig. 
2.0 (ie. current develop branch) will be based on &lt;cstdint&gt;

aleks-f:
@obiltschnig this turned out to be more complicated than expected. basically, the stumbing block is 64 bit long on clang - as things are now, when long is 64-bit, there is no way to pass long and unsigned long to NumberFormatter and BinaryReader/Writer - only Poco::(U)Int64.

Issue #1870, opened 2017-09-01
Portability: all public Poco API  integral parameter should use Poco integral type


zosrothko:
Hello

This is a portability issue. Below is resume the length of all native integral types defined in C/C++
```
Type           ILP64   LP64   LLP64
char              8      8       8
short            16     16      16
int              64     32      32
long             64     64      32
long long        64     64      64
pointer          64     64      64
```
All modern Unix compilers and macOS compilers are following the LP64 mode while VisualStudio is following the LLP64 mode. Thus there is a portability issue for the `long` type.  

In order to insure a full portabibility of application using Poco, this issue requests that the integral parameters of all public Poco's API use `only` the corresponding Poco integral types as defined in `Poco/Types.h`

zosrothko:
@obiltschnig @aleks-f  any comment on this issue?

aleks-f:
noble goal, but not small amount of work. FWIW, Poco types are now [aligned with std types](https://github.com/pocoproject/poco/issues/1147) in develop

PR #1882, opened 2017-09-11
Add support for (unsigned) long long when long is 64bit

Burgch:
This provides a temporary workaround for #1147 by providing `long long` and `unsigned long long` template implementations when `Poco::Int64` and `Poco::UInt64` are `#defined` as `long` and `unsigned long`. This prevents `BadCastException`s when trying to set and extract `int64_t` and `uint64_t` from `Poco::JSON::Object`s. Additional test cases have been added to the relevant tests, and pass on Windows, Mac and Linux (Mac being the key, as it is where the original issue is highlighted by `Poco::Int64` being `long` whilst `int64_t` is `long long`).

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -187,6 +187,16 @@ public:
 	void convert(unsigned long& val) const;
 		/// Calls convert(UInt32).
 
+#else
+
+	virtual void convert(long long& val) const;
+		/// Throws BadCastException. Must be overriden in a type
+		/// specialization in order to suport the conversion.
+
+	virtual void convert(unsigned long long & val) const;
+		/// Throws BadCastException. Must be overriden in a type
+		/// specialization in order to suport the conversion.
+
 #endif
 
 	virtual void convert(bool& val) const;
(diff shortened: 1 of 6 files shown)


zosrothko:
@Burgch May be you should also take into consideration this issue [#1870](https://github.com/pocoproject/poco/issues/1870)

Burgch:
@zosrothko I don't think it makes any difference? `long long` is 64 bit in all three schemes, so providing specialisations for it which assume 64 bit is safe, and providing those implementations prevents runtime exceptions when users of Poco expect it to work with the standard types (which is the motivation for this pull request).

zosrothko:
@Burgch Yes, it will make a difference because all those APIs with native type like `long long` will be removed and replaced by only APIs with Poco's defined types.

aleks-f:
@zosrothko this pull is for 1.8 branch - it will not go into develop and 2.0. So, I'd vote yes here and then consider #1870 for 2.0 and forward, alongside mapping poco integral types to the standard ones

zosrothko:
@aleks-f ok...

aleks-f:
@obiltschnig merge this for 1.8 release?

PR #1885, opened 2017-09-12
If openssl > 1.0.2, use it to verify cert matches hostname. Fix GH #1817

### LINKED ISSUE #1817: Wildcard Certificate verification too lenient?
I've been playing with certificates from [Bad SSL](https://badssl.com/), and noticed that the logic used in `Poco::Net::X509Certificate::matchWildcard` seems to be more lenient than the recommendations made in [RFC 6125](https://tools.ietf.org/html/rfc6125#section-6.4.3) - specifically it allows [wrong.host.badssl.com](https://wrong.host.badssl.com/) for a certificate `*.badssl.com`.

Is the Poco implementation intended to be RFC compliant? If so, would you consider using (at least in the `NetSSL_OpenSSL` module) the [X509_check_host](https://www.openssl.org/docs/man1.0.2/crypto/X509_check_host.html) implementation introduced in OpenSSL 1.0.2? Happy to look into creating a Pull Request if so.

Burgch:
This provides an implementation and tests for using OpenSSL's `X509_check_host` and `X509_check_ip_asc` if Poco is being compiled against a version >= 1.0.2, or falling back to the previous implementation if not. The test covers both cases, and the different behaviour between them. Fixes #1817. Note that applications which link against Poco when it has been compiled with a newer version of OpenSSL also need to link against that newer version at both build time and runtime.

### DIFF
--- a/NetSSL_OpenSSL/src/X509Certificate.cpp
+++ b/NetSSL_OpenSSL/src/X509Certificate.cpp
@@ -5,7 +5,7 @@
 // Package: SSLCore
 // Module:  X509Certificate
 //
-// Copyright (c) 2006-2009, Applied Informatics Software Engineering GmbH.
+// Copyright (c) 2006-2017, Applied Informatics Software Engineering GmbH.
 // and Contributors.
 //
 // SPDX-License-Identifier:	BSL-1.0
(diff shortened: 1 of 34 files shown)


aleks-f:
thanks, just for future contributions - VS projects older than VS2015 are going away so no need to bother :)

Burgch:
Thanks, good to know!

Burgch:
Is there a way to also get this into the 1.8.0 branch?

obiltschnig:
yes, should eventually get there.

Issue #1897, opened 2017-09-21
DateTime wrong binding/extraction for MySQL database


Erbengi:
`DateTime` is not bind/extract correctly from MySQL database. Milisecond and microseconds are corrupted.

### Binding
Miliseconds are bind by this code:

https://github.com/pocoproject/poco/blob/238306a6ed70a7c438d4535c64110ddf90f9e0a4/Data/MySQL/src/Binder.cpp#L172

But `MYSQL_TIME::second_part` contains both milisecond and microsecond as one number. So the previous code makes microseconds from miliseconds. For example 0.333555 seconds (333 miliseconds and 555 microseconds) is bind as `mt.second_part = 333` (it means 333 microseconds), but it should be bind as `mt.second_part = 333555`.

### Extraction
Extraction is done by code:

https://github.com/pocoproject/poco/blob/238306a6ed70a7c438d4535c64110ddf90f9e0a4/Data/MySQL/src/Extractor.cpp#L211

Assigment `mt.second_part` to `DateTime` causes assertion violation, because miliseconds in `DateTime` must be smaller or equal 999. But `mt.second_part` may contains much bigger values.

PR #1898, opened 2017-09-21
Fixed binding and extraction of DateTime for MySQL database #1897

Erbengi:
Fixed `DateTime` binding/exctraction for MySQL database #1897


aleks-f:
Thank you for this fix - I will merge it as soon as we resolve its relationship to #1899

PR #1899, opened 2017-09-21
Bugfix/mysql datetime

Erbengi:
Fixed `DateTime` binding/extraction for MySQL database #1897


aleks-f:
@Erbengi is this a duplicate of #1898 ?

Erbengi:
@aleks-f it is almost the same as #1898 but this one is based on branch poco-1.8.0 and it differs in tests. There is two more tests in develop which I changed according to my fix. I wasn't sure if the change will be merged also to develop so I created two commits - one based on develop and one based on poco-1.8.0. 
The common merge base of poco-1.8.0 and develop is very far in history and it not seems that commits from poco-1.8.0 are merged back to develop. If you think that it is not good idea to have two almost same commits, feel free to reject one of the pull requests.

aleks-f:
develop and 1.8 will never be merged - 1.8 is a dead-end, it will be the last release of 1.x series and develop will become 2.0. unfortunately, that means that branches have to be maintained in parallel. So, I'm closing this, will merge 1898, and whatever changes you have to 1.8 branch, please send pulls there

PR #1909, opened 2017-09-27
CMake: use C11 as standard for all targets

bachp:
This makes the JSON module compile with GCC < 5 where the default was still C90


bachp:
@aleks-f This should fix the travis errors here: https://travis-ci.org/pocoproject/poco/jobs/280454653

PR #1911, opened 2017-09-28
Fix in close and open

micheleselea:
- close(): fix possible throw on broken socket that could lead in CLOSE_WAIT leak (especially under linux OS). I moved variable that keep track of logged in and server ready before the instruction that can throw

- open(): fix open function to deal with control socket created from outside and login without user. Do not recreate the controlsocket if it's already done and just be sure we read the server connection reply

### DIFF
--- a/Net/src/FTPClientSession.cpp
+++ b/Net/src/FTPClientSession.cpp
@@ -204,9 +204,12 @@ void FTPClientSession::close()
 	try { logout(); }
 	catch(...){ }
 	_serverReady = false;
-	_pControlSocket->close();	
-	delete _pControlSocket;
-	_pControlSocket = 0;	
+	if (_pControlSocket)
+	{
+		_pControlSocket->close();
+		delete _pControlSocket;
+		_pControlSocket = 0;
+	}
 }
 
 
--- a/Net/src/FTPClientSession.cpp
+++ b/Net/src/FTPClientSession.cpp
@@ -75,7 +75,13 @@ FTPClientSession::FTPClientSession(const std::string& host,
 {
 	_pControlSocket->setReceiveTimeout(_timeout);
 	if (!username.empty())
+	{
 		login(username, password);
+	}
+	else
+	{
+		receiveServerReadyReply();
+	}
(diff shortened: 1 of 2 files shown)


aleks-f:
@micheleselea so far, travis is failing on Net and NetSSL tests:
```
testLogin1: Aborted (core dumped)
failed=NetSSL_OpenSSL
```

micheleselea:
I'll check it tomorrow

micheleselea:
in close() function we should check if controlsocket is null because if you call close from the outside and than close is called from the destructor you can end up with null->close();

micheleselea:
@aleks-f  it's quite cryptic to me the travis result....what's going wrong?

zosrothko:
@micheleselea Below is the result of running the Net testsuite
```

testSocketConnectorTimeout: 
testLogin1: 
testLogin2: 
testLogin3: 


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
while a complete Net testsuite produces
```
testSocketReactor: 
testParallelSocketReactor: 
testSocketConnectorFail: 
testSocketConnectorTimeout: 
testLogin1: 
testLogin2: 
testLogin3: 
testLoginFailed1: 
testLoginFailed2: 
testCommands: 
testDownloadPORT: 
testDownloadEPRT: 
testDownloadPASV: 
testDownloadEPSV: 
testUpload: 
testList: 
testDownload: 
testList: 
testUserInfo: 
testPasswordProvider: 
testMissingPasswordProvider: 
```

This means that one test running after the lastest displayed `testLogin3: ` is not producing any output while it should. Probably beacause of a loop or waiting on a mutex/semaphore to be waked up.

You can run the full suite of unittests on your machine with
```
 travis/Linux/runtests.sh
```

micheleselea:
ok @zosrothko I'm going o check

micheleselea:
ok understand, is just a try catch missing, but it's correct because the reply is wrong, but it's different from previews behavior...I'll change to respect the previous behavior

PR #1918, opened 2017-10-03
Add support for sign and verify certification in crypto

Bjoe:
Some time ago, I added support for signing and verification of x509 certificate in crypto and I wanted to contribute to poco.
To see how signing and verification works, see in CryptoTest. I add and extend there the tests for signing and verification of x509 certificate.
I develop, build and test on Ubuntu 17.04 with clang and gcc. As build system I use cmake.

I read C++ Coding Style Guide carefully and I hopefully met all rules. Please review.
Also the API. 
- X509Crl, X509Request and X509Certificate are very similar. I was thinking to create one base class for this, or maybe a template or interface.
- I'm not sure if X509Store is the right class name. I use for all the same name from OpenSSL. Maybe X509Verify or X509Chain is better?

Comments, fixes, test, ideas are very welcome.

### DIFF
--- a/Crypto/src/X509Certificate.cpp
+++ b/Crypto/src/X509Certificate.cpp
@@ -33,6 +33,12 @@ namespace Poco {
 namespace Crypto {
 
 
+X509Certificate::X509Certificate():
+	_pCert(X509_new())
+{
+	X509_set_version(_pCert, 2);
+}
+
 X509Certificate::X509Certificate(std::istream& istr):
 	_pCert(0)
 {
@@ -78,27 +84,40 @@ X509Certificate::X509Certificate(const X509Certificate& cert):
 	_issuerName(cert._issuerName),
 	_subjectName(cert._subjectName),
 	_serialNumber(cert._serialNumber),
-	_pCert(cert._pCert)
+	_pCert(0)
 {
-	_pCert = X509_dup(_pCert);
+	duplicate(cert._pCert, &_pCert);
+}
+
+X509Certificate::X509Certificate(X509Certificate&& other) :
+	_issuerName(),
+	_subjectName(),
+	_serialNumber(),
+	_pCert(0)
+{
+	swap(*this, other);
 }
 
 
-X509Certificate& X509Certificate::operator = (const X509Certificate& cert)
+X509Certificate& X509Certificate::operator=(X509Certificate other)
 {
-	X509Certificate tmp(cert);
-	swap(tmp);
+	swap(*this, other);
 	return *this;
 }
 
 
-void X509Certificate::swap(X509Certificate& cert)
+void X509Certificate::swap(X509Certificate &first, X509Certificate &second)
 {
 	using std::swap;
-	swap(cert._issuerName, _issuerName);
-	swap(cert._subjectName, _subjectName);
-	swap(cert._serialNumber, _serialNumber);
-	swap(cert._pCert, _pCert);
+	swap(first._issuerName, second._issuerName);
+	swap(first._subjectName, second._subjectName);
+	swap(first._serialNumber, second._serialNumber);
+	swap(first._pCert, second._pCert);
+}
+
+void X509Certificate::swap(X509Certificate& cert)
+{
+	swap(cert, *this);
 }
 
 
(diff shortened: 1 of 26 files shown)


Crypto/src/X509Crl.cpp
```diff
@@ -0,0 +1,347 @@
+//
+// X509Crl.cpp
```
aleks-f: since CRL is an acronym, it should be capitalized

Crypto/src/X509Certificate.cpp
```diff
@@ -215,12 +221,44 @@ void X509Certificate::init()
 }
 
 
+void X509Certificate::setVersion(long ver)
+{
+	X509_set_version(_pCert, ver - 1);
```
aleks-f: We should not set this indiscriminately. In fact, I'm not even sure this should be exposed to end-user - what is the benefit of allowing user to set this, as opposed to us forcing it internally?

aleks-f:
hey, thank you for the contribution, @Bjoe :). I skimmed over it and it looks mainly good, a couple of comments for now, will look into more detail soon.

I think having a template for all X509 would probably be the best thing to do. Did not have time to check, but if the only thing they differ in is the underlying type (or at least we won't have to do a lot of type reflection in the template), then yes.

Since you did not run checks on windows, let's give apveyor some time to build and run tests - regarding those windows builds,, please get in touch with @zosrothko, they sometimes hang or take awful long time to complete

aleks-f:
@Bjoe can you add new files to [Crypto](https://github.com/pocoproject/poco/blob/develop/Crypto/Makefile) and [testsuite](https://github.com/pocoproject/poco/blob/develop/Crypto/testsuite/Makefile) makefiles?
obiovusly, windows non-cmake build will not succeed, either. I will update the VS projects once we have the design agreed upon. i'm thinking this may be a way to do it:

rename X509Certificate, CRL, etc.. - add `*Impl` to class names, and introduce template and typedefs
```
class X509<typename X>
{
// ...
}

typedef X509<X509CertificateImpl> X509Certificate
//...
```

I did not try it; hopefully, compilers will not bark at this and there may be other obstacles to this design that are not immediately obvious.

Bjoe:
Hi @aleks-f . Many thanks for your quickly review and comments. Also for your idea to introduce a template for X509 stuff.

Here my todo plan:

* First I will fix the test build issues.
* After test build is green again, I should fix review findings.
* After fixing findings etc. (hopefully this week), I will start with the refactoring of the X509 stuff and try to introduce a template.

Some site note:
I'm enthusiastic C++ developer (Part of C++ user group in NRW and meeting C++ member) and I love and used the newest C++1x features. I should first use the still old C++03 standard, because I need this feature here for my BB10/QNX project and there is no C++11 support :-(.
I see that you mark this feature for 2.0.0 milestone and I think that is a good idea and will be feasible. But I hope there will be also another 1.x.x release with this feature because 2.x.x is only build with a new C++11 compiler. I know that the effort will be become bigger, but I think it will not be for a year.
Long talk short, after all is done, I will implement C++11/1x features.

aleks-f:
@Bjoe [1.8.0](https://github.com/pocoproject/poco/tree/poco-1.8.0) will be released soon. I've ported all the recent Crypto changes there, anything [C++11 ifdef'd](https://github.com/pocoproject/poco/blob/poco-1.8.0/Crypto/include/Poco/Crypto/EVPPKey.h#L87) (although there's no move operations for X509, but it probably would be good to have), so it will all compile for older compilers.

once you're done here, port your changes there and you'll be fine for your target. If you don't make it for this release, there will be 1.8.x releases in the future.

Bjoe:
Hm OSX build failed. Compile error in src/PatternFormatter.cpp:80. I haven't change anything there. Anyways, I try to fix it, but the weird thing is, before there is a warning in src/NumberFormatter.cpp:384: argument has typ 'UIntPtr' (aka 'unsigned long long'). How can that be when only in Types.h:61 UIntPtr is unsigned long long. This means _WIN64 is defined, I would expect this is not defined in OSX. It is possible to have access to the travis ci to get files or add some preprocessor debug output?

aleks-f:
> OSX build failed

don't worry about that, I broke that build  and you probably merged it; will be fixed soon

aleks-f:
@Bjoe can you please update [testsuite Makefile](https://github.com/Bjoe/poco/blob/develop/Crypto/testsuite/Makefile)?

Also, see [this patch](https://github.com/pocoproject/poco/commit/c6cf02c9fed95a63bfd38ad0eaa244a0db537c46#commitcomment-24935751) or Windows build.

Bjoe:
@aleks-f sorry I didn't see your comment on commit c6cf02c. I only check if somebody add a comment here in the pull request. Why I didn't see here in the pull request the comment on commit c6cf02c ?

Question, which versions of OpenSSL will Poco still support? 
In continuous integration OpenSSL 1.0.1f 6 Jan 2014 is installed on Ubuntu 14.04. 
OpenSSL support for version 1.0.1 ended on 31st December 2016. Support for versions 0.9.8 and 1.0.0 ended on 31st December 2015. Those versions are no longer receiving security updates.
Ubuntu 14.04 support will end 2019, I think.

aleks-f:
@Bjoe not sure about comment updates. As for openssl, we support 1.0.0 to 1.1.x; @zosrothko maintains the CI environments.

Bjoe:
@zosrothko  I didn't see any error in travis. Last line was only: The job exceeded the maximum time limit for jobs, and has been terminated.
This means, the build job is killed by travis? Test execution take to long?

aleks-f:
@Bjoe yes, I restarted the timed out jobs. some jobs occasionally exceed the allowed time

Bjoe:
@aleks-f Ok. I added some new tests in Crypto. Maybe this needs some more minutes then before. Hm how we can fix this?

aleks-f:
more tests won't help it for sure, but it has been happening for a while even without those. not sure if anything can be done, i think it is a hard travis limit. @zosrothko ?

zosrothko:
On Travis, there are some weird issues like 
````
E: Failed to fetch http://apt.llvm.org/trusty/pool/main/l/llvm-toolchain-5.0/libllvm5.0_5.0~svn315836-1~exp1_amd64.deb  404  Not Found

E: Failed to fetch http://apt.llvm.org/trusty/pool/main/l/llvm-toolchain-5.0/libclang1-5.0_5.0~svn315836-1~exp1_amd64.deb  404  Not Found
````
This happens sometime for an unknown reason. The only solution is to restart the job with the right top button.

Issue #1093, opened 2015-12-19
CMake: use embedded openssl by default


zosrothko:
Hi

The CMake FindOpenSSL.cmake is looking for default location as C:\OpenSSL-Win32 & C:\OpenSSL-Win64 and if found, use those locations. This precludes the usage of the internal openssl libs which are now packaged with the poco repo.

Thus the CMake  build should use by default the embedded openssl libs/includes except if a flag like ENABLE_EXTERN_OPENSSL=ON specifies to use the standard FindOpenSSL.cmake

bachp:
I think there are more possibilies:

1. Use external OpenSSL
2. Use bundled OpenSSL
3. Don't use OpenSSL --> Disable functionality depending on OpenSSL

These should be reflected by the options.

aleks-f:
If I understand it right, CMake generates it's  own VS projects, so there would be no conflict with progen-generated ones (which use bundled SSL by default) from that standpoint. I would, however, follow the same logic we have in a CMake-less build, i.e. use bundled by default, unless user specifies otherwise. 

Note that, for any unbundled SSL build on windows, [POCO_EXTERNAL_OPENSSL](https://github.com/pocoproject/poco/blob/develop/Crypto/include/Poco/Crypto/Crypto.h#L106) must be defined, otherwise bundled will be linked automatically.

This only applies for Windows, on other platforms there is no bundled SSL. The rationale for bundled openssl was that Crypto and NetSSL projects always had to be edited in order to build, which is quite annoying - we wan't to build out-of-the-box whenever reasonably possible.

zosrothko:
An additional issue about openssl on Windows: When building poco with cmake in a  x64 environement, FindOpenSSL.cmake is chosing the Win32 directory when there is no C:\OpenSSL-Win64 directory

`-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/x86_amd64/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Poco package version: 1.7.0
-- Found message compiler: C:/Program Files (x86)/Windows Kits/8.1/bin/x86/mc.exe
-- Building dynamic libraries
-- Building with unittests & samples
[...]

PR #1919, opened 2017-10-04
Readd CMake build for some time

zosrothko:


aleks-f:
CMake: use embedded openssl by default #1093

PR #1931, opened 2017-10-10
Fix writing into closed socket from streambuf

villytiger:
In case of error occured in writeToDevice pptr may become one byte
farther than epptr. This can lead to crash in streambuf::xsputn from
libstdc++.


villytiger:
May be I have chosen wrong target branch. I'd like to see this fix in 1.8.0.

aleks-f:
you can send send same pull for 1.8.0 branch

Issue #1950, opened 2017-10-22
Net Exception: Address family not supported with clang


Bjoe:
I write a simple HTTP client to download a file and I get following exception:  
> Net Exception: Address family not supported

I try to find out the root cause of these error because my code was working.
I found following weird bug:

I print following member stack addresses:
* _counter member in RefCountedObject ->  poco/Foundation/src/RefCountedObject.cpp:26
* _addr member in IPv4SocketAddressImpl -> poco/Net/src/SocketAddressImpl.cpp:63
* _addr member in IPv4SocketAddressImpl addr function -> poco/Net/include/Poco/Net/SocketAddressImpl.h:103

(see also my https://github.com/Bjoe/poco.git Branch: stack_address_issue) 

and I get following result:

 RefCountedObject::RefCountedObject() Stack address from _counter:             **0x2639bc8**
 IPv4SocketAddressImpl::IPv4SocketAddressImpl() Stack address from _addr: **0x2639bcc**
 IPv4SocketAddressImpl::addr() Stack address from _addr:                               **0x2639bc8**
 
 

WTH... why is the stack address from _addr in function addr the same stack addr in _counter **0x2639bc8** ?

I try some compiler and it looks like it only happens with clang and -std=XXX argument without any optimization:

clang++-3.8 -std=c++11
clang++-4.0 -std=c++11
clang++-3.8 -std=c++14
clang++-4.0 -std=c++14

If I move the function
```
inline const struct sockaddr* IPv4SocketAddressImpl::addr() const
````
from the header in the source file it works.

Have somebody an idea?
How looks like on android? They use also clang there.
Is this really a compiler bug?
Looks like on FreeBSD, they use clang as compiler, they have also this issue, see #1684 

Here my code to test:
```
#include <Poco/Net/SocketAddress.h>
int main()
{
  Poco::Net::SocketAddress sa("localhost", 80);
  const sockaddr* s = sa.addr();
  return 0;
}

```
with my branch:
https://github.com/Bjoe/poco.git Branch: stack_address_issue

aleks-f:
@Bjoe that's with 1.7.9, right?

aleks-f:
@Bjoe I can't reproduce it, tried with clang 3.8 debug build. Somehow, there seems to be a compile-time mixup and AutoPtr ends up being mixed with SOO mode, then you have class inheriting from RefCountedObject, and the first data member is _counter. That's the only explanation I have for that behavior.

[...]

PR #1959, opened 2017-10-26
Fix issue #1950 Address familiy not supported exception

Bjoe:
Add workaround for clang, move inline functions from header to source file. See issue #1950

### DIFF
--- a/Net/include/Poco/Net/SocketAddressImpl.h
+++ b/Net/include/Poco/Net/SocketAddressImpl.h
@@ -77,6 +77,7 @@ private:
 // inlines
 //
 
+#ifndef __clang__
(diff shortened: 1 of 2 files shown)


aleks-f:
@bjoe, let's just un-inline unconditionally, both 1.8.0 and develop. see [here](https://github.com/pocoproject/poco/commit/e0a5ebc06effa2b2d7e57d75178ad964f21c2073#diff-3c48530d19cb754dbb21615cf7a57c43)

aleks-f:
and both IPAddress and SocketAddress

Bjoe:
Close pull request because this issue will be fixed with #1971

PR #1960, opened 2017-10-26
Allow compile in freebsd.

proller:
TODO: fix platform-specific code in Foundation/src/Thread_STD_POSIX.cpp


Foundation/src/Thread_STD_POSIX.cpp
```diff
@@ -158,7 +158,7 @@ void ThreadImpl::setStackSizeImpl(int size)
 
 void ThreadImpl::setAffinityImpl(int cpu)
 {
-#if defined (POCO_OS_FAMILY_UNIX) && POCO_OS != POCO_OS_MAC_OS_X
+#if defined (POCO_OS_FAMILY_UNIX) && POCO_OS != POCO_OS_MAC_OS_X && POCO_OS != POCO_OS_FREE_BSD
 #ifdef HAVE_PTHREAD_SETAFFINITY_NP
```
aleks-f: @proller where exactly is this defined?

Issue #1973, opened 2017-11-02
clang with libstdc++ in cmake build system


Bjoe:
Poco version: 1.8.0
Commit: 6f3aef61d13aeccbae6000500b973726f265a880
Distributor ID: Ubuntu
Description:    Ubuntu 17.04
Release:        17.04
Codename:       zesty
$ clang++-3.8 -v
clang version 3.8.1-18ubuntu1 (tags/RELEASE_381/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/i686-linux-gnu/6.3.0
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0
Found candidate GCC installation: /usr/lib/gcc/i686-linux-gnu/6.3.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6.3.0
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0
Candidate multilib: .;@m64
Selected multilib: .;@m64

In the cmake build system I cannot use anymore clang with libstdc++ because of this lines in CMakeLists.txt:74

```
if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
endif()
```

If I comment this line 74, all is fine.

@aleks-f I see in the commit 87ec631b4a374bfe4e6b4731ef90e90d94d13cd2 that you add this lines. Can you remember what was the purpose?

aleks-f:
yes,  I was hitting [this bug](https://sourceware.org/bugzilla/show_bug.cgi?id=13566) on travis because clang uses g++ stdlib unless libc++ explicitly specified
perhaps you can try to detect in cmake which stdlib version is on the build host and select accordingly. I think bug applies only to stdlib accompanying g++ 4.8, but I'm not 100% sure

aleks-f:
alternatively, maybe replace `gets()` - the problem was somewhere in CppUnit, I don't think we use gets in poco

Bjoe:
Hm [this bug in glibc](https://sourceware.org/bugzilla/show_bug.cgi?id=13566) is fixed for all glibc version since 2012-03-06 . Is this really the case?
What is with the "make build" system ? We should also add here libc++ ....  
[...]

PR #1975, opened 2017-11-04
Fix clang build error in poco 1.8.0

Bjoe:
Try to fix clang build error in poco 1.8.0 to get rid off to set libc++ for clang. See also issue #1973

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,11 +70,6 @@ else()
     endif()
 endif()
 
-if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
-endif()
-
-
 if(NOT MSVC_IDE)
   if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
--- a/cmake/CXX1x.cmake
+++ b/cmake/CXX1x.cmake
@@ -22,12 +22,21 @@
 macro(check_for_cxx11_compiler _VAR)
     message(STATUS "Checking for C++11 compiler")
     set(${_VAR})
-    if((MSVC AND (MSVC10 OR MSVC11 OR MSVC12 OR MSVC14)) OR
+    try_compile(_COMPILER_TEST_RESULT ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/test_compiler.cpp CMAKE_FLAGS -DCMAKE_CXX_STANDARD=11 -DCMAKE_CXX_STANDARD_REQUIRED=ON)
+    if(NOT _COMPILER_TEST_RESULT AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+      try_compile(_COMPILER_TEST_RESULT ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/test_compiler.cpp CMAKE_FLAGS -DCMAKE_CXX_FLAGS="-stdlib=libc++" -DCMAKE_CXX_STANDARD=11 -DCMAKE_CXX_STANDARD_REQUIRED=ON)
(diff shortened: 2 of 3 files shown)


aleks-f:
@obiltschnig see #1973 
@Bjoe perhaps a cmake command-line switch would work better?

Bjoe:
@aleks-f Hm I'm not sure what you mean exactly with command-line switch, but it is possible to set -DCMAKE_CXX_FLAGS="-stdlib=libc++", then poco is build with clang and C++14 via cmake. Should I add these in .travis.yml? At the moment, poco is build with clang and C++11 via cmake.

aleks-f:
With my change clang will always compile with `-stdlib=libc++`. There should be a way to prevent that.

Bjoe:
We can add an new cmake variable and verify if this variable is set via command parameter.
But I removed this line:
```
 set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
because this will be make more problems then solve.

For example I get following error on my ubuntu without installing libc++:
```
20:54:54: Starte "/usr/bin/cmake" --build . --target all
[1/595 11.1/sec] Building CXX object Foundation/CMakeFiles/Foundation.dir/src/Base32Encoder.cpp.o
FAILED: Foundation/CMakeFiles/Foundation.dir/src/Base32Encoder.cpp.o 
/usr/bin/clang++-5.0   -DFoundation_EXPORTS -DHAVE_PTHREAD_SETAFFINITY_NP -DPCRE_STATIC -DPOCO_ENABLE_CPP11 -DPOCO_ENABLE_CPP14 -DPOCO_HAVE_FD_EPOLL -DPOCO_NO_AUTOMATIC_LIBS -DPOCO_OS_FAMILY_UNIX -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -D_REENTRANT -D_THREAD_SAFE -D_XOPEN_SOURCE=500 -I/poco/Foundation/include -I/poco/Foundation/src -stdlib=libc++ -g -D_DEBUG -fPIC   -std=gnu++14 -MD -MT Foundation/CMakeFiles/Foundation.dir/src/Base32Encoder.cpp.o -MF Foundation/CMakeFiles/Foundation.dir/src/Base32Encoder.cpp.o.d -o Foundation/CMakeFiles/Foundation.dir/src/Base32Encoder.cpp.o -c /poco/Foundation/src/Base32Encoder.cpp
In file included from /poco/Foundation/src/Base32Encoder.cpp:15:
In file included from /poco/Foundation/include/Poco/Base32Encoder.h:21:
In file included from /poco/Foundation/include/Poco/Foundation.h:113:
/poco/Foundation/include/Poco/Alignment.h:35:10: fatal error: 'cstddef' file not found
#include <cstddef>
         ^~~~~~~~~
```
Nobody will understand that they have a missing libc++ library.
.libc++ is not the default standard C++ library on linux systems. Also not on android, but on android clang is the default compiler. We should use the "default" standard c++ lib" from the platform.

aleks-f:
> We should use the "default" standard c++ lib" from the platform.

That's fine, just make sure travis builds ok.

I don't particularly like the solution proposed in this pull - trying to compile a hello world does not tell much. What I have done (and what has caused this issue with clang/libstdc++) is alignment with the [make](https://github.com/pocoproject/poco/blob/develop/build/script/cpp11-clang) determination of compiler c++11/14 capabilities, based on completeness of standard support (and which I now see should also be backported from develop to 1.8).

/cc @obiltschnig

Bjoe:
> I don't particularly like the solution proposed in this pull - trying to compile a hello world does not tell much.
Ok. I try to compile against std::gets but this make no sense because it is removed in C++14 ...
Anyways I change my solution and set libc++ if it is possible and give a status message to install it. See my changes.
Hopefully this is more acceptable? Or any other ideas/comments?

But the best solution in my opinion is to use the cmake command **target_compile_features** See [compile-feature-requirements](https://cmake.org/cmake/help/v3.2/manual/cmake-compile-features.7.html#compile-feature-requirements)

obiltschnig:
@bachp are you ok with the CMake changes here?

Issue #1984, opened 2017-11-08
pd_json strerror deprecation warning on Windows


obiltschnig:
@aleks-f Could you look into this?

```
      "P:\git\poco-1.8.0\JSON\JSON_vs150.vcxproj" (rebuild target) (1) ->
      (ClCompile target) ->
        src\pd_json.c(47): warning C4996: 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use
      _CRT_SECURE_NO_WARNINGS. See online help for details. [P:\git\poco-1.8.0\JSON\JSON_vs150.vcxproj]
        src\pd_json.c(144): warning C4996: 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use
       _CRT_SECURE_NO_WARNINGS. See online help for details. [P:\git\poco-1.8.0\JSON\JSON_vs150.vcxproj]
        src\pd_json.c(162): warning C4996: 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use
       _CRT_SECURE_NO_WARNINGS. See online help for details. [P:\git\poco-1.8.0\JSON\JSON_vs150.vcxproj]
```

PR #1987, opened 2017-11-09
Fix/json strerror

aleks-f:
Fix `pd_json strerror deprecation warning on Windows` #1984:

- [x] replace `strerror` (not thread safe) in pd_json with `strerror_r`
- [x] define `strerror_r` for `_MSC_VER`
- [x] remove `_CRT_SECURE_NO_WARNINGS` from JSON progen
- [x] regenerate JSON VS projects


aleks-f:
@obiltschnig we should consider replacing `strerror` in [Error](https://github.com/pocoproject/poco/blob/poco-1.8.0/Foundation/src/Error.cpp#L71-L77) (that code has two ifdef cases that do the same thing) and [File_UNIX](https://github.com/pocoproject/poco/blob/poco-1.8.0/Foundation/src/File_UNIX.cpp#L447) as well

If strerror fix in this pull is ok, I will also apply it in develop and send it upstream to `pdjson`

PR #1989, opened 2017-11-09
CMake patches for FreeBSD

jspricke:


digitalist:
@jspricke what FreeBSD version did you test?
It doesn't build on FreeBSD 12 without 
`add_definitions( -D_XOPEN_SOURCE=700 -D__BSD_VISIBLE -DPOCO_HAVE_FD_POLL)`

jspricke:
Actually that was kFreeBSD in Debian, see https://buildd.debian.org/status/logs.php?pkg=poco&arch=kfreebsd-amd64 for the build logs.

digitalist:
thanks. 
I wonder if it's possible for cmake to distinguish between freebsd and kfreebsd, because that poll patch helps, but  link_libraries(dl) fails on freebsd (it's seems like a common linuxism),

It could be useful to separate options for freebsd and kfreebsd.

jspricke:
I've opened a pull request to address this, sorry for the trouble and thanks for the ping :).

digitalist:
I wouldn't find how to build this if it wasn't your commit! 
I'll try to install kfreebsd now and check how to match it for cmake

PR #1991, opened 2017-11-11
Support building poco 1.8.0 as cmake sub-project

guruofquality:
I support optionally building poco as a cmake subproject/git submodule. After updating to 1.8.0 series, I had to make the following minor change due to use of cmake variables. This is a pretty simple substitution and should be non-breaking for any use case.

The key here was replacing a few instances of CMAKE_SOURCE_DIR with PROJECT_SOURCE_DIR so cmake/test_compiler.cpp can be properly located in cmake/CXX1x.cmake. I also replaced CMAKE_BINARY_DIR with PROJECT_BINARY_DIR in this file to be consistent,  but that wasn't actually an issue.

* Reference: https://stackoverflow.com/questions/22214349/modify-cmake-source-dir/22220680#22220680


obiltschnig:
Merged manually.  Please don't submit pull requests against master branch.

PR #1993, opened 2017-11-12
implement Poco::SharedPtr using std::shared_ptr

obiltschnig:


mkrivos:
Hello, what is added value of this change?

obiltschnig:
Making it possible to use `Poco::SharedPtr` and `std::shared_ptr` together, and getting all the `std::shared_ptr` goodness like `std::make_shared` and `std::weak_ptr`.

And eventually getting rid of Poco::SharedPtr.

mkrivos:
Ok, this is 'value for money'. Otherwise the std::shared_ptr is slower and quite bigger than poco original version. std::weak_ptr is good thing for caches by example.

mkrivos:
By the way on gcc 7.2 this change (ptr.get() != _ptr.get() instead of ptr.get() != _ptr) fixes compilation:

	template <class Other>
	SharedPtr& assign(const SharedPtr<Other>& ptr)
	{
		if (ptr.get() != _ptr.get())
		{
			SharedPtr tmp(ptr);
			swap(tmp);
		}
		return *this;
	}

obiltschnig:
Could you please provide a pull request?

obiltschnig:
Do you have any concrete numbers regarding Poco::SharedPtr vs std::shared_ptr performance?

atangfan:
Dear Sir,

Implementing Poco::SharedPtr using std::shared_ptr should be good for a compiler which has std++11 support, but since POCO [1.8.x will be last major non-c++11 release](https://github.com/pocoproject/poco/pull/1795) , why do you remove **ReleasePolicy**/**ReleaseArrayPolicy** in this version? Besides, this change does NOT reveal in [summary of changes](https://pocoproject.org/docs/99100-ReleaseNotes.html) , especially in **Incompatible Changes and Possible Transition Issues**.

Is that possible keeping the original implementation of **SharedPtr**? 
After explicitly implementing SharedPtr by std::shared_ptr, this part can be removed without dispute. Thank you.

aleks-f:
[ReleasePolicy](https://github.com/pocoproject/poco/blob/poco-1.8.0/Foundation/include/Poco/SharedPtr.h#L59-L84) was **not** removed in 1.8

atangfan:
@aleks-f ,I apologize for the mistake! ReleasePolicy is still there even in 1.8.1 branch. Seems like I disordered my local POCO repo! Sorry again for the wrong information....
Thank you for the response.

PR #1994, opened 2017-11-14
guarantee the thread safety of redis client

richardor:


aleks-f:
@obiltschnig is there going to be 1.8.1 branch?

fbraem:
Is it the responsibility for the Redis Client class to make sure the writes are thread safe? We don't do that in MongoDB and the Data classes. If you use Redis in a multithreaded environment, I think it is the caller that must be sure that the client can't be used from other threads ...

obiltschnig:
@aleks-f There will be a 1.8.1 branch eventually. 

@fbraem The same is true for Data as well, so I see no need to make the Client thread safe. That particular patch only protects the `execute()` method, but clients may call other methods as well, thus breaking thread safety.

aleks-f:
@richardor thanks for the contribution but, as stated above, this functionality is not a responsibility of the redis library

richardor:
@aleks-f @obiltschnig @fbraem 
thank you guys.
only protect the execute() method is not considerate.
But s a library, which  encapsulated the redis operations,   if we make sure the thread safe of DB client.  then the caller will be more simple and clear.  I think that's the purpose of poco.
The mongodb and other Data classes are the same.
If you decide to not make the client thread safe. it's necessary to  point out it int the docs.

fbraem:
@richardor Thread safety in Poco is done using the ObjectPool class. This can be used for the Redis client and MongoDB for example. Separation of concerns: If we would do this in the client we are adding functionality which is not always necessary and must be repeated in all other client classes. While ObjectPool can add thread-safety for all Poco classes and your classes.

richardor:
@fbraem  
thank you.
I didnot  use the PooledConnection correctly. I get the redis client from the PooledConnection , and then release the PooledConnection. Thus,  the redis client is not thread safety.

PR #2002, opened 2017-11-15
Fix EVPTest on RHEL/Fedora by switching to different EC curve

swt2c:
RHEL/Fedora seem to have a much more limited set of EC curves available by
default.  secp384p1 seems to be available on Fedora, RHEL6, and RHEL7.

### DIFF
--- a/Crypto/testsuite/src/EVPTest.cpp
+++ b/Crypto/testsuite/src/EVPTest.cpp
@@ -330,7 +330,7 @@ void EVPTest::testECEVPSaveLoadStream()
 			/*TODO: figure out why EVP_PKEY_cmp() fails for identical public keys
 			assert (key == key2);
 			assert (!(key != key2));*/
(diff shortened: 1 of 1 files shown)


aleks-f:
Thanks for reporting, just so we avoid breaking some other platform, would you mind adding a [ECKey::hasCurve()](https://github.com/pocoproject/poco/blob/develop/Crypto/include/Poco/Crypto/ECKey.h#L97) check (and maybe try 2 or 3 names) - not bullet proof but obviously better than single name.

aleks-f:
Alternatively, add `randomCurveName()` for testing purposes

obiltschnig:
Should we also back port for 1.8.1?

swt2c:
Updated to use `curveName` instead which was already being used in other places in the test suite.

A back port would be great so we can reduce the number of patches we carry in Fedora.

Issue #1475, opened 2016-10-28
Allow client sockets to be bound to a local IP address


obiltschnig:
See https://pocoproject.org/forum/viewtopic.php?f=12&t=6774

Also related: #892, #411

roccocorsi:
I have the same requirement for this feature.  I found this issue through a post I made on stackoverflow -

[https://stackoverflow.com/questions/47229554/how-to-bind-specific-source-ip-address-using-poconethttpsclientsession](https://stackoverflow.com/questions/47229554/how-to-bind-specific-source-ip-address-using-poconethttpsclientsession)

Is there any progress with the implementation?  Is there anything I can do to help it along?

Thank you!

aleks-f:
> Is there any progress with the implementation?

not that I know of, @obiltschnig  ?

> Is there anything I can do to help it along?

yes, implement it and send a pull request to [develop](https://github.com/pocoproject/poco/tree/develop) branch. we'll make sure it ends up in all the right places (ie. to be released in 2.0 and be backported to 1.8.x series of releases).

roccocorsi:
Hi Alex.  Guess you are the same Alex as on stackoverflow.  :-)

In [https://pocoproject.org/forum/viewtopic.php?f=12&t=6774](https://pocoproject.org/forum/viewtopic.php?f=12&t=6774) there is discussion that a fix was core dumping.  Is this what is holding this up or is it just lack of anyone implementing?

Thanks

roccocorsi:
I meant Aleks, sorry about that.

aleks-f:
> I meant Aleks, sorry about that.

not a problem, I go by both and i've been called worse names :)

> Is this what is holding this up or is it just lack of anyone implementing?

I'm not aware of any work done yet, some promises but no delivery so far ...
The core dump may refer to an old poco version or some workaround the forum poster tried and was crashing.
I know of no reason this would not work and I think it is not a large amount of work. If you are serious about doing it, please do read [CONTRIBUTING.md](https://github.com/pocoproject/poco/blob/develop/CONTRIBUTING.md) and send a pull.

roccocorsi:
Hi Aleks

I got this to work for HTTP and HTTPS.

Should I post here to get comments or should I immediately do pull request?

Thanks,

aleks-f:
Please do pull request to develop branch - it's the easiest way to discuss a contribution. I will have limited time during the next couple of weeks, but @obiltschnig may be able to give some feedback.

[...]

PR #2016, opened 2017-11-27

roccocorsi:
My solution for adding the feature to allow binding the source IP address for HTTPClientSession and HTTPSClientSession reported in issue (#1475).

I've successfully checked the code with cppcheck and valgrind; did the development and testing in Linux OpenSuse Leap 42.3

I wrote a small test program (I can attach if you want to see it) to test HTTP and HTTPS protocol.  I have a vm with two virtual network interfaces, some routing setup code to allow the traffic to pass through the non-default route, and checked with tcpdump that traffic was going to the right interface and that the source IP address was the one I wanted.  All looks good.

I'm not sure how to run the unittest, so haven't run that.  I also did not write any tests, but if you think that is needed or possible, I could look into that once I figure out how to run the unittests.

Thanks,


roccocorsi:
I just made some additional updates to the same commit.  I did a "git push --force" to keep them in the same commit and not add another commit on top, hopefully that is acceptable?

Thanks,

PR #2028, opened 2017-12-01
Support templated array parameter syntax

WeiChungChang:
Although SharedPtr has become only a wrapper now, it sill stands there.

The removal of ReleasePolicy will bring up potential dangerous for people who are not familiar with the usage & history about SharedPtr.

For example, a simple program will make system crash.
```
{
    ....
    SharedPtr<myClass> p = new myClass[10];
    /// crash when p is out of the life scope.
}
```
As the result, here provides a small wrapper to use templated array parameter.
Ex, for the same logic above:
```
{
    ....
     /// declare by myClass[] to indicate it is an array.
    SharedPtr<myClass[]> p = new myClass[10];
    /// no crash
}
```
It provides SharedPtr<**T[]**> syntax to distinguish from original SharedPtr<**T**>.
It avoids the complex usage of deleter & provides a compromised way to support SharedPtr. 

Thanks

### DIFF
--- a/Foundation/include/Poco/SharedPtr.h
+++ b/Foundation/include/Poco/SharedPtr.h
@@ -337,6 +337,45 @@ protected:
 	std::shared_ptr<C> _ptr;
 };
 
+/// Provide a simple way of using SharedPtr to dynamically allocate array.
+/// ex, SharedPtr<class_A[]> p(new class_A[10]);
+/// For non-promitive type it prevents from crash.
+template <class C>
+class SharedPtr<C[]>
+{
+public:
+	typedef C  element_type;
+	typedef C* pointer;
+
+	SharedPtr(C* ptr):
+		_uni_ptr(ptr), _ptr(std::move(_uni_ptr)), m_pT(ptr)
+	{
+	}
+
+	SharedPtr& assign(C* ptr)
+	{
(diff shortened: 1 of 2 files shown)


Foundation/include/Poco/SharedPtr.h
```diff
@@ -337,6 +337,45 @@ class SharedPtr
...
+		return m_pT[i];
+	}
+
+	~SharedPtr()
+	{
+	}
+
+protected:
+	/// do NOT change the order!
+	std::unique_ptr<C[]> _uni_ptr;
```
aleks-f: [coding style](https://www.appinf.com/download/CppCodingStyleGuide.pdf)

Foundation/src/SharedMemory_POSIX.cpp
```diff
@@ -32,7 +32,7 @@ SharedMemoryImpl::SharedMemoryImpl(const std::string& name, std::size_t size, Sh
 	_access(mode),
 	_name("/"),
 	_fileMapped(false),
-	_server(server)
+	_server((mode == SharedMemory::AM_WRITE))
```
aleks-f: this was sent in #2032 , let's not mix it here please

aleks-f:
please add test case for this feature

PR #2043, opened 2017-12-09
Small code style change

cpsw:
Making operator precedence explicit.


aleks-f:
I'll merge, but this should be sent upstream to [libharu](https://github.com/libharu/libharu) or it will be overwritten on next update

PR #2054, opened 2017-12-17

roccocorsi:
Added example/test program and a readme to explain how to setting source IP address can be used.


roccocorsi:
Hi Günter 

I think I applied what you requested in your review.  I wasn't completely sure about what you meant with the source port part, but I left it open that a user of the class can set either IP address or port, or both at the same time.  Not sure if all those possibilities have actual useful uses, but you never know what someone might end up doing with this capability.

Made a new commit and will make a new pull request after I do some additional testing to make sure my changes didn't break anything.

Should I close this pull request?  Is that what would be done normally?

roccocorsi:
I will close this pull request and create a new one with updates.

Issue #2058, opened 2017-12-18
Synchronization issue in Poco::Util::Timer at destruction


SebastianTaube:
#### Expected behavior
When on a Timer object `cancel()` is called and the object is deleted immediately afterwards, the object's destructor should return.
#### Actual behavior
Dependent on scheduler behavior, constructor returns or is stuck at `_thread.join();`
#### Steps to reproduce the problem
When calling `cancel()` and deleting the Timer object immediately afterwards, first `CancelNotification` is enqueued by `cancel()` and afterwards `StopNotification` is enqueued by `~Timer()`. If inbetween there was no rescheduling to the Timer's thread, the following happens, once the thread is running again:

Due to the execution of `CancelNotification` the `_queue` is being cleared, so the `StopNotification` is being deleted without being executed. As a result, the thread's while loop is not left and `_thread.join();` in `~Timer()` is blocking forever.
#### POCO version
1.7.2
#### Compiler and version
Visual Studio 2015
#### Operating system and version
Windows 7 Build 7601

aleks-f:
I see two solutions:

1) not relying on return from `StopNotification::execute()`, but have stop flag
2) in `CancelNotification`, examine the queue to prevent inadvertent removal of `StopNotification`

But that's just from me reading the code, there may be better solutions.

obiltschnig:
The correct way would be to call `cancel(true)` in that case, which will block until the `CancelNotification` has been processed.

PR #2062, opened 2017-12-20
#2058

SebastianTaube:
- Introduced stop flag to ensure returning of run()
- removed previously used boolean return value of notifications

### DIFF
--- a/Util/include/Poco/Util/Timer.h
+++ b/Util/include/Poco/Util/Timer.h
@@ -164,6 +164,7 @@ private:
 	
 	Poco::TimedNotificationQueue _queue;
 	Poco::Thread _thread;
+	bool _stop;
 };
 
 
--- a/Util/src/Timer.cpp
+++ b/Util/src/Timer.cpp
@@ -37,7 +37,7 @@ public:
 	{
 	}
 	
-	virtual bool execute() = 0;
+	virtual void execute() = 0;
 	
 	Poco::TimedNotificationQueue& queue()
 	{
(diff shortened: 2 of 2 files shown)


Util/include/Poco/Util/Timer.h
```diff
@@ -164,6 +164,7 @@ class Util_API Timer: protected Poco::Runnable
 	
 	Poco::TimedNotificationQueue _queue;
 	Poco::Thread _thread;
+	bool _stop;
```
obiltschnig: Should probably be std::atomic<bool>

RobertAcksel: it would be nice but std::atomic is c++11 and can not be used therefore

aleks-f: sure it can - 1.x is open for c++11 conditionally, develop ([2.0](https://github.com/pocoproject/poco/milestone/13)) unconditionally

RobertAcksel: what are the conditions. just to say we can not in our project since our customer required us to stick with c++03 🤦‍♂️

aleks-f: then you should stick with [1.x](https://github.com/pocoproject/poco/blob/poco-1.8.1/Foundation/include/Poco/Config.h#L27) releases

RobertAcksel: sure that's what we do. and the PR is for the next 1.x release. isn't the develop still aiming for the next 1.x version

aleks-f: no, develop will become 2.0. If you want it in [1.8.1](https://github.com/pocoproject/poco/tree/poco-1.8.1), then another pull has to be sent there, but see with @obiltschnig about the exact 1.8.1 release date

Issue #2075, opened 2018-01-02
XML/DOM: Remove raw pointers from public interfaces


obiltschnig:
Raw pointers in XML/DOM interfaces should be replaced with smart pointers (`Poco::AutoPtr`).

Unfortunately, this cannot be done without breaking existing code. After extensive analysis, a method has been found to implement this with minimal backwards compatibility issues. The main concern was that code requiring changes should not break more or less silently or crash at runtime (e.g., creating memory leaks or double deletes), but break at compile time.

For passing DOM objects as method arguments and return values, the following mechanism will be used:

  1. The `duplicate()` and `release()` methods in `DOMObject` will be made `protected`, therefore     preventing explicit calls to them from client code. This will require changes to client code not already using `Poco::AutoPtr` (as it should already).

  2. A new class template `Poco::XML::NodePtr` will be created for passing DOM Node objects as method arguments. This will be a subclass of `Poco::AutoPtr`, but with the semantics of the constructor taking a raw pointer changed to call duplicate(). This way, existing code using raw pointers will continue to work with correct reference counting. However, this constructor will be marked as deprecated, so that a warning will be issued when it's being used.

  3. All methods returning raw `DOMObject` pointers will be changed to return `Poco::AutoPtr`. This should not break any code, except code that explicitly calls `release()`, e.g., after receiving a `NodeList` or `NamedNodeMap`. However, since `release()` is no longer public, breakage will occur during compilation, and not during runtime in the form of memory leaks or double deletes.

  4. Eventually, in a later version, `Poco::XML::NodePtr` will be removed and replaced with `Poco::AutoPtr`.

dawidpilarski:
Any reason, why this  issue was marked as wontfix? 

I am willing to contribute to that particular issue as it already caused lots of troubles in one of the projects I am working on.

obiltschnig:
No way to do this in a way that keeps it backwards compatible to existing code. So basically we'll have to implement a new API in parallel to the existing one. There's simply no resources to do it.

dawidpilarski:
Thanks for quick response @obiltschnig 

Are changes breaking the code at compile time (as previously considered) also not an option?

> The main concern was that code requiring changes should not break more or less silently or crash at runtime (e.g., creating memory leaks or double deletes), but break at compile time.

I think such an approach would be possible.

obiltschnig:
[...]

PR #2078, opened 2018-01-03

roccocorsi:
New pull request with changes suggested by Günter.

I was thinking about this over the holidays and it is obvious that this is an IPv4 only solution.  I'm not an IPv6 expert at all so not really sure what would be the proper way to make it IPv4 and IPv6 compliant, and how smart to make it.

### DIFF
--- a/Net/samples/SetSourceIp/SetSourceIp.readme
+++ b/Net/samples/SetSourceIp/SetSourceIp.readme
@@ -1,0 +1,274 @@
+
+SetSourceIP is an example program and test tool that will set the source IP address in the socket of clients using the HTTPClientSession or HTTPSClientSession classes.
+
+
+Setting the source IP address in a client is not a common thing to do, but could be useful in the following sitations:
+
+1) In a multi-homed system where multiple network interfaces are used and the client http/https traffic must go out through the interface that is not considered the "default" routed interface.  This is sometimes referred to as traffic separation or traffic isolation.  Setting the source IP address is one part of the puzzle in order to allow this to work.  Additonal OS routing setup is required to successfully make this work.
+
+2) In a clustered environment, where two hosts provide a client service in a Active/Standby situation and it is desirable to present a single source IP address to the remote server and/or intermediate firewalls.  This might be referred to as Floating, Mobile, or Virtual IP addressing.
+
+3) In a host where a network interface supports multiple IP addresses and the client wants to assign the source IP address for some application specific reason.
+
+
+For more in depth information, please do additional research with keywords such as:
+Linux Multihomed
+Routing with multiple network interfaces
+Linux Advanced Routing & Traffic Control
+
+
+
+What follows is an example of the routing setup required in a Linux VM in order to experiment and get comfortable with this type of source IP address routing.
+
+
+WARNING: Thoroughly verify and validate the procedure and all the included commands in a test environment before using in any sort of production system.  No promises are made to the fitness of any information that is provided in this document.
+
+
+1- Create a VM in your vm environment of choice. (e.g. Virtual Box, VmWare Server)
+
+2- Make sure that the VM has two network interfaces defined and that they are attached to through NAT. NAT is required so that they both can reach the Internet when required.
+
+3- Install your OS of choice on the VM. (e.g. OpenSuse Leap 42.3)
+
+4- After the VM is started up, check that the two network interfaces are well defined (i.e. they are 'UP', they have a reasonable IP address, netmask, broadcast address.)  Usually they will be named 'eth0' and 'eth1', but it might be something else in your OS.  Make sure that the Internet can be reachable at least on one of the interfaces.
+
+Command:
+ifconfig -a
+
+
+
+5- Find which one of the two interfaces is the 'default' destination router. Usually it will be the 'eth0' interface, but it might be the 'eth1' one instead.
+
+Command:
+netstat -r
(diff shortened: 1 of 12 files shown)


Net/samples/SetSourceIp/src/SetSourceIp.cpp
```diff
@@ -0,0 +1,150 @@
+
```
aleks-f: since secure sockets are supported here, it properly belongs in NetSSL_OpenSSL/samples

aleks-f:
I'm ok with IPv4-only for now. If anyone has IPv6 itch, they can scratch it :)

Net/samples/SetSourceIp/SetSourceIp.readme
```diff
@@ -0,0 +1,274 @@
+
```
aleks-f: since secure sockets are supported here, it properly belongs in NetSSL_OpenSSL/samples

roccocorsi: Sure I can move it there.

roccocorsi: Aleks, did you  retrigger running the checks?  I remember them being successful last time I ran them.

roccocorsi: OK, I see that you made some small changes (adding copyright notice and fixing tabs), must have retriggered the checked with that.

roccocorsi:
OK, I moved the SetSourceIP example to sample directory in NetSSL_OpenSSL.  First time that I update a branch like this and pleasantly surprised that the checks are run automatically.  I thought I had to make a new force commit at same level.

Lets hope checks pass.  :-)

roccocorsi:
One of the check jobs failed due to some scripting issue.

Should I squash all changes into one commit and try again?

Net/samples/SetSourceIp/SetSourceIp.readme
aleks-f: yes, can you also add CMakeLists.txt and Makefile for the sample - it should be simple, just take the ones from an existing example and modify the names accordingly.

roccocorsi:
Hi Aleks

I received an email with comments from you about adding files CMakeLists.txt and Makefile for SetSourceIp example program.   I can't seem to find your comments on the github site, so I can't answer them directly.

Anyway, I tried to copy CMakeLists.txt from HTTPSTimeServer and make changes to it to refer to SetSourceIp where expected.  Ran cmake on the file, seems to work, and it created a 'Makefile' file.

Tried to 'make' the Makefile, but I get a lot of errors when compiling related to enabling C++ 11.

Add the the following line in CMakeLists.txt
add_definitions(-std=c++11)

Everything works with that addition.   Not sure if forcing c++11 is OK?  When I compile other samples I get same C++ 11 issues.  Do I have too old or too new cmake or g++?!?!

Any suggestions to move forward?

roccocorsi:
I've made a new commit with CMakeLists.txt and Makefile changes.  Not 100% sure how to properly test them.  Would appreciate feedback.

roccocorsi:
I think travis-ci might be stuck!!  Is there a way to abort and restart the check jobs?

zosrothko:
@roccocorsi You can either restart the whole build or restart a single job. Click on the small icon on the right of each job or on the top left icon.

aleks-f:
I restarted the build. As for CMake, I will look at the issue, hopefully by the end of week. Generally speaking, yes - C++11 is ok in develop branch

roccocorsi:
Sorry, been away for some time.  Is there anything else that is needed from me on this commit to get it merge on develop branch?
Thank you.

aleks-f:
https://github.com/pocoproject/poco/commit/d0312c6f6007aba47908fae250a098f6faf65053

had to add VS projects, then had upstream conflict. reset, then commit again and @roccocorsi name in commits got lost. sorry, we like people to get proper credit for their contributions, but it's too much to go through the whole cycle again

aleks-f:
tried to amend to at least point to your contributions, ended up wth two commits and a merge, lol. anyway, thank you for the contribution :)

roccocorsi:
Glad to hear that it is all merged.  No problem if my name is not part of the commit.

I might be back in a few weeks with IPv6 part, maybe...

roccocorsi:
Any ideas when 2.0.0 might go stable?

aleks-f:
hard to tell, months for sure. here's the [project](https://github.com/pocoproject/poco/projects/1), and there is one fairly large outstanding [issue](https://github.com/pocoproject/poco/issues/2075) that needs to be fixed before release (first release will be beta).

If you can help, it will be faster :)

p-neels:
Have tried the changes from this pull request but still it doesn't work.
setsockfd was the solution.
But setsockfd doesn't work straight forward.Hence have created below issue for same.
https://github.com/pocoproject/poco/issues/2400

Is there any plans to expose setsock option for secure connection directly to user for a secure connection?

gelldur:
There is issue when we close connection and try to reconnect. Such session wont use correct source address. 
I found that bug is caused by:
```
void SecureSocketImpl::connect(const SocketAddress& address, const Poco::Timespan& timeout, bool performHandshake)
```

because in first line we do: 
```
if (_pSSL) reset();
```

Which will recreate new socket without binding. I'm working on fix.

PR #2081, opened 2018-01-08
Fix bugs in `Poco::Net::Socket::select`, fixes #2080

### LINKED ISSUE #2080: Bugs in Poco::Net::Socket::select when POCO_HAVE_FD_POLL is defined
When `POCO_HAVE_FD_POLL` is defined, the `Poco::Net::Socket::select` implementation contains two bugs:

* It never initializes the memory it uses for the `pollfd` array to pass to `::poll`, and so the `events` bitfield can contain events that the caller isn't interested in, such as the `POLLOUT` events on a socket which is only in the `readList`. This can cause `::poll` to return prematurely (or even immediately), when a socket is e.g. writeable, but still doesn't match the conditions requested, e.g. readable. This means a loop calling Poco::Net::Socket::select effectively goes into a tight loop.
* The `remainingTime` timespan is not used when calling `::poll`

About to submit a PR to fix both.

Burgch:
See issue #2080 for details.

### DIFF
--- a/Net/src/Socket.cpp
+++ b/Net/src/Socket.cpp
@@ -211,7 +211,8 @@ int Socket::select(SocketList& readList, SocketList& writeList, SocketList& exce
 	nfds_t nfd = readList.size() + writeList.size() + exceptList.size();
 	if (0 == nfd) return 0;
 
-	SharedPollArray pPollArr = new pollfd[nfd];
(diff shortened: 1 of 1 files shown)


Burgch:
Ah, sorry, I knew that was the case, and yet somehow still managed to used the wrong branch without thinking!

PR #2088, opened 2018-01-12
Fix race condition in TCPServerDispatcher.cpp

### LINKED ISSUE #2085: Crash due to race condition in TCPServerDispatcher
#### Expected behavior
Clean server shutdown with no crash.

#### Actual behavior
#0  0x00007fc7d030a0c1 in pthread_kill () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007fc7c0da4ed1 in ?? () from <signal handler library>.so
#2  <signal handler called>
#3  0x0000000009078380 in ?? ()
#4  0x00007fc7c2ff429f in Poco::PooledThread::run() () from bin/glnxa64/libPocoFoundation.so.48
#5  0x00007fc7c2feff86 in Poco::ThreadImpl::runnableEntry(void*) () from bin/glnxa64/libPocoFoundation.so.48
#6  0x00007fc7d0305064 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#7  0x00007fc7d1ba462d in clone () from /lib/x86_64-linux-gnu/libc.so.6

This is a very rare issue that has been causing sporadic crashes in our build and test environment. I believe the issue is reproducible when there is a server.stop() call concurrent with a new socket connection.

I believe the problem is that a new socket connection enqueues the TCPServerDispatcher instance without changing the reference count. That means that if you shutdown the server before the TCPServerDispatcher::run method starts the ref count would go to zero before the AutoPtr in run() has a chance to increment the ref count.

I think that incrementing the ref count before enqueuing TCPServerDispatcher (and decrementing at the end of run()) would resolve the issue and ensure that the run() method can be safely called during server shutdown.

#### POCO version
1.7.9

#### Compiler and version
gcc 6.3.0-GLIBC2.11

debian 8

Burgch:
Should fix #2085

### DIFF
--- a/Net/src/TCPServerDispatcher.cpp
+++ b/Net/src/TCPServerDispatcher.cpp
@@ -92,7 +92,7 @@ void TCPServerDispatcher::release()
 
 void TCPServerDispatcher::run()
 {
-	AutoPtr<TCPServerDispatcher> guard(this, true); // ensure object stays alive
+	AutoPtr<TCPServerDispatcher> guard(this, false); // ensure _rc is decreased when function exits
 
 	int idleTime = (int) _pParams->getThreadIdleTime().totalMilliseconds();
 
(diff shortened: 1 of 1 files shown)


obiltschnig:
Had to revert this fix as it breaks the tests.

Burgch:
The implementation of `duplicate()` is different between the develop and 1.9.0/1.9.1 branches. Can you re-merge, replacing the call to `duplicate()` with `++_rc` (safe because the `enqueue()` function already owns the mutex).

PR #2089, opened 2018-01-14
HTTP Authentication Schemes

Bjoe:
Allow hyphen in authenticate param token to implement other HTTP authentication schemes. See [HTTP Authentication Schemes](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
)


obiltschnig:
manually merged into 1.9/develop

PR #2138, opened 2018-02-07
Only escape compulsory characters for JSON by default (#2137)

### LINKED ISSUE #2137: JSON Stringify Unicode Escape functionality breaking change with 1.8.0
#### Expected behavior

In versions prior to 1.8.0, the JSON encoding wasn't fully [spec](https://tools.ietf.org/html/rfc7159#section-7) compliant, as tracked by #1042. However, also as per the JSON spec, it didn't escape every unicode character (the spec says every character **may** be escaped, but as JSON is required to be encoded in either UTF-8, UTF-16, or UTF-32 - see [section 8.1](https://tools.ietf.org/html/rfc7159#section-8.1) - there's no need to escape every unicode character). The expected fix is that the compulsory characters are escaped, but not every other unicode character (doing so can dramatically increase the size of a JSON string unnecessarily). 

#### Actual behavior

The 1.8.0 release has broken code which uses it because it now escapes every single unicode character.

The existing tests demonstrate the "problem".

#### POCO version

1.8.0+

#### Compiler and version

Any

Any

#### Other relevant information

PR on its way.

Burgch:
Fixes #2137

### DIFF
--- a/Foundation/src/JSONString.cpp
+++ b/Foundation/src/JSONString.cpp
@@ -17,19 +17,51 @@
 
 namespace Poco {
 
-void toJSON(const std::string& value, std::ostream& out, bool wrap)
+void toJSON(const std::string& value, std::ostream& out, bool wrap, bool escapeAllUnicode)
 {
 	if (wrap) out << '"';
-	out << UTF8::escape(value.begin(), value.end());
+	if (escapeAllUnicode)
+	{
+		out << UTF8::escape(value.begin(), value.end());
+	}
+	else
+	{
+		for (std::string::const_iterator it = value.begin(),
+			end = value.end(); it != end; ++it)
+		{
+			// Forward slash isn't strictly required by JSON spec, but some parsers expect it
+			if ((*it >= 0 && *it <= 31) ||  (*it == '"') || (*it == '\\') || (*it == '/'))
+			{
+				out << UTF8::escape(it, it+1);
+			}
+			else out << *it;
+		}
+	}
 	if (wrap) out << '"';
 }
 
 
-std::string toJSON(const std::string& value, bool wrap)
+std::string toJSON(const std::string& value, bool wrap, bool escapeAllUnicode)
 {
 	std::string ret;
 	if (wrap) ret.append(1, '"');
-	ret.append(UTF8::escape(value.begin(), value.end()));
+	if (escapeAllUnicode)
+	{
+		ret.append(UTF8::escape(value.begin(), value.end()));
+	}
+	else
+	{
+		for (std::string::const_iterator it = value.begin(),
+			end = value.end(); it != end; ++it)
+		{
+			// Forward slash isn't strictly required by JSON spec, but some parsers expect it
+			if ((*it >= 0 && *it <= 31) ||  (*it == '"') || (*it == '\\') || (*it == '/'))
+			{
+				ret.append(UTF8::escape(it, it+1));
+			}
+			else ret.append(1, *it);
+		}
+	}
 	if (wrap) ret.append(1, '"');
 	return ret;
 }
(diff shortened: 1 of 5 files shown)


Burgch:
Fixed a bad merge in `JSONTest.cpp` and squashed the commits. The first CI build could be killed.

Foundation/src/JSONString.cpp
```diff
@@ -17,19 +17,45 @@
 
 namespace Poco {
 
-void toJSON(const std::string& value, std::ostream& out, bool wrap)
+void toJSON(const std::string& value, std::ostream& out, bool wrap, bool escapeAllUnicode)
 {
 	if (wrap) out << '"';
-	out << UTF8::escape(value.begin(), value.end());
+	if (escapeAllUnicode) {
```
aleks-f: please fix style

JSON/testsuite/src/JSONTest.cpp
```diff
@@ -1933,7 +1933,7 @@ void JSONTest::testEscapeUnicode()
 	result = parser.parse(utf8Text);
 	object = result.extract<Object::Ptr>();
 	ss.str(""); object->stringify(ss);
-	assert (ss.str() == "{\"name\":\"g\\u00FCnter\"}");
+	assert (ss.str() == "{\"name\":\"g\xC3\xBCnter\"}");
```
aleks-f: I'd leave this test as is and create another one that tests functionality without escaping. This change will break the code, so there should be a way to force the parser to escape.

Burgch: I disagree - the code was broken by 1.8.0, and this is returning it back to the former functionality. This test would fail as is with versions < 1.8.0

aleks-f:
@Burgch Somehow I ended up commenting in your repo, so both those comments and the ones in pull are showing here. In any case - overall, I agree with the pull and reasoning that escaping should not be default, but since we are breaking existing functionality here, somehow JSON library should have the option to escape, too (didn't yet look into details of what would be the best way to achieve this).

Foundation/src/JSONString.cpp
Burgch: Shall do.

Burgch: Done.

JSON/testsuite/src/JSONTest.cpp
aleks-f: Perhaps I did not state it correctly. This test is named `testEscapeUnicode` and it should do what its name implies - test (un)escaping the unicode. What was before does not matter, we should do the right thing now. And that includes allowing end user to choose whether the returned JSON string is unicode-escaped or not - probably just change in the JSON::stringifier.

Burgch: I'm not able to commit much more time on this; is there a small tweak to the PR that would mean you'd accept it? Perhaps renaming the test to `testParsingUnicode`, which is most of what it does anyway?

aleks-f: That would be fine, but the main concern is that a json containing unicode characters can not be escaped when stringified (that does not look like a big change, either).

Burgch: They don't need to be escaped though - any compliant JSON parser must accept the non-escaped characters as well as the escaped ones, as this test is demonstrating.

aleks-f: user puts in
```
"{ \"name\" : \"\\u4e2d\" }"
```
does something with it (add or remove something), and gets out
```
"{\"name\":\"\xE4\xB8\xAD\"}"
```
now, it is true that those things are equivalent for JSON and a compliant parser must accept them both. but from my experience, people can be more picky than parsers, especially about the superficial things. I'll finish it, thanks for reporting the issue and the fixes so far.

Burgch: Thanks very much, appreciate it!

aleks-f:
moved to #2145

Issue #2142, opened 2018-02-07
JSON::Object keys hold pointers to values map


aleks-f:
[keys](https://github.com/pocoproject/poco/blob/poco-1.8.1/JSON/src/Object.cpp#L183) (needed to preserve JSON keys order) hold pointers into the values map.

obiltschnig:
Hold an iterator instead?

aleks-f:
probably, yes

aleks-f:
the problem was only in the [assignment operator](https://github.com/pocoproject/poco/blob/develop/JSON/src/Object.cpp#L72), the rest was synced. but i changed it to iterators, cleaner.

aleks-f:
[fixed](https://github.com/pocoproject/poco/commit/df5968ce1633f02a42505f5a14a9db60f628facc) in develop

obiltschnig:
While we're at it, how about also doing this in `toJSON()` and `stringify()`? And ideally everywhere we're using bool parameters... (eventually)

aleks-f:
I'm assuming you are referring to options instead of bools as was done in JSON::Object ...

doable, but it's a fragile dependency; ideally, there would be a single options enum in Poco, used by Poco::toJSON, JSON::Object and JSON::Stringifier (and anything else in JSON that needs it); unfortunately, it is not possible in a backward-compatible way (see 3. below)

1. Object had preserveOrder (bool), now needs options 1 (preserve order),2 (escape unicode) - all good here, existing code works as expected

2. Stringifier had nothing, now needs escape unicode option, but it would have to default to 2 for compatibility with object, not perfect, but still ok

3. toJSON had wrap (bool), now needs wrap and escape unicode options, wrap would have to be 1 for backwards compatibility, so we could get by in a JSON::Object compatible way, but only with two separate sets of options with matching value of 2 for escape unicode

obiltschnig:
Keep the existing functions taking a single bool and mark as deprecated.
Add a new overload one taking int, with single set of options.

PR #2145, opened 2018-02-07
Json unicode escape && preserveOrder keys sync

aleks-f:
- [x] JSON Stringify Unicode Escape functionality breaking change with 1.8.0 #2137
- [x] JSON::Object preserveOrder keys not synced on assignment #2142

### DIFF
--- a/JSON/include/Poco/JSON/Object.h
+++ b/JSON/include/Poco/JSON/Object.h
@@ -64,6 +64,7 @@ public:
 	typedef ValueMap::value_type                ValueType;
 	typedef ValueMap::iterator                  Iterator;
 	typedef ValueMap::const_iterator            ConstIterator;
+	typedef std::vector<std::string>            NameList;
 
 	explicit Object(bool preserveInsertionOrder = false, bool escapeUnicode = false);
 		/// Creates an empty Object.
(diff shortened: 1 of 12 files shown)


JSON/include/Poco/JSON/Object.h
```diff
@@ -64,12 +64,16 @@ class JSON_API Object
 	typedef ValueMap::value_type                ValueType;
 	typedef ValueMap::iterator                  Iterator;
 	typedef ValueMap::const_iterator            ConstIterator;
+	typedef std::vector<std::string>            NameList;
 
-	explicit Object(bool preserveInsertionOrder = false);
+	explicit Object(bool preserveInsertionOrder = false, bool escapeUnicode = false);
```
obiltschnig: Maybe we should think about introducing an options enum for the various places where we use preserveInsertionOrder and escapeUnicode, and use an `int options` instead of a potentially growing list of bool args. If we just change the Object constructor to:

```
Object(int options = 0);
```

and define:
```
enum Options
{
    JSON_OPT_PRESERVE_INSERTION_ORDER = 1,
    JSON_OPT_ESCAPE_UNICODE = 2
};
```
Then the new constructor taking `int` will be backwards compatible to the one taking `bool`.

Burgch:
This looks great, thanks again!

PR #2155, opened 2018-02-14
Allow building with LibreSSL

TcT2k:
Building with LibreSSL defines a higher OpenSSL version but does not
provide some of the newer symbols. The necessary alternative code
is already available in the Poco code base and just needs additional
precompiler checks.


aleks-f:
OK, but just for the record - we do not (yet) officially support LibreSSL. I like it, but the API is still lagging behind OpenSSL and I went through enough pains with OpenSSL on Windows not to want to get into more with LibreSSL at this time. But thanks for the contribution.

PR #2170, opened 2018-02-18
Include Crypto.h in NetSSL.h to avoid missing link lib link error

zosrothko:


aleks-f:
I guess I did not test NetSSL build. In any case, it's now a bit more complicated because we are naming those libraries differently, and there is no "md" version of them, see [here](https://github.com/pocoproject/poco/blob/develop/Crypto/include/Poco/Crypto/Crypto.h#L87-L117) how to do it right

aleks-f:
but since NetSSL depends on crypto, as long as Crypto.h is included from somewhere, those pragmas should not be needed

zosrothko:
Those fixes on #pragma are needed otherwise in case of a link in debug mode, both `libssl.lib` /`libssld.lib`  and `libcrypto.lib`/ `libcryptod.lib` are required, which is wrong. Only the libs suffixed by 'd' should be required

aleks-f:
look at my first response again; what you have here wil only work for DLLs. it will not work for static, because libraries are [not named](https://github.com/pocoproject/openssl/tree/develop/build/win64/lib/debug) libcrypto and libssl anymore; additionally, there is no eg. libryptomdd.lib. Just include Crypto.h from NetSSL.h and problem is solved.

zosrothko:
Added `#include "Crypto.h"` in `NetSSL.h`. But there are redondant definitions for the POCO_LIB_SUFFIX between Foundation.h and Crypto.h and IMO it could be simplified in Crypto.h.

aleks-f:
> there are redondant definitions for the POCO_LIB_SUFFIX between Foundation.h and Crypto.h

Not sure what you mean - there's no POCO_LIB_SUFFIX definition in Crypto.h

> it could be simplified in Crypto.h

POCO_LIB_SUFFIX is opaque - it does not do what we need for linking OpenSSL libraries, which have no `md[d] incarnation, because static libs are only meant to be linked with static runtimes, hence only `mt[d]`

PR #2174, opened 2018-02-21
Improve openssl verify callback handler

Bjoe:
At the moment in https://github.com/pocoproject/poco/blob/e5f71bb08526f2feeb514a4c9f569302be98e6ac/NetSSL_OpenSSL/src/SSLManager.cpp#L202 the certification handler is called when pre verification in openssl fails.
I need the possibility to extend the x509 certification verification with my own "checks"/code when pre verification was successful in openssl. I create first a proposal how we can implement this. I think it needs improvements.
My first idea was to add another function like "onValidCertificate()" in https://github.com/pocoproject/poco/blob/e5f71bb08526f2feeb514a4c9f569302be98e6ac/NetSSL_OpenSSL/include/Poco/Net/InvalidCertificateHandler.h#L29 but the interface is named **Invalid**CertificateHandler and this interface is the "base interface" in https://github.com/pocoproject/poco/blob/e5f71bb08526f2feeb514a4c9f569302be98e6ac/NetSSL_OpenSSL/include/Poco/Net/CertificateHandlerFactory.h#L45
The next idea was to add a CertificateHandler interface and **Invalid**CertificateHandler is inherited from CertificateHandler but then I think I break the API.
So I decided to add a new CertificateHandler "structure" (it is more or less a copy of the InvalidCertificateHandler structure) and create a proposal to discuss with the poco team :-)
Btw. CertificateHandler is implemented as NVI pattern.

### DIFF
--- a/NetSSL_OpenSSL/Makefile
+++ b/NetSSL_OpenSSL/Makefile
@@ -11,12 +11,13 @@ SYSLIBS += -lssl -lcrypto
 objects = AcceptCertificateHandler RejectCertificateHandler ConsoleCertificateHandler \
 	CertificateHandlerFactory CertificateHandlerFactoryMgr \
 	Context HTTPSClientSession HTTPSStreamFactory HTTPSSessionInstantiator \
-	InvalidCertificateHandler KeyConsoleHandler \
+	CertificateHandler KeyConsoleHandler \
 	KeyFileHandler PrivateKeyFactory PrivateKeyFactoryMgr \
 	PrivateKeyPassphraseHandler SecureServerSocket SecureServerSocketImpl \
 	SecureSocketImpl SecureStreamSocket SecureStreamSocketImpl \
 	SSLException SSLManager Utility VerificationErrorArgs \
-	X509Certificate Session SecureSMTPClientSession
+	X509Certificate Session SecureSMTPClientSession \
+	VerificationArgs
 
 target         = PocoNetSSL
 target_version = $(LIBVERSION)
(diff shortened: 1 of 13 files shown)


aleks-f:
Your solution makes sense and is more general, but now there's this unintuitive mix of related functionalities - not only are they in different places, but they are also differently designed. 

To satisfy naming consistency, I think `InvalidCertificateHandler` could be renamed to  `CertificateHandler`, with added no-op default `onVerifyCertificate()` (or `extendedVerify()`) member. `InvalidCertificateHandler` could then be a typedef - as it appears in my head right now, that would not break code.

Bjoe:
@aleks-f Thank you for the hint! I have created a new proposal, when you have time to have a quick look and give your ideas/statements would be great.

aleks-f:
conceptually ok, but there's a bit of "can of worms" we're opening here ...

for example, following [this](https://github.com/pocoproject/poco/pull/2174/files#diff-678417465c43c51479d8ae05a90bee67R52)

will cause this to fail:
https://github.com/pocoproject/poco/blob/dcdcee5afab6e8972f1b144ef7e7e884f85e83a3/NetSSL_OpenSSL/testsuite/TestSuite.xml#L16

(but changing xml would mean we'll break people's code, so best to leave as-is, or handle both "invalidCertificateHandler" and "certificateHandler")

then, since the file name has changed (InvalidCertificateHandler -> CertificateHandler), pre-generated VS projects should also be updated accordingly (for things like this, I usually open all `.vcproj`, .`vcxproj`, .`vcxproj.filters` files and search/replace, but if new files are added, then progen must be re-run to regenerate projects, and this can only be done on windows)

(I'm actually puzzled why appveyor build and travis tests do not fail with this, perhaps I'm missing something)

finally, there's NetSSL_Win, with parallel functionality (where maybe we can leave things the old way, let's see what @obiltschnig  says)

obiltschnig:

Issue #1839, opened 2017-08-08
Unused parameter warnings


hakan-akan:
Can we do something about unused parameters? There are too many of them. (More than 100 warnings generated on my build.) A few examples of source files with many warnings are, DefaultHandler.cpp, XMLWriter.cpp, TextEncoding, etc.

I personally like the 
void f(int *ptr) {
  (void) ptr;
}

Or something like Q_UNUSED(x) macro can be defined where macro defined as

#if defined(Q_CC_RVCT)
template <typename T>
inline void qUnused(T &x) { (void)x; }
#define Q_UNUSED(x) qUnused(x);
#else
#define Q_UNUSED(x) (void)x;
#endif

Maybe even something line this 

template<class ... Ts>
inline void unused(Ts ... ) { }
 
void f(int *p, bool b, const int &k)
{
    unused(p, b, k);
}

aleks-f:
There's [POCO_UNUSED](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Platform.h#L257). Scratch your itch and send a pull :)

hakan-akan:
Fair enough. :)

lakshayg:
Has this issue been resolved? I could find only 2 unused functions when compiling with -Wunused.

aleks-f:
no, see my comment to your pull

hakan-akan:
To my taste POCO_UNUSED is too ugly. I'd rather have something less invasive. Especially the fact that it pollutes the function signature is irksome to me.  

For example,
void f(int POCO_UNUSED blah) // Instead of this 
{
}

void f(int blah)
{
POCO_UNUSED blah; // I would rather have but this but it does not silence warnings. (I use Qt mingw )
(void) blah; //This does silence warnings
}

PR #2178, opened 2018-02-22
add POCO_UNUSED attribute to remove build warnings

lakshayg:
Two functions in XML/src/siphash.h were reported as unused when POCO was built with the -Wunused flag. Adding POCO_UNUSED attribute to remove these warnings.

Relates to #1839


aleks-f:
we can merge this, but it is very likely to get overwritten unless it is somehow addressed upstream

PR #2180, opened 2018-02-23
BMP support and add image styles

micheleselea:
Add support for embed BMP images in PDF.
Add a couple of styles to let image better fit in page.
offsetX and offsetY to set a relative position of the image (can be positive or negative)
aspectratioH if true take the height value and recalculate width accordling to aspect ratio


micheleselea:
I did not follow some HTML standard in new styles, and I did it just for images, probably we could use a more standard tags...I just insert what I need because I have very few time at this moment, but can be useful as concept

PDF/include/Poco/PDF/hpdf.h
```diff
@@ -756,6 +756,10 @@ HPDF_EXPORT(HPDF_Image)
 HPDF_LoadJpegImageFromFile (HPDF_Doc      pdf,
                             const char    *filename);
 
+HPDF_EXPORT(HPDF_Image)
+HPDF_LoadBMPImageFromFile(HPDF_Doc      pdf,
+	const char    *filename);
+
```
aleks-f: this should go somewhere else, because we want to keep Haru files as a simple drop-in (otherwise, next time we update, this will get overwritten). 

so, options are:
1) move this function to Poco::PDF code (probably a new Util.h/cpp)
2) try to get it accepted upstream, in Haru library

micheleselea: Yes you're right, probably we can try with Poco::PDF... Or do you think I can fork somehow Haru lib and propose a pull request?

aleks-f: probably best to separate it for now in our own pdf util. libharu maintenance is intermittent, development stopped for a long time, then some work was done, now it seems orphaned again.

but if you want to send them pull, here's the repo: https://github.com/libharu/libharu

micheleselea: ok, I'll do the changes as soon as I can. I don't understand the test failing in windows 32bit...is not dependent at all from PDF because it's seems to be in Foundation

micheleselea: done, have a look if can be ok

aleks-f: can you please rename BmpUtil to simply Util - that way we can add in there anything else in the future, rather than proliferating various util files.

aleks-f: otherwise, it's fine

PR #2184, opened 2018-02-26
PropertyListConfiguration feature (OS X plist)

DraconPern:
A much more cleaned up version of the plist configuration feature.  This replaces #1762.

new class called PropertyListConfiguration subclassed from AbstractConfiguration
Handles OS X plist file type.
Supports all the types for reading as string data, and the following types for writing: string, data, integer, real.
The plist data type is supported with buffers.
Supports nested array.
Events correctly fired for set*() functions.
Test covers reading and writing to plist that's in a format generated by XCode.

### DIFF
--- a/Util/include/Poco/Util/PropertyListConfiguration.h
+++ b/Util/include/Poco/Util/PropertyListConfiguration.h
@@ -1,0 +1,94 @@
+//
+// PropertyFileConfiguration.h
+//
+// Library: Util
+// Package: Configuration
+// Module:  PropertyListConfiguration
+//
+// Definition of the PropertyListConfiguration class.
+//
+// Copyright (c) 2017-2018, Applied Informatics Software Engineering GmbH.
+// and Contributors.
+//
+// SPDX-License-Identifier:	BSL-1.0
+//
+
+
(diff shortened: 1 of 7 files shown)


Util/include/Poco/Util/PropertyListConfiguration.h
```diff
@@ -0,0 +1,94 @@
...
+
+
+#ifndef Util_PropertyListConfiguration_INCLUDED
+#define Util_PropertyListConfiguration_INCLUDED
+
+
+#include "Poco/Util/Util.h"
+
+
+#ifndef POCO_UTIL_NO_XMLplistCONFIGURATION
```
aleks-f: where is this define coming from?
this should probably be simply #ifndef POCO_UTIL_NO_XML_CONFIGURATION, just like the original XMLConfiguration, plus there should be an ifdef so this code only comes in on macOS

DraconPern: Ok, yes it should be POCO_UTIL_NO_XML_CONFIGURATION since it does depend on XML.  It doesn't depend on any OS X code, so I think it should be available for all platforms like ini, json, xml, etc.

PR #2191, opened 2018-02-28
Find libmysqlclient instead of the obsolete libmysqlclient_r

jspricke:
Bug-Debian: https://bugs.debian.org/889625


cmake/FindMySQL.cmake
```diff
@@ -50,7 +50,7 @@ if (WIN32)
 				 ${BINDIR32}/MySQL/*/lib/vs10
 				 $ENV{SystemDrive}/MySQL/*/lib/${libsuffixDist})
 else (WIN32)
-	find_library(MYSQL_LIB NAMES mysqlclient mysqlclient_r
+	find_library(MYSQL_LIB NAMES mysqlclient mysqlclient
```
aleks-f: what is the purpose of two identical names?

jspricke:
Never mind, I'm blind ;).

PR #2193, opened 2018-03-01
Do not compile PDF/src/BmpUtil.cpp

zosrothko:


aleks-f:
Bmputil should not have been there to start with https://github.com/pocoproject/poco/commit/156ac44fb9545502bf6e53c40d6a0a4bdc047bab

PR #2204, opened 2018-03-07
Start POCO as Windows service with parameters fix

DomXC:
Fix of https://github.com/pocoproject/poco/issues/2190 is done using `static ArgVec _argsSvc; ` member, where args are passed to this member using `bool isService(const ArgVec& args); ` or `bool isService(int argc, wchar_t** argv);`.
Removed `int main(int argc, char** argv)` macro and `int ServerApplication::run(int argc, char** argv)` as it seems to be obsolete, because of wchar_t support only.

PR #2205, opened 2018-03-07
Start POCO as Windows service with parameters fix #2190

### LINKED ISSUE #2190: Start POCO as Windows service with parameters
#### Expected behavior
I would like to start an application (based on POCO) as Windows service. This works fine, until I use required parameters.

```
void RRCApplication::defineOptions(OptionSet& options)
{
	ServerApplication::defineOptions(options);

	options.addOption(
		Option("cfg", "f", "Load configuration data from an XML file. E.g.: /cfg=\"C:\\RRC-M\\cfg\\app.xml\"")
		.required(true)
		.repeatable(false)
		.argument("file")
		.callback(OptionCallback<RRCApplication>(this, &RRCApplication::handleConfig)));
}
```

#### Actual behavior
When starting the service, application fails on error 78. It seems, the arguments are not passed.
I attached debugger to application when starting and size of `args` is 1, so no params are passed:

\poco-1.8.0.1-all\Util\src\ServerApplication.cpp, line 179:
```
#if defined(POCO_WIN32_UTF8) && !defined(POCO_NO_WSTRING)
		std::vector<std::string> args;
		for (DWORD i = 0; i < argc; ++i) // argc == 1 (name of executable)
		{
			std::string arg;
			Poco::UnicodeConverter::toUTF8(argv[i], arg);
			args.push_back(arg);
		}
		app.init(args);
#else
```
I have registered the service as 

`sc create MySVC binpath= "C:\MySVC\MySVC.exe /cfg=\"C:\MySVC\cfg\app.xml\"" displayname= "MySVC" start= auto`

In properties of the service (in Windows service list using mmc), I can see the path:
`C:\MySVC\MySVC.exe /cfg="C:\MySVC\cfg\app.xml"`

However, the parameter **cfg** is obviously not passed. When starting application from command line, everything is OK.
The question is, **how to pass params using service**?
Thank you!

#### POCO version
1.8.0.1

#### Compiler and version
VC14

Windows 7

#### Other relevant information
Server application as service

DomXC:
Args are passed to _argsSvc member using isService, params are appended to params of "Path to executable" service property.

### DIFF
--- a/Util/include/Poco/Util/ServerApplication.h
+++ b/Util/include/Poco/Util/ServerApplication.h
@@ -195,7 +195,13 @@ private:
 #endif
 
 	bool hasConsole();
-	bool isService();
+	bool isService(const ArgVec& args);
+#if defined(POCO_WIN32_UTF8) && !defined(POCO_NO_WSTRING)
+	bool isService(int argc, wchar_t** argv);
+#else
+	bool isService(int argc, char** argv);
+#endif
+
 	void beService();
 	void registerService();
 	void unregisterService();
@@ -209,6 +215,7 @@ private:
 	std::string _displayName;
 	std::string _description;
 	std::string _startup;
+	static ArgVec _argsSvc;
 
 	static Poco::Event           _terminated;
 	static SERVICE_STATUS        _serviceStatus; 
--- a/Util/src/ServerApplication.cpp
+++ b/Util/src/ServerApplication.cpp
@@ -56,6 +56,7 @@ namespace Util {
 
 
 #if defined(POCO_OS_FAMILY_WINDOWS)
+Poco::Util::Application::ArgVec ServerApplication::_argsSvc;
 Poco::NamedEvent      ServerApplication::_terminate(Poco::ProcessImpl::terminationEventName(Poco::Process::id()));
 #if !defined(_WIN32_WCE)
 Poco::Event           ServerApplication::_terminated;
@@ -146,7 +147,6 @@ void ServerApplication::ServiceControlHandler(DWORD control)
 	SetServiceStatus(_serviceStatusHandle,  &_serviceStatus);
 }
 
-
 #if defined(POCO_WIN32_UTF8) && !defined(POCO_NO_WSTRING)
 void ServerApplication::ServiceMain(DWORD argc, LPWSTR* argv)
 #else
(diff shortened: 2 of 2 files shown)


aleks-f:
1.8.2 branch is not valid anymore, it has become 1.9.0, but i'm not sure if @obiltschnig wants to include this for that release. so, this should go against either 1.9.0 or (future) 1.9.1 branch

DomXC:
OK, thank you, so, is there any guideline, what to do in this case?
Anyway, build for develop failed (https://github.com/pocoproject/poco/pull/2204), is there a guideline what to do in this case (I know what is wrong...)? Thank you.

aleks-f:
> build for develop failed (#2204), is there a guideline what to do in this case (I know what is wrong...)

fix it and push to your branch, changes will show in the pull and CI will trigger again.

obiltschnig:
Please re-target the pull request to poco-1.9.1 branch.

obiltschnig:

evamueller:
It's not in the 1.10 version, please fix!

obiltschnig:
Would be great to have a proper PR for the `poco-1.10.2` branch.

intensifi:
The following code reads the "Start Parameters" field of the Service Properties Dialog box.  This is very useful for ONE time starts of a service.   This is NOT permanent.  For that you would have to parse the "Image" registry key for the service.  For a FULL feature add, I would suggest that both capabilities by added.  The "Start Parameters" field is very useful for one start parameters such as increasing the logging level for problem resolution.

Yes this was a quick hack copying code from a Microsoft support web page.

NOTE: in 1.10.1 on Windows10, unregistering a service does not work.  an error message complains about only using /unregisterservice once.  No matter though.  I just used

sc stop SERVICENAME
sc delete SERVICENAME

in a batch file. 

-----

int
main( int argc, char** argv )
{
	LPWSTR *szArglist = NULL;
	int nArgs = 0;
#endif

	try
	{
		MyClass serverApplication;

		std::vector< std::string > arguments;
		for ( int i = 0; i < argc; ++i )
		{
			arguments.push_back( argv[ i ] );
		}

		// Windows Services don't process command line arguments like they should!!!
		if ( serverApplication.config().getBool( "application.runAsService", false ) )
		{
			// do service specific things
			szArglist = CommandLineToArgvW(GetCommandLineW(), &nArgs );

			if ( NULL != szArglist )
			{
				for (int i = 0; i < nArgs; i++)
				{
					std::string argument;
					UnicodeConverter::toUTF8( (Poco::UTF16Char*)szArglist[i], argument );
					arguments.push_back(argument);
				}
				LocalFree( szArglist );
				szArglist = NULL;
			}
		}
#endif
		return serverApplication.run( arguments );
	}

	catch ( const Poco::Exception& anException )
	{
		std::cerr << anException.displayText() << std::endl;
		if ( szArglist )
		{
			LocalFree( szArglist );
		}
#endif
	}

	catch ( const std::exception& )
	{
		if ( szArglist )
		{
			LocalFree( szArglist );
		}
#endif
	}

	return Poco::Util::Application::EXIT_SOFTWARE;
}

PR #2214, opened 2018-03-10
Backport of Gradle stuff from the develop branch to produce the NuGet and MSI packages.

zosrothko:
Hi Guenter

This PR adds only gradle stuff to the poco-1.9.0 branch so that both NuGet and MSI packages could be produced. You will need this PR if you want to produce the MSI package on your own Windows machine. I will also back port this gradle stuff to poco-1.9.1


aleks-f:
Since 1.9.0 is released already, i think this should go to 1.9.1

zosrothko:
Yeap it will go also in poco-1.9.1, but that's for Guenter's only usage if he wants to produce and publish the MSI package of poco-1.9.0 on the http://pocoproject.org website

PR #2219, opened 2018-03-13
Poco 1.9.1 version of #2088 - Fix for TCPServerDispatcher race/crash

### LINKED ISSUE #2085: Crash due to race condition in TCPServerDispatcher

Burgch:
#2088 was not compatible with the changes missing from 1.9.1 but in develop. This is a 1.9.1 friendly version. Fixes #2085

### DIFF
--- a/Net/src/TCPServerDispatcher.cpp
+++ b/Net/src/TCPServerDispatcher.cpp
@@ -96,7 +96,7 @@ void TCPServerDispatcher::release()
 
 void TCPServerDispatcher::run()
 {
-	AutoPtr<TCPServerDispatcher> guard(this, true); // ensure object stays alive
+	AutoPtr<TCPServerDispatcher> guard(this); // ensure _rc is decreased when function exits; it was incremented in enqueue()
 
 	int idleTime = (int) _pParams->getThreadIdleTime().totalMilliseconds();
 
@@ -149,6 +149,10 @@ void TCPServerDispatcher::enqueue(const StreamSocket& socket)
 			{
 				_threadPool.startWithPriority(_pParams->getThreadPriority(), *this, threadName);
 				++_currentThreads;
+				// Ensure this object lives at least until run() starts
+				// Small chance of leaking if threadpool is stopped before this
+				// work runs, but better than a dangling pointer and crash!
+				++_rc;
 			}
 			catch (Poco::Exception&)
 			{


Net/src/TCPServerDispatcher.cpp
```diff
@@ -96,7 +96,7 @@ void TCPServerDispatcher::release()
 
 void TCPServerDispatcher::run()
 {
-	AutoPtr<TCPServerDispatcher> guard(this, true); // ensure object stays alive
+	AutoPtr<TCPServerDispatcher> guard(this, false); // ensure _rc is decreased when function exits
```
aleks-f: false does nothing here, so it can be just `guard(this)`. It would be good to point to the relationship of this with enqueue `++_rc`, since it's a functionality in two distant places

Burgch: Updated, thanks.

PR #2238, opened 2018-03-21
cmake: add missing options

yegorich:
Both --no-fpenvironment and --no-wstring are not covered within
CMakeLists.txt. Add both options but with the reversed logic, i.e.
ENABLE_WSTRING and ENABLE_FPENVIRONMENT.

If the option is not enabled, then the related define will be added.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,16 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
     set( CMAKE_BUILD_TYPE "RelWithDebInfo" )
 endif()
 
+option(ENABLE_WSTRING "Enable std::wstring support" ON)
+if (NOT ENABLE_WSTRING)
+    add_definitions(-DPOCO_NO_WSTRING)
+endif()
+
+option(ENABLE_FPENVIRONMENT "Enable floating-point support" ON)
+if (NOT ENABLE_FPENVIRONMENT)
+    add_definitions(-DPOCO_NO_FPENVIRONMENT)
+endif()
+
 # Include some common macros to simpilfy Poco CMake files
 include(PocoMacros)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,16 +73,6 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
     set( CMAKE_BUILD_TYPE "RelWithDebInfo" )
 endif()
 
-option(ENABLE_WSTRING "Enable std::wstring support" ON)
-if (NOT ENABLE_WSTRING)
-    add_definitions(-DPOCO_NO_WSTRING)
-endif()
-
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -73,6 +73,16 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
     set( CMAKE_BUILD_TYPE "RelWithDebInfo" )
 endif()
 
+option(ENABLE_WSTRING "Enable std::wstring support" ON)
+if (NOT ENABLE_WSTRING)
+    add_definitions(-DPOCO_NO_WSTRING)
```
Bjoe: Please use `target_compile_definitions` and add this to the Foundation target. Use also PUBLIC as scope, because this define is used in UnicodeConverter.h and should be exported via Foundation_Config.cmake so that "users" has the correct compile definitions to use PocoFoundation.libs ...

CMakeLists.txt
```diff
@@ -73,6 +73,16 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
...
+endif()
+
+option(ENABLE_FPENVIRONMENT "Enable floating-point support" ON)
+if (NOT ENABLE_FPENVIRONMENT)
+    add_definitions(-DPOCO_NO_FPENVIRONMENT)
```
Bjoe: Please use `target_compile_definitions` and add this to the Foundation target. Use also PUBLIC as scope, because this define is used in FPEnvironment.h and should be exported via Foundation_Config.cmake so that "users" has the correct compile definitions to use PocoFoundation.libs ...

CMakeLists.txt
```diff
@@ -73,6 +73,16 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
     set( CMAKE_BUILD_TYPE "RelWithDebInfo" )
 endif()
 
+option(ENABLE_WSTRING "Enable std::wstring support" ON)
```
Bjoe: Change to `POCO_ENABLE_WSTRING`, I will merge next time the pull request #1437 so POCO properties are in one group in cmake....

CMakeLists.txt
```diff
@@ -73,6 +73,16 @@ if (CMAKE_BUILD_TYPE STREQUAL "")
...
+if (NOT ENABLE_WSTRING)
+    add_definitions(-DPOCO_NO_WSTRING)
+endif()
+
+option(ENABLE_FPENVIRONMENT "Enable floating-point support" ON)
```
Bjoe: Change to `POCO_ENABLE_FPENVIRONMENT`, I will merge next time the pull request #1437 so POCO properties are in one group in cmake ....

yegorich:
@Bjoe please fix them as you're about rename them anyway. Thanks.

Bjoe:
I fixed the findings. Let me see if all tests are green, then I will squash and merge this branch.

PR #2242, opened 2018-03-22
Replace assert by assertTrue

zosrothko:


aleks-f:
@zosrothko it would be good to introduce `CppUnit/include/Poco/CppUnit/*h` as forward headers in 1.x. That way there are no problems when porting tests back from develop to 1.x. Either that, or reinstate `CppUnit/include/CppUnit/*h` in develop, but we should sync the two development lines, it is annoying enough to have to do development on two fronts, then this kind of thing comes on top.

PR #2257, opened 2018-03-29
 SecureSocket: Allow bind with reusePort

proller:
im not sure about reusePort = false , maybe =true is closer to default SocketImpl::bind


aleks-f:
> maybe =true is closer to default SocketImpl::bind

probably, because of port 0

@obiltschnig ?

PR #2262, opened 2018-04-03
Add libcrypto.lib and libssl.lib for VS2017 debug/release_shared targets

feiyunw:
I failed to build poco on VS2017 for release_shared target. My build batch snippet:
```
echo ===== Building POCO
cd /d %DEPSRC%poco
cd
set INCLUDE=%INCLUDE%;%DEPSRC%poco\openssl\VS_120\include;%DEPSRC%mysql\include
set LIB=%LIB%;%DEPSRC%poco\openssl\VS_120\win64\lib\release;%DEPSRC%mysql\lib
call buildwin 150 rebuild shared release x64 nosamples notests msbuild
```
It failed as Crypto and NetSSL_OpenSSL does not link openssl libraries for release_shared (and debug_shared) target.
It looks like the "Automatically link Crypto and OpenSSL libraries" mechanism in Crypto\include\Poco\Crypto\Crypto.h does not work as expected for release_shared (and debug_shared) target.

My patch fixed vs150.vcxproj as example only.
I suggest to fuse the vcxproj files into one, using MSBuild "/p:PlatformToolset=" option in buildwin.cmd to override the settings in vcxproj file, like this:
```
set MSBUILD=MSBuild.exe /m /p:Configuration=Release;Platform=x64;PlatformToolset=v141;WindowsTargetPlatformVersion=%UCRTVersion%;CharacterSet=MultiByte
set MSBUILDDBG=MSBuild.exe /m /p:Configuration=Debug;Platform=x64;PlatformToolset=v141;WindowsTargetPlatformVersion=%UCRTVersion%;CharacterSet=MultiByte
set MSBUILD32=MSBuild.exe /m /p:Configuration=Release;Platform=Win32;PlatformToolset=v141;WindowsTargetPlatformVersion=%UCRTVersion%;CharacterSet=MultiByte
set MSBUILD32DBG=MSBuild.exe /m /p:Configuration=Debug;Platform=Win32;PlatformToolset=v141;WindowsTargetPlatformVersion=%UCRTVersion%;CharacterSet=MultiByte
```
I see no reason to have so many .sln and .vcxproj files for different win32/x64 and VS version combinations.

### DIFF
--- a/Crypto/Crypto_x64_vs150.vcxproj
+++ b/Crypto/Crypto_x64_vs150.vcxproj
@@ -144,7 +144,7 @@ exit 0</Command>
       <CompileAs>Default</CompileAs>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>ws2_32.lib;iphlpapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>ws2_32.lib;iphlpapi.lib;libcrypto.lib;libssl.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <OutputFile>..\bin64\PocoCrypto64d.dll</OutputFile>
       <SuppressStartupBanner>true</SuppressStartupBanner>
       <GenerateDebugInformation>true</GenerateDebugInformation>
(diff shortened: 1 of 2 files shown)


aleks-f:
thanks and sorry, but this won't be merged for several reasons:
- first and foremost, you do not state which version you are using; in any case, the pre-built windows openssl and auto-linking was introduced precisely to avoid distributing what this pull wants to do - it works, please check better what exactly are you doing so it doesn't work for you; open an issue if you don't know how to make it work and we'll look at it

For future contributions

- VS projects for 1.x series of releases are generated using [progen](https://github.com/pocoproject/poco/tree/poco-1.9.0/ProGen), not edited manually
- last but not least,  - master is for the most recent release and is never directly committed to

> I see no reason to have so many .sln and .vcxproj files for different win32/x64 and VS version combinations.

that's ok with us, everyone is entitled to their opinion :)

PR #2267, opened 2018-04-05
As default behavoir, terminate if an unhandle exception occurs in a thread

Bjoe:
A proposal to terminate a thread if an unhandle exception occurs in the default error handler.

### DIFF
--- a/Foundation/src/ErrorHandler.cpp
+++ b/Foundation/src/ErrorHandler.cpp
@@ -36,18 +36,21 @@ ErrorHandler::~ErrorHandler()
 void ErrorHandler::exception(const Exception& exc)
 {
 	poco_debugger_msg(exc.what());
+    std::cerr << "An exception occured: " << exc.what() << '\n';
 }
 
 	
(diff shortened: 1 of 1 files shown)


Foundation/src/ErrorHandler.cpp
```diff
@@ -36,18 +36,21 @@ ErrorHandler::~ErrorHandler()
...
 {
 	poco_debugger_msg(exc.what());
+	throw exc;
 }
 
 
 void ErrorHandler::exception()
 {
 	poco_debugger_msg("unknown exception");
+	throw;
```
aleks-f: we have a discrepancy here because documentation says no exception should be thrown from `exception()` (a vague name IMO); I think that `exception()` set of functions behaving as it is now (no throw), should've been named `handleException()` and then there should be a set of `throwException()`.

In any case, hard to please everyone, i think that's why this was done like this in the first place - let users customize what they want to do
/cc @obiltschnig

Bjoe:
Hi, I added a new quick idea ...

Issue #2271, opened 2018-04-06
HTTPClientSession source IP address


roccocorsi:
#### Expected behavior

When the target host IP family used by HTTPClientSession is not known ahead of time, it would be useful to allow setting both an IPv4 and an IPv6 source IP address, and selecting the appropriate one once the IP family of the host is determined by HTTPClientSession.

This is useful when passing a domain name for the host to HTTPClientSession, and no prior address family verification is done on host address by the application. Adapt to whatever host we might be dealing with, without the application having to make decisions or DNS lookups to figure out address family ahead of time to set appropriate family source IP address.

Useful in environments that are handling mixed address family traffic.

#### Actual behavior

Current implementation supports target IP address and source IP address if they are from the same family (IPv4 or IPv6).  If mismatched IPv4 (target) and IPv6 (source) values; or the reverse case IPv6 (target) and IPv4 (source), then a family exception will be thrown.

#### Steps to reproduce the problem

Here is a simple example using the SetSourceIP sample program to force the bad IPv4/IPv6 situation, but normally the address in uri would be a domain name where we don't know ahead of time if the DNS is going to return an IPv4 or IPv6 address.

./SetSourceIP --sourceip 192.168.15.4 http://[fd17:625c:f037:a80f:6dbe:34f6:2afa:66e9]:1500/
Host: fd17:625c:f037:a80f:6dbe:34f6:2afa:66e9
Port: 1500
Path: /
Exception: name (Net Exception) message [Address family not supported]

#### POCO version
2.0.0

#### Other relevant information

Related and an enhancement to issue https://github.com/pocoproject/poco/issues/1475

roccocorsi:
Aleks, made a combined commit as requested.

roccocorsi:
I did get emails that you made some comments, but I can't seem to find the comments directly on github to respond to them. Sorry for my lack of resourcefulness.  :-)

Also someone commented on my cmake modification (two commented lines really) to compile in C++11 the SetSourceIP sample program.  I just put that there is case it might help someone compiling just the SetSourceIP sample program.  Either it doesn't work for my environment (OpenSuse Leap 42.3) without that line or I just don't know how to use cmake.

roccocorsi:
Also Günter marked the commits for 1.10.0. but I thought this was going into 2.0.0 because it is C++11.  Did something change and 1.10.0 will now be C+11?

Issue #2270, opened 2018-04-06
HTTPClientSession not supporting binding source address for proxy connect


roccocorsi:
#### Expected behavior

When using HTTP proxy with HTTPClientSession class, expecting that source IP address binding is used when connecting to proxy, just like when no proxy is used and connecting directly.

#### Actual behavior

HTTP proxy connects without source IP address binding even if an IPAddress is set with setSourceAddress().

#### POCO version
2.0.0

#### Other relevant information

Related to issue https://github.com/pocoproject/poco/issues/1475

roccocorsi:
Should I rebase my two fixes/features to top of develop?

roccocorsi:
Aleks, made a combined commit as requested.

roccocorsi:
I did get emails that you made some comments, but I can't seem to find the comments directly on github to respond to them. Sorry for my lack of resourcefulness.  :-)

Also someone commented on my cmake modification (two commented lines really) to compile in C++11 the SetSourceIP sample program.  I just put that there is case it might help someone compiling just the SetSourceIP sample program.  Either it doesn't work for my environment (OpenSuse Leap 42.3) without that line or I just don't know how to use cmake.

roccocorsi:
Also Günter marked the commits for 1.10.0. but I thought this was going into 2.0.0 because it is C++11. Did something change and 1.10.0 will now be C+11?

aleks-f:
https://github.com/pocoproject/poco/commit/804ce9ee594169d3740db83ec9dd1c09836e5a1e
https://github.com/pocoproject/poco/commit/cd850522317074248c492d291c4d81bc6f7104f7

PR #2273, opened 2018-04-08
HTTPClientSession to support dual IPv4/IPv6 family when setting source IP address #2271

roccocorsi:
Fix for issue: HTTPClientSession to support dual IPv4/IPv6 family when setting source IP address https://github.com/pocoproject/poco/issues/2271

### DIFF
--- a/NetSSL_OpenSSL/samples/SetSourceIP/CMakeLists.txt
+++ b/NetSSL_OpenSSL/samples/SetSourceIP/CMakeLists.txt
@@ -1,5 +1,8 @@
 set(SAMPLE_NAME "SetSourceIP")
 
+# uncomment following line, might solve compiling issues
+#set(CMAKE_CXX_STANDARD 11)
+
 set(LOCAL_SRCS "")
 aux_source_directory(src LOCAL_SRCS)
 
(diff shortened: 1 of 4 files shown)


roccocorsi:
Fix is built on top of my other commit related to https://github.com/pocoproject/poco/pull/2272 and the issue https://github.com/pocoproject/poco/issues/2270

NetSSL_OpenSSL/samples/SetSourceIP/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 set(SAMPLE_NAME "SetSourceIP")
 
+# uncomment following line, might solve compiling issues
+#set(CMAKE_CXX_STANDARD 11)
```
aleks-f: this should not be needed, since there is automatic [detection](https://github.com/pocoproject/poco/blob/develop/cmake/CXX1x.cmake)
@Bjoe is cmake maintainer, you may want to check with him

aleks-f:
@roccocorsi , I made a bit of a mess here - I merged this one before #2272, which now conflicts, and this one does not give me merge option anymore
can you please sort it out and perhaps send both as a single pull

Issue #2012, opened 2017-11-22
Data/SQLite: Exception messages contain duplicate text


obiltschnig:
SQLite exception displayText() contains duplicate messages, example: "Abort due to constraint violation: Abort due to constraint violation"

Problem is Utility::throwException() method. Need to review throw statements and remove unnecessary exception constructor arguments leading to duplicate text.

PR #2276, opened 2018-04-09
Added check for null pointer before accessing and leaving SQLite mutex.

jiajen:
A null pointer dereference occurs when a SQLite Session fails to be constructed. This can be recreated with the following code:
```
try
{
	Poco::SQL::Session tmp(Poco::SQL::SQLite::Connector::KEY, "\\/some\\/invalid\\/path\\/sqlite.db", 1);
}
catch (Poco::SQL::ConnectionFailedException&)
{
}
```
The issue occurs when the database cannot be created and attempts to throw an exception. However, since all exceptions invoke a SQLite mutex (since the fix for issue #2012), `sqlite3_db_mutex(pDB)` causes a segmentation fault when `pDB` is null which can occur if the database cannot be created. Checking for a null pointer before entering and leaving the SQLite mutex solves the problem.

### DIFF
--- a/SQL/SQLite/src/Utility.cpp
+++ b/SQL/SQLite/src/Utility.cpp
@@ -59,15 +59,18 @@ Utility::TypeMap Utility::_types;
 Poco::Mutex Utility::_mutex;
 
 
-Utility::SQLiteMutex::SQLiteMutex(sqlite3* pDB): _pMutex(sqlite3_db_mutex(pDB))
+Utility::SQLiteMutex::SQLiteMutex(sqlite3* pDB)
 {
-	sqlite3_mutex_enter(_pMutex);
+	_pMutex = (pDB) ? sqlite3_db_mutex(pDB) : NULL;
+	if (_pMutex)
+		sqlite3_mutex_enter(_pMutex);
 }
 
 
 Utility::SQLiteMutex::~SQLiteMutex()
 {
-	sqlite3_mutex_leave(_pMutex);
+	if (_pMutex)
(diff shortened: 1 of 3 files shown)


SQL/SQLite/src/Utility.cpp
```diff
@@ -59,15 +59,18 @@ Utility::TypeMap Utility::_types;
 Poco::Mutex Utility::_mutex;
 
 
-Utility::SQLiteMutex::SQLiteMutex(sqlite3* pDB): _pMutex(sqlite3_db_mutex(pDB))
+Utility::SQLiteMutex::SQLiteMutex(sqlite3* pDB)
 {
-	sqlite3_mutex_enter(_pMutex);
+	_pMutex = (pDB) ? sqlite3_db_mutex(pDB) : NULL;
```
aleks-f: a couple of things:
1) this is ok, but please leave it in initialization
2) use 0, not NULL

aleks-f:
@jiajen thanks, just so it is properly tracked, would you mind opening issue and sending the same fix to [poco-1.9.1](https://github.com/pocoproject/poco/tree/poco-1.9.1) branch

SQL/SQLite/src/Utility.cpp
aleks-f: strictly speaking `nullptr` would be the right thing to do, but since 1.x releases do not yet unconditionally support c++11, let's have 0

jiajen: Added the changes. Thanks for pointing out!

Issue #2285, opened 2018-04-12
SQLite::Connector::open() crashes on db file with non existing directory


mkrivos:
It is expected to throw exception. But crashes on nullptr access. Patch is on the way.

To test you can try open non existing file "foo/bar/xy.db".

Affected versions: Poco from 1.8.x to latest develop

All operating systems.

Stack Dump:

#0  sqlite3_db_mutex (db=0x0) at /opt/rsys/poco/SQL/SQLite/src/sqlite3.c:142708
#1  0x00007f752fcd64cc in Poco::SQL::SQLite::Utility::SQLiteMutex::SQLiteMutex (this=0x7ffe37fd7690, pDB=<optimized out>) at /opt/rsys/poco/SQL/SQLite/src/Utility.cpp:62
#2  0x00007f752fcd652e in Poco::SQL::SQLite::Utility::lastError[abi:cxx11](sqlite3*) (pDB=pDB@entry=0x0) at /opt/rsys/poco/SQL/SQLite/src/Utility.cpp:162
#3  0x00007f752fcd85e8 in Poco::SQL::SQLite::Utility::throwException (pDB=0x0, rc=rc@entry=14, addErrMsg="") at /opt/rsys/poco/SQL/SQLite/src/Utility.cpp:229
#4  0x00007f752fcd2a22 in Poco::SQL::SQLite::SessionImpl::open (this=this@entry=0x55bc27fa6cd0, connect="") at /opt/rsys/poco/SQL/SQLite/src/SessionImpl.cpp:168
#5  0x00007f752fcd3d57 in Poco::SQL::SQLite::SessionImpl::SessionImpl (this=0x55bc27fa6cd0, fileName=..., loginTimeout=60) at /opt/rsys/poco/SQL/SQLite/src/SessionImpl.cpp:55
#6  0x00007f752fcbb3ee in Poco::SQL::SQLite::Connector::createSession (this=<optimized out>, connectionString="foo/bar/nonexisting.db", timeout=60) at /opt/rsys/poco/SQL/SQLite/src/Connector.cpp:49
#7  0x00007f752f9697e1 in Poco::SQL::SessionFactory::create (this=0x7f752fc75840 <Poco::SQL::SessionFactory::instance()::sf>, key="sqlite", connectionString="foo/bar/nonexisting.db", timeout=timeout@entry=60) at /opt/rsys/poco/SQL/src/SessionFactory.cpp:73
[...]

PR #2286, opened 2018-04-12
Fix for issue #2285

mkrivos:


SQL/SQLite/testsuite/src/SQLiteTest.cpp
```diff
@@ -1374,6 +1374,18 @@ void SQLiteTest::testEmptyDB()
 }
 
 
+void SQLiteTest::testNonexistingDB()
+{
+	try
+	{
+		Session tmp (Poco::SQL::SQLite::Connector::KEY, "foo/bar/nonexisting.db");
```
aleks-f: there should be a `fail()` here.

aleks-f:
@mkrivos while you're in there, can you also look at this CI failure, I think it's coming from some recent change:
```There was 1 error: 
 1: N7CppUnit10TestCallerI10SQLiteTestEE.testIllegalFilePath
    "N4Poco27AssertionViolationExceptionE: Assertion violation"
    in "<unknown>", line -1```

SQL/SQLite/testsuite/src/SQLiteTest.cpp
mkrivos: Fixed.

mkrivos:
All Fixed.

PR #2318, opened 2018-05-04
- add HTTP_PURGE constant and unit test

kedarX:
PURGE method is useful for clients that deal with proxies and server's that support purge of resources.


kedarX:
Closed due to merge conflict issue during a rebase pull.

PR #2336, opened 2018-05-17
Omit ContentLength in WebSocket accept response

OlegHahm:
According to https://tools.ietf.org/html/rfc7230#section-3.3.2
(see https://stackoverflow.com/a/44553521/8647581 for details)


lothrop:
Is there anything I can do to expediate the merge of this PR? The current implementation does not work with the (standards-conform) client we are using. It seems like a straightforward and correct commit.

obiltschnig:
Need to merge into 1.10 as well.

OlegHahm:
Should I open another PR against a release branch or what is the procedure?

PR #2342, opened 2018-05-21
BorringSSL Fork

AnDevi:
Those are changes necessary to build Poco Library with BoringSSL instead of OpenSSL. 
I tested that it's works on MacOs.


obiltschnig:
Please re-target pull request against poco-1.10.0 branch.

PR #2359, opened 2018-06-05
Improve CMake script for databases on Windows

OgreTransporter:
I've checked the database scripts and added some improvements.

MySQL: If you are working with the server instead of the connector $ENV{MYSQL_DIR}/lib is helpful to find the library

PostgreSQL: Version 10 is working at my test project.

ODBC: Since Microsoft moves headers and libs from "SDK" to "Kit", the hint path changed. It is a little bit tricky to use the same kit for header and library, but this code should work.


OgreTransporter:
Is it possible to get a feedback after more than 6 weeks?

cmake/FindODBC.cmake
```diff
@@ -13,45 +13,79 @@
...
-	/usr/local/iodbc/lib
-	"C:/Program Files/ODBC/lib"
-	"C:/ODBC/lib/debug"
-	"C:/Program Files (x86)/Microsoft SDKs/Windows/v7.0A/Lib"
-	DOC "Specify the ODBC driver manager library here."
-)
-
-# MinGW find usually fails
-if(MINGW)
+if(WIN32)
```
aleks-f: lots of windows version-specific stuff here, but is it really needed? unlike other platforms, I've never seen a problem finding ODBC headers or libraries on windows

OgreTransporter: > I've never seen a problem finding ODBC headers or libraries on windows

Have you got a pure Win10-Installation without the SDK for Win7 and Win8? In that case the hints
```
"C:/Program Files/Microsoft SDKs/Windows/v7.0/include"
"C:/Program Files/Microsoft SDKs/Windows/v6.0a/include"
```
will not work. Win10 is organizing the SDKs in "kits" for each version, so you have to look for odbc32.lib and use the corresponding include folder.

Bjoe: Maybe we should also include to set the paths `HINTS` via an environment settings ... like in MySQL ...

OgreTransporter: When installing MySQL the corresponding environment variables for MySQL are created. When ODBC is installed, environment variables are also created. Except when using the Microsoft SDKs under Windows. There are no environment variables for ODBC. And I don't know what it looks like on OSX.

OgreTransporter:
If you don't like the patch, just decline it.

aleks-f:
I'm fine with it, waiting for @Bjoe review

aleks-f:
@OgreTransporter  this broke pretty much all develop travis [builds](https://travis-ci.org/pocoproject/poco/builds/431792903), can you take a look please?

aleks-f:
never mind, it seems [the previous merge](https://github.com/pocoproject/poco/commit/1bf3b1c06ba489fc2dd01156e3368b0fcde0ab86) broke it

OgreTransporter:
@Bjoe 
Yes, there are multiple locations for ODBC. For my local office PC:

- C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\um\x86/x64
- C:\Program Files (x86)\Windows Kits\10\Lib\10.0.10240.0\um\x86/x64
- C:\Program Files (x86)\Windows Kits\10\Lib\10.0.17134.0\um\x86/x64

Each library has a corresponding Include directory. Within Windows 10, you may be able to use includes of 10.0.10240.0 and the library from 10.0.17134.0, but mixing 8.1 and 10.x definitely doesn't work. Currently available kits for Windows 10:

- 10.0.10240.0
- 10.0.10586.0
- 10.0.14393.0
- 10.0.15063.0
- 10.0.16299.0
- 10.0.17134.0

A path like ```[HKEY_LOCAL_MACHINE\\SOFTWARE\\ODBC\\ODBClib;InstallPath]``` doesn't exist. That's why I've adapted the CMake code in this way for Windows.

PR #2360, opened 2018-06-05
MinGW support

Good-Pudge:
All components work excluding:
* ApacheConnector (MSYS2 required)
* MySQL (MariaDB connector doesn't support MinGW)

Message compiler automatically disable in MinGW if it is not exists.
Add wrapper for `wmain()` that will be used in Application and ServerApplication. It can be disabled by `POCO_NO_WMAIN_WRAPPER` definition.
Change logic of NetworkInitialzer in `Util.h`.

Fix several tests.

Fix missing argument name in functions "setEscapeUnicode" in JSON package.


Crypto/samples/genrsakey/CMakeLists.txt
```diff
@@ -1,2 +1,4 @@
 add_executable(genrsakey src/genrsakey.cpp)
 target_link_libraries(genrsakey PUBLIC Poco::Crypto Poco::Util Poco::XML)
+
+POCO_ENABLE_EXE_WMAIN(genrsakey)
```
aleks-f: @Bjoe can you please confirm this is the right thing to do

Foundation/src/Error.cpp
```diff
@@ -39,12 +39,14 @@ namespace Poco {
...
 		std::string errMsg;
 		DWORD dwFlg = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
+
 	#if !defined(POCO_NO_WSTRING)
 		LPWSTR lpMsgBuf = 0;
-		if (FormatMessageW(dwFlg, 0, errorCode, 0, (LPWSTR) & lpMsgBuf, 0, NULL))
+		if (FormatMessageW(dwFlg, 0, errorCode, 0, (LPWSTR) & lpMsgBuf, 0, NULL)) {
 			UnicodeConverter::toUTF8(lpMsgBuf, errMsg);
-	#endif
+		}
```
aleks-f: style

Foundation/testsuite/src/FileTest.cpp
```diff
@@ -20,6 +20,9 @@
 #include <set>
 
 
+#define MAX_PATH 260
```
aleks-f: please shield this with `#ifndef MAX_PATH`

Net/CMakeLists.txt
```diff
@@ -18,11 +18,11 @@ set_target_properties( Net
...
 # Windows and WindowsCE need additional libraries
 if(WIN32)
-    target_link_libraries(Net PUBLIC "iphlpapi.lib")
+    target_link_libraries(Net PUBLIC "iphlpapi")
     if(WINCE)
-        target_link_libraries(Net PUBLIC "ws2.lib")
+        target_link_libraries(Net PUBLIC "ws2")
     else()    
-        target_link_libraries(Net PUBLIC "ws2_32.lib")
+        target_link_libraries(Net PUBLIC "ws2_32")
```
aleks-f: @bjoe?

NetSSL_Win/src/Context.cpp
```diff
@@ -91,7 +91,11 @@ void Context::init()
 	_hMemCertStore = CertOpenStore(
 					CERT_STORE_PROV_MEMORY,   // The memory provider type
 					0,                        // The encoding type is not needed
-					NULL,                     // Use the default provider
+	#ifdef _MSC_VER                           // Use the default provider
+					NULL,
+	#else
+					0,
+	#endif
```
aleks-f: please just use 0 here, but why is NULL not good here and is in the last argument?

NetSSL_Win/src/Context.cpp
```diff
@@ -101,7 +105,11 @@ void Context::init()
 	_hCollectionCertStore = CertOpenStore(
 			CERT_STORE_PROV_COLLECTION, // A collection store
 			0,                          // Encoding type; not used with a collection store
-			NULL,                       // Use the default provider
+	#ifdef _MSC_VER                     // Use the default provider
+			NULL,
+	#else
+			0,
+	#endif
```
aleks-f: same as above

NetSSL_Win/src/Context.cpp
Good-Pudge: Because MinGW required number here

Good-Pudge:
Please, cancel `build develop-3561` at AppVeyor as I pushed new commit.

cmake/PocoMacros.cmake
```diff
@@ -212,7 +214,14 @@ macro(POCO_MESSAGES out name)
...
         list(APPEND ${out} ${ARGN})
 
-    endif (WIN32)
+    endif (WIN32 AND CMAKE_MC_COMPILER)
+endmacro()
+
+
+macro(POCO_ENABLE_EXE_WMAIN target_name)
+    if(MINGW)
+        set_target_properties("${target_name}" PROPERTIES LINK_FLAGS "-municode")
```
Bjoe: `-municode` is needed to compile unicode applications. Should be activated when Uniicode support is enabled in poco or? 
I found also this 

> Help! I get an error: unrecognized command line option "-municode"!

in [MinGW wikiUnicode applications ](https://sourceforge.net/p/mingw-w64/wiki2/Unicode%20apps/)

Normaly CMake takes care that it works on all compilers and platforms. See also CMake rulse:

> 1. Don't make any assumption about the platform or compiler!

I would like to investigate if cmake has any support for such compile flag.

Net/CMakeLists.txt
Bjoe: Well I don't know yet if this will work with microsoft compile. But ws2_32 is better then ws2_32.lib .... I would like to investigate if cmake support .lib suffix for microsoft compiler...

Crypto/samples/genrsakey/CMakeLists.txt
Bjoe: See my review comment...

cmake/PocoMacros.cmake
Good-Pudge: Try to compile application with `wmain` using MinGW. You'll get `undefined reference to WinMain`.

CMakeLists.txt
```diff
@@ -81,6 +85,10 @@ else()
     option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 endif()
 
+if(BUILD_SHARED_LIBS AND MINGW)
+    add_definitions(-D_DLL)
```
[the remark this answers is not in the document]
Good-Pudge: Define `_DDL` use almost all packages and dependencies. Do you offer to define it for every target?

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.2.0)
 
 project(Poco)
 
+if(WINCE AND MINGW)
+    message(FATAL_ERROR "MinGW does not support WinCE")
```
Good-Pudge: I think I should delete this. CMake will cope with this.

Bjoe: I think also we should remove this.... mingw will not work at all on wince ...

CMakeLists.txt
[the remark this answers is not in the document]
Bjoe: You should ask your self, if it is mandatory to set also _DLL into your project they use the compiled poco library ? Mostly I checked in the code if this preprocessor definiton is in the header file .... if not, then it can be `PRIVATE` (it is only need to compile the poco librarie) ... if it is in the header file, then you should investigate is this header file published with the compiled library ... if yes then it is mostly mandatory that you added this as `PUBLIC` because it is needed to use the compiled poco library into your project. You should not added these to every target. Added with `target_compile_definition` with `PUBLIC` there where it is used and this definiton will be provided to the other targets they use this target. For example for me we should add this on the Poco::Foundation target. All of the other targets depends on Poco::Foundation and get this definiton ...See also [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html?highlight=target_compile_#command:target_compile_definitions) I found that _DLL is used in Crypto.h, CppUnit.h, Foundation,h, NetSSL.h, pngconf.h .... I see also we have POCO_DLL ... hmmm is this the same?

Net/src/PollSet.cpp
```diff
@@ -44,7 +44,6 @@
 namespace Poco {
 namespace Net {
 
-
```
Bjoe: Please read contribution documentation about code formatting

MongoDB/include/Poco/MongoDB/MongoDB.h
```diff
@@ -31,8 +31,6 @@
 // MongoDB_API functions as being imported from a DLL, whereas this DLL sees symbols
 // defined with this macro as being exported.
 //
-
-
```
Bjoe: Please read contribution documentation about code formatting

SQL/include/Poco/SQL/SQL.h
```diff
@@ -31,7 +31,6 @@
 // Poco_SQL_API functions as being imported from a DLL, whereas this DLL sees symbols
 // defined with this macro as being exported.
 //
-
```
Bjoe: Please read contribution documentation about code formatting :-)

cmake/PocoMacros.cmake
Bjoe: But wmain is needed when you will have unicode support or? For me it looks like this depends on POCO_ENBALE_WSTRING .... but I'm still investigating...

Util/include/Poco/Util/ServerApplication.h
```diff
@@ -232,7 +232,7 @@ class Util_API ServerApplication: public Application
 //
 // Macro to implement main()
 //
-#if defined(_WIN32) && !defined(__MINGW32__)
+#if defined(_WIN32) && !defined(POCO_NO_MINGW_UNICODE)
```
Bjoe: No unicode ? I was expected you should have wmain for unicode support?

Good-Pudge: Option `-municode` affect to WinAPI too. For example `InetPton`. It is define to InetPtonA or to InetPtonW. If option `-municode` set, InetPton will be InetPtonW. If not, InetPton will be InetPtonA.
What does mean `A` or `W`:
https://msdn.microsoft.com/ru-ru/library/windows/desktop/dd374089(v=vs.85).aspx

zosrothko: As stated in the URL
```
A Windows code page version with the letter "A" used to indicate "ANSI"
A Unicode version with the letter "W" used to indicate "wide"
```

aleks-f:
I'm ok with changes, @zosrothko there's appveyor NetSSL test failure but I don't think it is related to this pull, can you check?

Crypto/testsuite/src/CryptoTestSuite.cpp
```diff
@@ -10,7 +10,7 @@
 
 #include "Poco/Platform.h"
 // see https://github.com/openssl/openssl/blob/master/doc/man3/OPENSSL_Applink.pod
-#if defined(POCO_OS_FAMILY_WINDOWS)
+#if defined(_MSC_VER)
```
Bjoe: Sometimes I see, we use `POCO_OS_FAMILY_WINDOWS` and sometimes we use `_MSC_VER` .... May I ask what is the difference?

Net/CMakeLists.txt
```diff
@@ -18,12 +18,21 @@ set_target_properties( Net
...
-        target_link_libraries(Net PUBLIC "ws2.lib")
+        target_link_libraries(Net PUBLIC "ws2")
     else()    
-        target_link_libraries(Net PUBLIC "ws2_32.lib")
-    endif()    
+        target_link_libraries(Net PUBLIC "ws2_32")
+    endif()
+
+    if(MINGW) # Needed by WinAPI in MinGW
+        string(REGEX MATCH "^[0-9]+\\.[0-9]+" WIN_VERSION ${CMAKE_SYSTEM_VERSION})
```
Bjoe: In cmake it is possible to verify version numbers with `VERSION_LESS` etc. See https://cmake.org/cmake/help/latest/command/if.html?highlight=version_less

Crypto/testsuite/src/CryptoTestSuite.cpp
Good-Pudge: OpenSSL for MinGW has no `#include "openssl/applink.c"`.

Net/CMakeLists.txt
Good-Pudge: It doesn't check or verify version, it calculate it from version like `6.0.141241` to `0x0600`.
* Take first two numbers `6` and `0`, so we get `6.0`
* It take major version, so we get `6`
* It take minor version, so we get `0`
* Transform these two numbers to version like WinAPI. `256 * 6 + 0 = 1536` or `0x0600`
If windows version differ, WinAPI will be different too.

Bjoe: ok

MongoDB/include/Poco/MongoDB/MongoDB.h
Bjoe: ok

Crypto/testsuite/src/CryptoTestSuite.cpp
Bjoe: ok

SQL/include/Poco/SQL/SQL.h
Bjoe: ok

filimonov:
> MySQL (MariaDB connector doesn't support MinGW)

Did you try that? 
```
pacman -S --needed mingw-w64-x86_64-openssl mingw-w64-x86_64-libmariadbclient
```
And after that apply patch from https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-poco/012-find-mysql.patch

Also you can review other patches from https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-poco

Good-Pudge:
But how use `libmariadbclient` outside of MSYS2?

filimonov:
Static linking?

filimonov:
Here are the package file with patches needed to support mariadbcliect on mingw platform https://github.com/Alexpux/MINGW-packages/tree/bdc31c969f802816604254c938c524d3a8283530/mingw-w64-libmariadbclient

Good-Pudge:
About tests. As I understand, it are require run some services, eg. Redis, MySql, etc.

Good-Pudge:
And about patches. I'll try to use them myself on [mariadb-connector-c](https://github.com/MariaDB/mariadb-connector-c).

PR #2399, opened 2018-07-20
NTPPacket fix and NTPServer

micheleselea:


Net/src/NTPServer.cpp
```diff
@@ -0,0 +1,145 @@
+#include "Poco/Net/NTPServer.h"
```
aleks-f: naming, opening braces on new line, only english comments please, remove not used commented code

micheleselea: done: for NTPServer test I'll do it next week

Net/src/NTPPacket.cpp
```diff
@@ -16,10 +16,12 @@
 #include "Poco/Net/NetException.h"
 #include "Poco/Timestamp.h"
 #include "Poco/ByteOrder.h"
-#ifndef POCO_OS_FAMILY_WINDOWS
-#include <sys/time.h>
-#else
+#if defined(POCO_OS_ANDROID)
+#include <time.h>
+#elif defined(POCO_OS_FAMILY_UNIX)
```
aleks-f: I'm not sure how this compiled, but it definitely does not look right

micheleselea: Ok I'm going to change it

micheleselea: but, now is this way:

#if defined(POCO_OS_ANDROID)
#include <time.h>
#elif defined(POCO_OS_FAMILY_UNIX)
#include <winsock2.h>
#else
#include <sys/time.h>
#endif

do you think is not correct?

aleks-f: `winsock2.h` on unix? I do not think that's correct. But why are we including windows socket header at all here? it should already be included via Poco::Net headers

micheleselea: ah ah yes! correct :) should be POCO_OS_WINDOWS I'm going to correct...don't know how can compile :)

Issue #2401, opened 2018-07-24
Poco::Net::MailMessage::read stucks while reading message without final multipart boundary


avoronkov:
#### Expected behavior
Poco::Net::MailMessage::read finishes when reading messages without final multipart boundary.

#### Actual behavior
Process stucks in infinite loop with 100% CPU consumption on reading with following strace:
```
read(3, "\n", 8191)                     = 1
read(3, "", 8191)                       = 0
read(3, "", 8191)                       = 0
lseek(3, -1, SEEK_CUR)                  = 7252
read(3, "\n", 8191)                     = 1
read(3, "", 8191)                       = 0
read(3, "", 8191)                       = 0
lseek(3, -1, SEEK_CUR)                  = 7252
```

#### Steps to reproduce the problem
1. Compile and run the following program:
```
#include "Net/MailMessage.h"
#include <fstream>

int main() {
    std::ifstream ifs("mail_example.txt");
    Poco::Net::MailMessage message;
    message.read(ifs);
    return 0;
}
```
with `mail_example.txt` containing the following:
```
To: user@example.org
Subject: Test parsing incomplete multipart message
X-Mailer: Microsoft Office Outlook, Build 17.551210
From: user@examle.com
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="ec3ac6be4c8e86c0d91b89466e1f7bd7"
Date: Mon, 23 Jul 2018 14:29:45 +0200 (CEST)

Content-Transfer-Encoding: 7bit
This is a MIME encoded message.

--ec3ac6be4c8e86c0d91b89466e1f7bd7
Content-Type: text/html; charset="iso-8859-1"
Content-Transfer-Encoding: 8bit

<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
<html>
<head> <title>Looks like outlook sends such emails sometimes</title> </head>
<body>
<div><span style="background-color: #ffffff; color: #0000ff;"><strong></strong></span></div>
</body>
</html>
```
#### POCO version
1.9.0

#### Compiler and version
GCC 6.3.0
#### Operating system and version
CentOS 7.5
#### Other relevant information
Root cause: MessageHeader::read incorrectly puts eof into istream which results in storing char 255 in stream.
Patch:
```
--- ./Net/src/MessageHeader.cpp.orig    2018-03-08 21:28:06.000000000 +0700
[...]

PR #2405, opened 2018-07-26
(#2401) Poco::Net::MailMessage::read stucks while reading message 

avoronkov:
without final multipart boundary
https://github.com/pocoproject/poco/issues/2401

### DIFF
--- a/Net/src/MessageHeader.cpp
+++ b/Net/src/MessageHeader.cpp
@@ -113,7 +113,10 @@ void MessageHeader::read(std::istream& istr, RecipientList* pRecipients)
 		++fields;
 
 	}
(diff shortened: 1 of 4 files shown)


Net/src/MessageHeader.cpp
```diff
@@ -113,7 +113,9 @@ void MessageHeader::read(std::istream& istr, RecipientList* pRecipients)
 		++fields;
 
 	}
-	istr.putback(static_cast<char>(ch));
+	if (ch != eof) {
```
aleks-f: coding style

avoronkov:
Fixed codestyle, add check for `good()`.

PR #2417, opened 2018-08-03
Added missing IPv6 methods to SecureSocketImpl

dudym:
Solution for https://github.com/pocoproject/poco/issues/2418
I noticed that bind6 was missing for SecureSocketImpl. Simply a matter of adding the method in a similar manner like bind. Solved it for me


amorlang:
Any idea, when this is going life?

amorlang:
Thanks for considering this for the next release. The commit 06d9846 works on my systems like expected. For completeness sake "SecureServerSocketImpl" and "SecureSocketImpl" are lacking the bind6 function with the reuse address flag. Not a big thing.

Issue #2356, opened 2018-06-02
MinGW support


ark0f:
Why POCO doesn't have MinGW support? I think it is not difficult. I can even add support personally.

aleks-f:
if you want to own it, it's yours. send pulls with fixes and if they don't break other builds, they will be merged. see with @zosrothko if there is space in CI environment for it

zosrothko:
On AppVeyor, there is space.. on VSTS, it remains 25GB but one should keep it for VS addition. This I would suggest to CI the MinGW on AppVeyor.
@Good-Pudge Please ask for a account on Slack to easy the communication

aleks-f:
yes, appveyor will be fine. @Good-Pudge I have sent you invite for slack

qwertychouskie:
For the error in PocoMacros.cmake, these modified code chunks should work, at least with 1.9.0:
```
if (WIN32)
  # cmake has CMAKE_RC_COMPILER, but no message compiler
  if ("${CMAKE_GENERATOR}" MATCHES "Visual Studio")
    # this path is only present for 2008+, but we currently require PATH to
    # be set up anyway
    get_filename_component(sdk_dir "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Microsoft SDKs\\Windows;CurrentInstallFolder]" REALPATH)
    get_filename_component(kit_dir "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot]" REALPATH)
    get_filename_component(kit81_dir "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot81]" REALPATH)
    if (X64)
      set(sdk_bindir "${sdk_dir}/bin/x64")
      set(kit_bindir "${kit_dir}/bin/x64")
      set(kit81_bindir "${kit81_dir}/bin/x64")
    else (X64)
      set(sdk_bindir "${sdk_dir}/bin")
      set(kit_bindir "${kit_dir}/bin/x86")
      set(kit81_bindir "${kit81_dir}/bin/x86")
    endif (X64)
  endif ()
  find_program(CMAKE_MC_COMPILER mc.exe HINTS "${sdk_bindir}" "${kit_bindir}" "${kit81_bindir}"
    DOC "path to message compiler")
  if (NOT CMAKE_MC_COMPILER)
[...]

Issue #2428, opened 2018-08-15
POCO Net build fails with clang-cl 6.00 on Windows


rokel:
#### Expected behavior

Can compile POCO on Windows with clang-cl 6.00.

#### Actual behavior

 Build fails with the following:

```
[build] ..\Core\externals\OSAL\externals\POCO\Foundation\include\Poco/Tuple.h(4492,45):  error: missing 'typename' prior to dependent type name 'TypeWrapper<IPAddress>::TYPE'
[build]                 typename TypeWrapper<T1>::CONSTTYPE& t1 = POCO_TYPEWRAPPER_DEFAULTVALUE(T1),
[build]                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] ..\Core\externals\OSAL\externals\POCO\Foundation\include\Poco/Tuple.h(29,42):  note: expanded from macro 'POCO_TYPEWRAPPER_DEFAULTVALUE'
[build] #define POCO_TYPEWRAPPER_DEFAULTVALUE(T) TypeWrapper<T>::TYPE()
[build]                                          ^~~~~~~~~~~~~~~~~~~~
```

As clang-cl defines _MSC_VER for compatibility, but this is then used in Tuple.h to omit the `typename` keyword, presumably for compatibility with MSVC.

Note this is the only build error, which is impressive!

#### POCO version

1.9.1

#### Compiler and version

Clang 6.0.0

#### Operating system and version

Windows 10 x64

#### Other relevant information

Seems like this is an easy fix - remove the `#ifdef` conditional [here](https://github.com/pocoproject/poco/blob/poco-1.9.1/Foundation/include/Poco/Tuple.h#L28) and never omit typename.

At the very least it seems to work fine with MSVC 2017. If it fails with earlier MSVC compilers then this fix seems equally easy, only omit if version is less than 2017.

aleks-f:
as if it's not enough to have multiple compilers to deal with,  now we also have a compiler pretending to be all other compilers. anyway, thanks and see if the fix works for you

rokel:
Indeed. Thanks for the quick fix, worked perfectly.

PR #2433, opened 2018-08-20
Fix "redeclared without dllimport attribute..." warning for MinGW

ark0f:
#2356

### DIFF
--- a/CppParser/include/Poco/CppParser/CppParser.h
+++ b/CppParser/include/Poco/CppParser/CppParser.h
@@ -31,7 +31,7 @@
 // CppParser_API functions as being imported from a DLL, whereas this DLL sees symbols
 // defined with this macro as being exported.
 //
-#if (defined(_WIN32) || defined(__CYGWIN__)) && defined(POCO_DLL)
+#if (defined(_MSC_VER) || defined(__CYGWIN__)) && defined(POCO_DLL)
 	#if defined(CppParser_EXPORTS)
 		#define CppParser_API __declspec(dllexport)
 	#else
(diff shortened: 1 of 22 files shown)


CppParser/include/Poco/CppParser/CppParser.h
```diff
@@ -31,7 +31,7 @@
 // CppParser_API functions as being imported from a DLL, whereas this DLL sees symbols
 // defined with this macro as being exported.
 //
-#if (defined(_WIN32) || defined(__CYGWIN__)) && defined(POCO_DLL)
+#if (defined(_MSC_VER) || defined(__CYGWIN__)) && defined(POCO_DLL)
```
aleks-f: @ark0f these should probably be properly checked for MSVC compiler, see #2428 why and how it was fixed in https://github.com/pocoproject/poco/commit/5de2bf7ecffa35b1ecdc07a147942b0c6ff9162a

ark0f: I understood. It will be fixed

aleks-f:
@ark0f this tidy-up is a good effort, but it seems something caused many travis failures, those need to be taken care of before merging this

ark0f:
Thank you. I know about Travis. I see that jobs were interrupted and I can't view full logs as it are more than 4 MB (Travis log limit). I can try to use `git commit --amend` to restart build totally.

ark0f:
Restart didn't give result.

ark0f:
Hm... I see that Clang produce many warnings `unused typedef`. I'll try to fix it.

ark0f:
@aleks-f, I think we need to refactor `POCO_COMPILER_CLANG` definition to avoid problems with using `POCO_COMPILER_GCC` instead of `__GNUC__` as Clang is GCC-like compiler (like MinGW).

PR #2459, opened 2018-09-11
Add some more well known ports

ericcurtin:
smtp, dns, imap and their TLS variants


ericcurtin:
I would find 636 for ldaps useful too. Mind if I add that?

obiltschnig:
`ldaps` would be fine, too.

ericcurtin:
I'll add that and repush this as soon as I get home 😊

obiltschnig:
Thanks!

ericcurtin:
That's added and builds are green @obiltschnig

PR #2480, opened 2018-09-26
TLS mode/version when FIPS is enabled

Bjoe:
It should be possible to set TLS version independently from enable FIPS mode. FIPS (Federal Information Processing Standard) does not prohibit the use of any TLS mode/version.


obiltschnig:

PR #2481, opened 2018-09-26
Add optional Hunter support in cmake buildsystem

Bjoe:
Few months ago, I try to use Poco with conan but without much success. Then I found [Hunter](https://docs.hunter.sh/en/latest/)
Since that I use/build Poco with Hunter and I also build a POC for macchina.io https://github.com/macchina-io/macchina.io/pull/71
Poco was also used in Hunter before (PocoCPP) but they patched "Poco cmake build system." See also here: https://github.com/ruslo/hunter/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+PocoCPP

**Hunter is only optional and as default it is off. So the current cmake build system is not affected.** If you switched it on with the cmake parameter:
```
cmake -Hpoco -Bbuild-poco -DHUNTER_ENABLED=ON
```

First time, it will take a time, because hunter is building the third-party libraries. After third-party are build, it is only building Poco. 
With Hunter you can add Poco easily in your project without installing any new "dependencies". Hunter is based on cmake only.


obiltschnig:

PR #2514, opened 2018-10-23
Fix ctest execution in AppVeyor

Bjoe:
Fix the execution of ctest in AppVeyor


zosrothko:
@Bjoe You are right... The Foundation CI unit tests are failing on Channel::testAsync with a segfaullt. My previous check was invalid. Thus beside merging this PR, one need to check this segfault on Foundation::ChannelTest::testAsync, because it is green with poco-1.9.1

zosrothko:
@aleks-f This PR is ok for me.

Foundation/testsuite/src/ChannelTest.cpp
```diff
@@ -70,13 +70,15 @@ void ChannelTest::testSplitter()
 void ChannelTest::testAsync()
 {
 	AutoPtr<TestChannel> pChannel = new TestChannel;
-	AutoPtr<AsyncChannel> pAsync = new AsyncChannel(pChannel);
+#ifndef POCO_OS_FAMILY_WINDOWS
+	AutoPtr<AsyncChannel> pAsync = new AsyncChannel(pChannel); // FIXME It is crashing here on Windows!
```
aleks-f: @Bjoe can you create an issue for this, with more details; I haven't seen it, may be something appveyor related

zosrothko:
@aleks-f The crash is reproductible on my PC, thus letting AppVeyor out.

Foundation/testsuite/src/ChannelTest.cpp
Bjoe: @aleks-f As zosrothko said, but it is also reproductible on the "Azure" windows.

@zosrothko will you create an issue for that? I think it is a bug in `AsyncChannel`, or?

Bjoe:
@zosrothko Have you seen this PR #2508 ?

Foundation/testsuite/src/ChannelTest.cpp
zosrothko: This unittest is working well with poco-1.9.1.. so the problem should be elsewhere. I will work on it sooner or later

zosrothko:
> 
> 
> @zosrothko Have you seen this PR #2508 ? Maybe fixed the "chrash"?

I don't think so... The crash is happening on this  statement:
`AutoPtr<AsyncChannel> pAsync = new AsyncChannel(pChannel.get());`

Issue #2492, opened 2018-10-08
Net::Socket::address() crash on Android


fenghao119:
#### Actual behavior
crash
#### Steps to reproduce the problem
(please make this a [SSCCE](http://www.sscce.org/), if applicable and reasonable)
```c++
void test()
{
  	using Poco::Net::SocketAddress;
	using Poco::Net::ServerSocket;
	ServerSocket svs(SocketAddress("localhost", 12345));
	std::string s = svs.address().toString(); // crash in release build
}
```

tombstone (call from jni)
```text
Revision: '0'
ABI: 'arm'
pid: 16294, tid: 16294, name: bridge  >>> ./bridge <<<
signal 7 (SIGBUS), code 1 (BUS_ADRALN), fault addr 0xffac4f3e
    r0 f3bb2380  r1 f58c6020  r2 00000000  r3 f3b83510
    r4 ffac4f3e  r5 ffac4f3e  r6 f3e9f1b8  r7 f3bb2380
    r8 00000000  r9 00000000  sl 00000000  fp ffac4ef8
    ip f3bb238c  sp ffac4ef0  lr f5204904  pc f51913b0  cpsr 600f0010
    d0  2073736572646461  d1  ffffffffffffffff
    d2  0000000000000000  d3  0000000000000000
    d4  735a426e44783236  d5  456b6b6278797732
    d6  6c2f3d3d77614a49  d7  6c2f6d72612f6269
    d8  0000000000000000  d9  0000000000000000
    d10 0000000000000000  d11 0000000000000000
    d12 0000000000000000  d13 0000000000000000
    d14 0000000000000000  d15 0000000000000000
    d16 0000003f00000041  d17 63757274736e6f63
    d18 203a73726f746375  d19 0000018000000180
    d20 0000000800000008  d21 0000000800000008
    d22 0000000400000004  d23 0000000400000004
    d24 0000218100000181  d25 0000618100004181
    d26 0000218000000180  d27 0000618000004180
    d28 0000000000000000  d29 0000000000000000
    d30 0000000000000000  d31 0000000000000000
    scr 60000000

backtrace:
    #00 pc 012793b0  /data/app/com.xxx.app/lib/arm/libxxxxx.so (Poco::Net::Impl::IPv4SocketAddressImpl::IPv4SocketAddressImpl(sockaddr_in const*)+36)
[...]

PR #2523, opened 2018-10-29
Fix sockaddr stack aligment

Bjoe:
This pull request will fix issue #2492 

Use `sockaddr_storage` for proper stack aligment.

From the POSIX standard:
> The sockaddr_storage structure solves the problem of declaring storage for automatic variables which is both large enough and aligned enough for storing the socket address data structure of any family.

See also http://pubs.opengroup.org/onlinepubs/009696699/basedefs/sys/socket.h.html


Bjoe:
Ok, let me do this ....

PR #2535, opened 2018-11-08
Fixes #1097

### LINKED ISSUE #1097: Process::isRunning returns true for defunct tagged process
After starting a process with:

``` c++
Pipe out, err;
auto foo = Process::launch("foo", {}, nullptr, &out, &err);
```

The following snippet never exits the loop:

``` c++
while (Process::isRunning(foo)) {
    cout << foo.id() << " still running" << endl;
    sleep_for(1s);
}
```

I've checked with `ps -A | grep foo` that the given foo process has been tagged as `defunct` by the system just after `Process::launch`, meaning that it has been started and exited successfully, but it still stays around as defunct until the parent application terminates:

```
.
.
.
32564 pts/15   00:00:00 zsh
32565 pts/5    00:00:00 foo <defunct>
32680 ?        00:10:33 geary
.
.
.
```

I'm running on x64 Linux with Poco 1.6.1.

josh-stoddard-tanium:
cherry-pick of 0425866486a4652f999d71c359d3645f3aad4ce9

### DIFF
--- a/Foundation/include/Poco/Process_UNIX.h
+++ b/Foundation/include/Poco/Process_UNIX.h
@@ -19,6 +19,9 @@
 
 
 #include "Poco/Foundation.h"
+#include "Poco/Event.h"
+#include "Poco/Mutex.h"
+#include "Poco/Optional.h"
 #include "Poco/RefCountedObject.h"
 #include <unistd.h>
 #include <vector>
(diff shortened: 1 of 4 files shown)


aleks-f:
@josh-stoddard-tanium unfortunately, it won't be as simple as a cherry pick - event has changed in develop (c++11), so there are some things missing, I'm afraid you'll have to dig a bit deeper. 1.x supports C++11 [conditionally](https://github.com/pocoproject/poco/blob/poco-1.9.1/Foundation/include/Poco/Config.h#L27), develop [unconditionally](https://github.com/pocoproject/poco/blob/develop/Foundation/include/Poco/Config.h#L24)

josh-stoddard-tanium:
I've exhausted the effort I'm able to donate to this issue.  Please reopen the issue until it is fixed in a released version of poco.

Issue #2546, opened 2018-11-20
MySQL session state is not cleared in SessionPool


Erbengi:
#### Expected behavior
Session returned by `SessionPool::get()` isn't affected by it's previous usage (only MySQL).

#### Actual behavior
Session returned by `SessionPool::get()` has state from it's previous usage (only MySQL).

#### Steps to reproduce the problem
It is quite difficult to simulate it. Lets say we have 2 sessions in `SessionPool`: *session1* and *session2*.

Get *session1* by call `SessionPool::get()` and use it to insert some data to table:
```sql
INSERT INTO my_table ("test");
```
Get *session2* from session pool and use it to remove row from table:
```sql
DELETE FROM my_table WHERE column="test";
```
Get *session1* from session pool and try to find deleted data:
```sql
SELECT column FROM my_table WHERE column="test"
```
The statement found the row which was deleted with *session2*

#### POCO version
1.9.0

#### Compiler and version
gcc 7.3.0

#### Operating system and version
Ubuntu 18.04, Ubuntu 16.04

#### Other relevant information
This issue can be fixed by using [mysql_reset_connection()](https://dev.mysql.com/doc/refman/5.7/en/mysql-reset-connection.html) when `Session` is returned to the `SessionPool`.

aleks-f:
@Erbengi not that fixing it in develop does not fix it for 1.x series of releases. for those, you should send pull to poco-1.9.1

Erbengi:
I already did, see #2547

PR #2547, opened 2018-11-20
Reset connection when a session is returned to the SessionPool

Erbengi:
Fix of issue #2546

### DIFF
--- a/Data/MySQL/include/Poco/Data/MySQL/SessionHandle.h
+++ b/Data/MySQL/include/Poco/Data/MySQL/SessionHandle.h
@@ -65,7 +65,7 @@ public:
 	void rollback();
 		/// Rollback transaction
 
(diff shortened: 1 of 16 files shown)


Erbengi:
Ok, I've renamed the methods.

Erbengi:
It seems that function `mysql_reset_connection()` was added in MySQL 5.7, but CI machines use 5.6. Probably `mysql_refresh()` should be a solution but I need to test it first.

Erbengi:
`mysql_refresh()` seems working same as `mysql_reset_connection()`

obiltschnig:
Ideally, since `mysql_refresh()` is marked as deprecated, you should check for the MySQL client version and switch between `mysql_refresh()` and `mysql_reset_connection()` using the preprocessor.

PR #2549, opened 2018-11-20
Fix keepAlive in http client session

alesapin:


alesapin:
Is there anybody here?

obiltschnig:
Will look into it after my vacation.

Net/src/HTTPStream.cpp
```diff
@@ -43,7 +43,8 @@ void HTTPStreamBuf::close()
 	if (_mode & std::ios::out)
 	{
 		sync();
-		_session.socket().shutdownSend();
+		if (!_session.getKeepAlive())
```
obiltschnig: This is unnecessary as HTTPOutputStream is not used with persistent connections.

rmbbgh:
Left a comments on the commits: https://github.com/pocoproject/poco/commit/69fee5d2969c540fdf7b7bc6c95d3abf09060f95#commitcomment-32195228
https://github.com/pocoproject/poco/commit/c19ba4106314aee70ed62ab4aca78edffc664237#commitcomment-32210063
The patch needs to be adjusted in my opinion.

hamstergene:
Is this correct though?

1. What if request is big and gets split into two `write()` calls due to buffering: the first write succeeds, the seconds receives I/O error. The retry code will only send the second half of the request.

    Even worse: think about `POST` request and `write()` throwing while sending the body...

2. I have just encountered a proxy that respects the nonstandard `Proxy-Connection: Close` header: Poco thinks keep alive is on (because it's HTTP/1.1 and `Connection:` is absent) and sends second request to the same connection, but the connection is being closed at the same moment, and `receiveResponse()` gets zero bytes and throws "No message received". Regardless of this proxy's nonstandard behaviour, the HTTP standard does allow the peer to close connection at any moment (rfc2616 8.1.4) so...

Is it `HTTPClientSession` responsibility to reconnect and retry? If so, the current code is not enough, because connection loss may be only discovered later on `read()`, and probably is not correct, because only the first `write()` seems safe to resend.

Or should the client code do it's own try/catch and retry both `sendRequest` and `sendResponse` from scratch? In that case this whole `_reconnect = keepAlive` logic seems unnecessary since the client has to handle reconnection anyway.

PR #2584, opened 2019-01-19
Adding standard macOS legacy encodings

csomor:
Created using tec form the Apple vendor files on unicode.org


obiltschnig:
manually merged into poco-1.10.0

PR #2599, opened 2019-01-29
Fix cross-compilation against Poco pre-built

lmayencourt:
This two patches remove hard-coded absolute paths to zlib and pcre when building with `POCO_UNBUNDLED` inside the exported `PocoFoundationTarget.cmake`.
This enables cross-compiling against pre-built binaries of poco provided inside a sysroot. 

- The `ZLIB::ZLIB` imported target set by `FindZLIB.cmake` is used instead of `ZLIB_LIBRARIES`. The installed target will use the `findZLIB cmake` module to define the correct path of the zlib library. 

- Use the` PCRE_NAMES` variable instead of the `PCRE_LIBARIES` provided by `FindPCRE.cmake`. This will add `pcre` to `INTERFACE_LINK_LIBRARIES` and will let the linker find the PCRE library.

The current build will generate: (PocoFoundationTargets.cmake)
```
set_target_properties(Poco::Foundation PROPERTIES
   ...
 INTERFACE_LINK_LIBRARIES "pthread;dl;rt;/usr/lib/aarch64-linux-gnu/libpcre.so;/usr/lib/aarch64-linux-gnu/libz.so"
)
```
Which will fails to link, because the path `/usr/lib/aarch64-linux-gnu/libxxx.so` doesn't exist on the host machine.

and with this modification: (PocoFoundationTargets.cmake)
```
set_target_properties(Poco::Foundation PROPERTIES
  ...
  INTERFACE_LINK_LIBRARIES "pcre;ZLIB::ZLIB;pthread;dl;rt"
)
```
Which will link properly.

This was tested on ubuntu 18.04: natively on x86 and  cross-compiled for aarch64.


filiperinaldi:
Hi @Bjoe, you seem to have some experience on Poco's build system. Do you have some time for a quick look into this PR? Thanks.

Foundation/CMakeLists.txt
```diff
@@ -97,7 +97,10 @@ set_target_properties(Foundation
 	DEFINE_SYMBOL Foundation_EXPORTS
 	)
 
-target_link_libraries(Foundation PUBLIC ${PCRE_LIBRARIES} ${ZLIB_LIBRARIES})
+if (POCO_UNBUNDLED)
+    target_link_libraries(Foundation PUBLIC ${PCRE_NAMES} ZLIB::ZLIB)
```
bachp: Using `PCRE_NAMES` here seems very hacky. It's basically equivalent to just putting `pcre`.

I would prefer to change the `FindPCRE.cmake` module to provide an exported target that can be used.

lmayencourt: Hi @bachp,
Thanks for your comments !

I'm also not really happy with my solution for pcre but I didn't find a simpler solution for this. My first idea was also to modify `FindPCRE.cmake` to provide an exported target, but this not enough:
The `FindPCRE.cmake` is available during the compilation, but it is not part of the installed files. This end up with pre-build looking for an exported target that can not be resolved. This can be solved by installing `FindPCRE.cmake`. The problem is not here with `zlib` because cmake have already a  `FindZLIB.cmake` module.

If installing a module to find PCRE as part of Poco is not an issue, I would be happy to push the required changes.

bachp: I think installing the modified `FindPCRE.cmake` along Poco would be an acceptable solution. In my opinion you can continue.

Bjoe: First, thanks @lmayencourt . I fail to see this issue when I refactor/cleaup all XXX_LIBRARIES places to use the modern art of cmake.
I agree with @bachp , because what with the "include path" for PCRE ? 
Let me cleanup the FindPCRE.cmake and install when POCO_UNBUNDLED is true.
I would like to use your merge request, is this ok for you @lmayencourt?

Finally, I found another place where `ZLIB_LIBRARIES` is used: PDF/CMakeList.txt:129 ....
I can also fix this.

Bjoe:
> Hi @Bjoe, you seem to have some experience on Poco's build system. Do you have some time for a quick look into this PR? Thanks.

Sorry for the late answer....

Bjoe:
@lmayencourt  I create a pull request in your repo with my changes. see https://github.com/lmayencourt/poco/pull/1 Can you test please. Thanks

Bjoe:
@lmayencourt I have added to the cmake build to install FindPCRE. It will be installed in CMAKE_INSTALL_PREFIX/lib/cmake/Poco/FindPCRE.cmake .... I'm not sure if this is a "good place".

Btw. I find that "expat" and "sqlite" has the same issue/problem. I will fix this also if this solution works.

Bjoe:
@lmayencourt Thanks for merging my fixes ....
@bachp Can you please review again?

Again, at the moment I install `FindXYZ.cmake` (for example `FindPCRE.cmake`) on Unix in `CMAKE_INSTALL_PREFIX/lib/cmake/Poco/FindPCRE.cmake` and set the `CMAKE_MODULE_PATH` in `PocoFoundationConfig.cmake` via `list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")` .... I didn't like this and I hope this solution did not create any side effects .... at the moment this is my quickest idea ... if any one have a better solution, please give me a hint or fix this here.

Bjoe:
Btw. we should also merge this to `poco-1.9.1` branch.
I can do this also but please wait before you merge....

Foundation/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ if (POCO_UNBUNDLED)
 		src/pcre_ucd.c
 		src/pcre_tables.c
 	)
-
+
```
bachp: Unneeded whitespace?

PR #2603, opened 2019-02-05
Remove incorrect upper size limits for SSL certificates in NetSSL_Win

roman-kruglov:
In our experience, many valid certificates which we used were larger than the limits set in code (4096 bytes) and thus couldn't be used with our app at all. I believe this restriction (4096) comes from confusing max key size in bits with the file size. In any case I think that calls to the underlying WinAPI should return an error in case the certificate file is defective in some way.

We've been using this "fix" for several months in production, no problems whatsoever. Don't know about the lower size in X509Certificate.cpp (size < 32) though - following the same logic maybe it should be removed too. Anyways we haven't seen a certificate as small as that.

The linux NetSSL seems to work regardless of certificate sizes, so I suppose there is no such confusion there. We haven't experienced any issues with our linux builds and larger certs.

### DIFF
--- a/NetSSL_Win/src/Context.cpp
+++ b/NetSSL_Win/src/Context.cpp
@@ -193,7 +193,6 @@ void Context::importCertificate()
 	Poco::File certFile(_certNameOrPath);
 	if (!certFile.exists()) throw Poco::FileNotFoundException(_certNameOrPath);
 	Poco::File::FileSize size = certFile.getSize();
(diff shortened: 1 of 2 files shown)


roman-kruglov:
Hello guys, any new info on this one? It seems to be a rather simple fix, we tested it in our production for about half a year now. Is there any reason you don't even want to consider this one? Maybe I can help you somehow or change the description?

roman-kruglov:
Wow, that was fast) Thank you! We have several more corrections to this Win SSL module, we'll submit them in the nearest future.

PR #2610, opened 2019-02-12
throw exception when key does not found in a json

nmarandi:
when use a log system for debugging, I think it is better to throw exception when the key does not found in json when use get method instead of return null object. it is easier for debugging purpose.


aleks-f:
perhaps so, but I think it is more important not to break the existing code like this

nmarandi:
it didn't break the code. the unit test expecting the null object and not a exception.

PR #2616, opened 2019-02-15
Restore pre-1.8.0 behaviour of Poco::Net::ServerSocket::bind.

### LINKED ISSUE #2615: Broken backwards compatibility with reuseAddress and reusePort in Poco::Net::ServerSocket::bind in version 1.8.0 or higher.
#### Expected behavior

The behaviour of `Poco::Net::ServerSocket::bind(const SocketAddress& address, bool reuseAddress = false)` should not change between versions.

#### Actual behavior

With the introduction of `reusePort` support in Poco 1.8.0 the behaviour has changed so that `reusePort` is always true when the aforementioned bind delegates to `bind(const SocketAddress& address, bool reuseAddress, bool reusePort)`. Previously the behaviour was to only set `SO_REUSEPORT` if `reuseAddress` was true.

Create and bind two server sockets to the same port with `reuseAddress` set to `false`. Previously the second would fail, whereas now it will pass.

#### POCO version

Broken in Poco >= 1.8.0

#### Compiler and version

Any system supporting `SO_REUSEPORT`.

Any system supporting `SO_REUSEPORT`.

Burgch:
Use `reuseAddress` as the value of `reusePort` for backwards compatibility with older Poco versions. Fixes #2615

### DIFF
--- a/Net/src/SocketImpl.cpp
+++ b/Net/src/SocketImpl.cpp
@@ -205,7 +205,7 @@ void SocketImpl::connectNB(const SocketAddress& address)
 
 void SocketImpl::bind(const SocketAddress& address, bool reuseAddress)
 {
-	bind(address, reuseAddress, true);
+	bind(address, reuseAddress, reuseAddress);
 }
 
 


Burgch:
Is this going to get merged into develop too? I think the Travis failures were infrastructure related?

Issue #2576, opened 2019-01-07
Add std::chrono support


Talkless:
#### Expected behavior

It would be usefull to have support for C+=11 std::chrono types
```
Foo::setTimeout(std::chrono::microseconds timeout) {
//...
    Poco::Timespan ts{timeout};
//...
}

//...

using namespace std::chrono_literals;
Foo f;
f.setTimeout(1s);

```
#### Actual behavior
Poco time-related classes does not accept std::chrono types

#### POCO version
1.9.0

PR #2623, opened 2019-02-21
(#2576) std::chrono support in Timespan

Aleksmia:
Poco::Foundation::Timespan: added std::chrono support in constructor and assign function


Aleksmia:
Solution for issue #2576

Aleksmia:
This solution does not compile in C++11 because std::chrono_literals was released in C++14. I think I should rewrite test.

Issue #2566, opened 2018-12-13
Data: RecordSet & StatementImpl CLOB support in 1.9.1 branch


astlin:
#### Expected behavior

StatementImpl & RecordSet support CLOB datatype

#### Actual behavior

RecordSet partially supports CLOB datatype only when accessing column by index. If accessed by name then "Data type not supported" exception is thrown.

StatementImpl doesn't support CLOB datatype, if there is a text field in selection then "Data type not supported" exception is thrown.

#### POCO version

1.9.1 (probably any others)

#### Compiler and version

any

#### Operating system and version

any

#### Other relevant information

in Data\src\RecordSet.cpp from line 167 there should be added:

`case MetaColumn::FDT_CLOB:      return value<CLOB>(name, row, useFilter);`

in Data\src\StatementImpl.cpp from line 346 there should be added:

```
case MetaColumn::FDT_CLOB:
addInternalExtract<CLOB>(mc); break;
```

omatai:
Something similar is happening again with [Issue 3295](https://github.com/pocoproject/poco/issues/3295)...

aleks-f:
https://github.com/pocoproject/poco/commit/1ba0f8e45bd17b40a4d27ad0efa8941da1e6de6e#diff-52343bc42f646a8a4c6b8ce53bfeecdbbe2d672bedef9670ec48f5674e413de3R191

PR #2677, opened 2019-04-19
Fix CLOB type support in StatementImpl and RecordSet

astlin:
As written in [this](https://github.com/pocoproject/poco/issues/2566#issue-390609008) this commit fixes issue with CLOB datatypes (i.e. text datatype in PostgreSQL)


obiltschnig:

PR #2720, opened 2019-06-12
Add DLL resource for MSVC

OgreTransporter:
In CMake using Visual Studio, adds the DLLVersion.rc file to all DLLs.


obiltschnig:
Does not really work that simple, unfortunately. Had to add:
```
# Version Resource
if(MSVC AND NOT POCO_STATIC)
    source_group("Resources" FILES ${CMAKE_SOURCE_DIR}/DLLVersion.rc)
    list(APPEND SRCS ${CMAKE_SOURCE_DIR}/DLLVersion.rc)
endif()
```
to each library's `CMakeLists.txt` to make it work.

PR #2733, opened 2019-06-23
fix : [ODBC]Binder.h  type deduction for clang-osx

bas524:
On OSX poco-develop has a compilation error into copyContainerValues (Binder.h [ODBC] ), this error depends from the template type and if this type is vector\<bool\>, then the iterator value type is not const reference of bool, and we have a compilation error.


bas524:
@obiltschnig , please review it

bas524:
@obiltschnig Why did you close it? I merged last changes from develop to this brunch, and those last changes broke compilation, not my changes. And without this PR mac os version still broken.

obiltschnig:
was not on purpose, but since we retired the develop branch (and renamed it develop-experimental) GitHub has closed all open PRs. Please re-submit against the new devel (or poco-1.10.0 branch).

PR #2753, opened 2019-07-19
return actual number of sent bytes in sendBytes()

OlegHahm:
According to https://pocoproject.org/docs/Poco.Net.StreamSocket.html#25805 `sendBytes()` should return the number of actual sent bytes. Without this patch `sendBytes()` does always return the number of bytes passed by the caller.

### DIFF
--- a/Net/src/WebSocketImpl.cpp
+++ b/Net/src/WebSocketImpl.cpp
@@ -104,8 +104,7 @@ int WebSocketImpl::sendBytes(const void* buffer, int length, int flags)
 	{
 		std::memcpy(frame.begin() + ostr.charsWritten(), buffer, length);
 	}
-	_pStreamSocketImpl->sendBytes(frame.begin(), length + static_cast<int>(ostr.charsWritten()));
-	return length;
+	return _pStreamSocketImpl->sendBytes(frame.begin(), length + static_cast<int>(ostr.charsWritten()));
 }
 
 


aleks-f:
As it is now, WebSocketImpl::sendBytes returns the payload length, not the total amount of bytes sent; since poco websocket is blocking, StreamSocketImpl::sendBytes will always send everything or throw in the case of error

OlegHahm:
Whaf happens if I call `setSendTimeout(X)` on the web socket? And what happens if I call `setBlocking(false)` on the web socket?

obiltschnig:
The current implementation is consistent with expected behavior as `sendBytes()` returns the length of the payload. Changing this may break existing code. Non-blocking mode is not supported currently for WebSocket. The result of calling setBlocking(false) is undefined. This may change in the future. The effect of `setSendTimeout()` is the same as for a blocking StreamSocket.

obiltschnig:
If anything should be changed at all, it's the documentation for `sendBytes()` or a note in the WebSocket class documentation.

OlegHahm:
I.e. the socket will always send `length` bytes or throw an exception (e.g., if the timeout occurs)?

obiltschnig:
Yes, if successful, a StreamSocket or WebSocket will have sent `length` (or more) bytes, and return `length`. If unsuccessful, an exception will be thrown. In case of a timeout, `Poco::TimeoutException`.

For non-blocking sockets, the actual number of bytes sent (and the return value) may be different.

PR #2762, opened 2019-08-08
Add support for QNX

dukeru:
a. Fix Makefiles so as not to use "dl" library because it's included in C library in QNX SDP7.0
b. Fix File_UNIX.cpp so as it include sys/statvfs.h instead of sys/statfs.h


obiltschnig:
Could you please re-submit the PR agains the poco-1.10.0 branch? We do not allow changes against the master branch. Thank you.

PR #2773, opened 2019-08-22
Fix the issue with incorrect type of store name parameter in CertOpenStore API call into NetSSL_Win

maxim-voronko:
We were unable to load SSL certificates from the Windows certificate store with the OPT_USE_MACHINE_STORE option enabled.
During debugging, we found that the problem was in calling the Windows CertOpenStore API function.
According to the documentation, if you use the CERT_STORE_PROV_SYSTEM parameter, the pvPara parameter must point to a Unicode string that contains the name of the system storage. But in fact, the pvPara parameter pointed to the Ansi string.
Therefore, we changed the std::string certStoreName parameter to std::wstring wcertStore.
We checked this solution. Currently, we can load SSL certificates from the Windows certificate store with the OPT_USE_MACHINE_STORE option enabled.

### DIFF
--- a/NetSSL_Win/src/Context.cpp
+++ b/NetSSL_Win/src/Context.cpp
@@ -167,7 +167,7 @@ void Context::loadCertificate()
 	if (!_hCertStore)
 	{
 		if (_options & OPT_USE_MACHINE_STORE)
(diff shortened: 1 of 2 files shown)


obiltschnig:
Also merged into poco-1.10.0 branch in 0784665ce

PR #2777, opened 2019-08-30
Fix configuration error while cross compiling

fekir:
During the configuration phase in a cross compilation scenario (at least with wine + msvc),
`include(InstallRequiredSystemLibraries)` fails even if
`MSVC_REDIST_DIR` is provided with the error message 
`cmake_host_system_information does not recognize <key> VS_15_DIR`.

This should not be an hard error, in case someone wants to
compile/use the library, and not package it.

As explained on https://reviews.llvm.org/D41220, the most sensible fix
is to include `InstallRequiredSystemLibraries` only on a Windows host.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,7 +284,9 @@ add_custom_target(uninstall
 #############################################################
 # Enable packaging
 
-include(InstallRequiredSystemLibraries)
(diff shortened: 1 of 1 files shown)


fekir:
Up

More than one year passed since I've made the PR, what can I do to have it merged?

aleks-f:
@fekir please redirect this pull to devel branch. master is for relases only

PR #2787, opened 2019-09-19
add buffersize property to the AsyncChannel

killerbot242:
GOAL : adding a bufferSize property to AsyncChannel

RATIONALE:
AsyncChannel is a channel that decouples the log() call from the actual writing. That works with a queue of messages to be written. In the original Poco AsyncChannel, there is no limit to the size of that buffer such that it could grow infinitely if the writing thread is not able to consume the messages as fast as the callers of log(). In our case, one of the callers is a high priority real-time thread. So there is a risk of being flooded.

This pull request introduces a property, bufferSize, on the AsyncChannel. If there are more than that many messages in the buffer, the buffer stops queuing and counts the messages that have been discarded. The next message which can be written will be preceded by the number of messages that have been discarded (sent in the same logging facility and with that logging level).

### DIFF
--- a/Foundation/include/Poco/AsyncChannel.h
+++ b/Foundation/include/Poco/AsyncChannel.h
@@ -50,14 +50,14 @@ public:
 	void setChannel(Channel::Ptr pChannel);
 		/// Connects the AsyncChannel to the given target channel.
 		/// All messages will be forwarded to this channel.
-		
+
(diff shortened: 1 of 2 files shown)


obiltschnig:
Merged into 1.10.1; renamed property "queueSize".

PR #2789, opened 2019-09-20
Log stream buf pre allocation

killerbot242:
GOAL : allowing pre-allocation of the buffer in LogStreamBuf

RATIONALE:
Like any iostream buffer, LogStreamBuf comes with an unknown buffer space pre-allocated. When the logging messages are long, it can induce a number of realloc. Pre-allocating the buffer allows to avoid those expensive reallocations. This is important when the logging happens often (inside a loop that is repeated a lot).


killerbot242:
guys, I made a little mistake, it seems I branched off from my preivous pull request branch, so this one is containing also the 2 modified files of the AsyncChannel buffer size propery pull request.

In here it is only about LogStream.h/cpp

killerbot242:
I already created a new on ; https://github.com/pocoproject/poco/pull/2791

closing this one.

PR #2793, opened 2019-09-23
more efficient logging macros, with boolean shorcut avoiding a messag…

killerbot242:
GOAL : do not create the log message when the logging level is not sufficient enough

RATIONALE:
The original logging macros are constructing the log message even if the logger is disabled. This proposal is using boolean shortcut to eliminate that message construction when the logger is disabled.

Examples : 
poco_stream_debug(stream) << "this is a test" << std::endl;
poco_stream_debug(stream) << createLogMessage() << std::endl;

### DIFF
--- a/Foundation/include/Poco/LogStream.h
+++ b/Foundation/include/Poco/LogStream.h
@@ -27,6 +27,19 @@
 namespace Poco {
 
 
+#define poco_stream(logStream, prio) \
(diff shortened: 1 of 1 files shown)


killerbot242:
little follow up, it seems this is merged into "https://github.com/pocoproject/poco/tree/develop" which is a dead end, and does not seem to exist.
Will be this be used ? And if so what needs to be done to revive it ? Initially this was accepted.

Issue #2803, opened 2019-10-14
[buildsystem] Make Poco git repository fetchable using CMake's FetchContent


brice-gros:
It is currently not possible to get the repository using:
```CMake
FetchContent_Declare(
  poco
  GIT_REPOSITORY "https://github.com/pocoproject/poco.git"
  GIT_TAG        "poco-1.10.0-release"
)
FetchContent_MakeAvailable(poco)
```
Once able to do so, one could only want to get the CppUnit library for reuse in another project, so one should be able to disable the Foundation library as for other components.
(Obviously Foundation should be enabled for building any other components except CppUnit)

PR #2804, opened 2019-10-14
Support the git repository being fetched from CMake

brice-gros:
Resolving #2803


brice-gros:
Retargetted to poco-1.10.0 in #2815

Issue #2809, opened 2019-10-17
Allow to filter long tests using a command line argument


brice-gros:
Currently long running tests are enabled at compile time, it would be nicer to be able to select them at  runtime with a command line argument for instance.
```bash
mytest --all               # run all normal tests except the long tests
mytest --all --long        # run all normal and long tests
mytest MyTestSuite         # run all normal tests in MyTestSuite
mytest --long MyTestSuite  # run all normal and long tests in MyTestSuite
```

PR #2810, opened 2019-10-17
allow long test to be run optionally from cmdline

brice-gros:
Resolving #2809

PR #2814, opened 2019-10-18
allow to select long tests to be run from cmdline

brice-gros:
Resolving #2809


brice-gros:
@obiltschnig Same as PR #2810 about running long test with a command line option with the ENABLE_LONG_RUNNING_TESTS define removed, but against against poco-1.10.0. All tests passed.

PR #2815, opened 2019-10-18
Make Poco git repository fetchable using CMake's FetchContent

brice-gros:
Resolving #2803


CMakeLists.txt
```diff
@@ -175,6 +175,7 @@ else()
 endif()
 
 # Allow enabling and disabling components
+option(ENABLE_FOUNDATION "Enable Foundation, required for all components except CppUnit" ON)
```
brice-gros: Not really required to solve the issue, but can be handy if only the test framework is used (as in one of the project I work on where CppUnit is used for consistency with other related projects in the organisation)

PR #2816, opened 2019-10-18
Modernise TLS configuration

lmamane:
Pushed by a poco-using client that would my web server would not let me connect to, I modernised a bit the default TLS configuration.


obiltschnig:
merged into poco-1.10.1 (made 2048 curves optional, subject to config param).

lmamane:
@obiltschnig, you made using 2048 bit DH _off_ by default. I wish to stress that:
1. From a general security perspective, 1024 bits is considered insufficient nowadays.
2. Implementations are starting to reject connections with only 1024 bits DH. E.g. Debian version 10 ("Buster"), released in July 2019, configures its OpenSSL to reject DH connections with less than 2048 bits by default, by setting OpenSSL's SECLEVEL setting to 2.

For maximum compatibility with latest versions of connection peers, I recommend defaulting to 2048 bit DH.

obiltschnig:
I understand, but I don't want to potentially break existing code relying on the current behavior in a minor update.

lmamane:
Yes, that's a good point.

obiltschnig:
I also had to remove some ECDH groups, as not all OpenSSL builds seem to support them. The set I ended up with for OpenSSL 1.1.1 is "X448:X25519:P-521:P-384:P-256".

PR #2818, opened 2019-10-21
Add getRawPort() method in URI

DimasKovas:
HTTP is often used for inner communication between servers in projects and the standard port in this case may differ from 80. Now there's no way of knowing whether the port 80 was set explicitly or it was set because of it is WellKnownPort.


obiltschnig:
Added to 1.10.1; named method getSpecifiedPort().

PR #2822, opened 2019-10-30
Add possibility to add custom headers through HTTPStreamFactory

### LINKED ISSUE #2820: [Poco::Net::HTTPStreamFactory] Open with custom header parameters
I actually use HTTPStreamFactory to send GET HTTP requests, and I need to add custom header parameters.
I want to use HTTPStreamFactory::open function, because it already manages proxy, credentials and redirections. But today, I didn't find a way to add custom header parameters to the request.
More precisely, the service I use required an API key as header parameter.
If possible, I would like also to modify the "Accept" to restrict to "application/json" (currently set to "*/*").

So what Is the way to provide header parameters ?

Thanks,
sgallou

jano42:
Add new class HPPURI, which overload an URI and provide a way to set http headers.
Enhance HTTP(S)StreamFactory to add http headers provided by HTTPURI
Add unit tests

This PR comes to fix #2820

### DIFF
--- a/Foundation/Foundation_CE_vs90.vcproj
+++ b/Foundation/Foundation_CE_vs90.vcproj
@@ -3691,6 +3691,8 @@
 				<File
 					RelativePath=".\src\URI.cpp"/>
 				<File
+					RelativePath=".\src\HTTPURI.cpp"/>
+				<File
 					RelativePath=".\src\URIStreamFactory.cpp"/>
 				<File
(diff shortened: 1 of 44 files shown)


aleks-f (changes requested):
- only VS90 project should be modified, other VS projects should be generated using ProGen
- HTTPURI class should go into Net library
- redirect pull to devel branch

matejk:
@jano42 , can you please prepare a PR for the branch `main`, please?

PR #2827, opened 2019-11-03
X509Certificate: Get rid of deprecated OpenSSL 1.1 APIs

neheb:
Added a missing implicit header and added compatibility macros.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/Crypto/src/X509Certificate.cpp
+++ b/Crypto/src/X509Certificate.cpp
@@ -27,7 +27,13 @@
 #include <openssl/x509v3.h>
 #include <openssl/err.h>
 #include <openssl/evp.h>
+#include <openssl/bn.h>
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#define ASN1_STRING_get0_data ASN1_STRING_data
+#define X509_get0_notBefore X509_get_notBefore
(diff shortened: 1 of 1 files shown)


obiltschnig:
Please re-target your PR against poco-1.10.0 branch.

neheb:
I think that should do it.

Crypto/src/X509Certificate.cpp
```diff
@@ -255,7 +261,7 @@ void X509Certificate::extractNames(std::string& cmnName, std::set<std::string>&
 			const GENERAL_NAME* name = sk_GENERAL_NAME_value(names, i);
 			if (name->type == GEN_DNS)
 			{
-				const char* data = reinterpret_cast<char*>(ASN1_STRING_data(name->d.ia5));
+				const char* data = reinterpret_cast<char*>(ASN1_STRING_get0_data(name->d.ia5));
```
neheb: My C++ is poor but note that ASN1_STRING_get0_data returns const unsigned char* whereas ASN1_STRING_data returns unsigned char*. There might be a const_cast needed.

obiltschnig: Thanks, will fix.

PR #2830, opened 2019-11-06
Fix wrong buffer size in client handshake when re-using a SecureSocket

moretromain:
In NetSSL_Win SecureSocket impl, receive buffer can be resized after initialization. Therefore when re-using a socket (`reconnect()`), receive buffer shall be reset during client handshake.


obiltschnig:
Thanks, but we do not accept PRs against master. Could you please re-target against poco-1.10.0?

obiltschnig:
merged manually

PR #2837, opened 2019-11-15
Deleting Registry Keys on the WOW6432Node is not possible on 64Bit Applications

### LINKED ISSUE #2010: WinRegistryKey::deleteKey throws NotFoundException
When deleting a native key (not under Wow6432 Node) from a 32-bit application running on a 64-bit OS, NofFoundException is thrown if the target key has a subkey.

```cpp
const std::string key = "HKEY_LOCAL_MACHINE\\SOFTWARE\\SOME\\KEY\\HAS\\SUBKEY";
Poco::Util::WinRegistryKey target(key, false, KEY_WOW64_64KEY );
target.deleteKey() // throws NotFoundException
```

There seems to be a problem in the following part of the deleteKey method of WinRegistryKey.cpp.
```cpp
for (Keys::iterator it = keys.begin(); it != keys.end(); ++it)
{
	std::string subKey(_subKey);
	subKey += "\\";
	subKey += *it;
	WinRegistryKey subRegKey(_hRootKey, subKey); //!!
 	//  It seems that it should be subRegKey(_hRootKey, subKey, false, _extraSam)
	subRegKey.deleteKey();
}
```

KevDi:
On 64 Bit Applications, its not possible to delete an Registry Key located inside the WOW6432Node.
The problem is the constructor call in the deleteKey function. If the Attributes _readOnly and _extraSam are passed to the constructor call the Function works as expected.
Should Resolve #2010

---

Sorry for the reopening i had some issues with git.

### DIFF
--- a/Util/src/WinRegistryKey.cpp
+++ b/Util/src/WinRegistryKey.cpp
@@ -269,7 +269,7 @@ void WinRegistryKey::deleteKey()
 		std::string subKey(_subKey);
 		subKey += "\\";
 		subKey += *it;
-		WinRegistryKey subRegKey(_hRootKey, subKey);
+		WinRegistryKey subRegKey(_hRootKey, subKey, _readOnly, _extraSam);
 		subRegKey.deleteKey();
 	}
 


obiltschnig:
Could you please re-target against poco-1.10.0 branch?

KevDi:
New Pull Request with Target poco-1.10.0

PR #2842, opened 2019-11-18
File fail on overwrite

KevDi:
With the optional Parameter `failOnOverwrite` it's possible to get an exception if the destination File exists on copy/move/rename. In some cases it is usefull to get the Information if the Copy/Move/Rename fail because of an existing File.


obiltschnig:
How about, instead of using a bool parameter, using an options parameter with a FAIL_ON_OVERWRITE flag? Would make the code much more readable. I know we have used bool parameters in way too many places in Poco already, but better late then never to start making it right.

Something like:

```
enum Options
{
    OPT_FAIL_ON_OVERWRITE = 0x01
};

...

void copyTo(const std::string& path, int options = 0) const;
```

KevDi:
I could change it to an Option Parameter like you showed if you want. 
But where should i place the enum? Inside the `File.h`? Or should it be placed in every `File_*` Header?

obiltschnig:
Just put the Options enum inside the File class definition in File.h:

```
class Foundation_API File: private FileImpl
	/// The File class provides methods for working with a file.
	/// ...
{
public:
	typedef FileSizeImpl FileSize;

	enum LinkType
		/// Type of link for linkTo().
	{
		LINK_HARD     = 0, /// hard link
		LINK_SYMBOLIC = 1  /// symbolic link
	};

	enum Options
	{
		OPT_FAIL_ON_OVERWRITE = 0x01
	};

...
```

KevDi:
So i have added the Options Advice. Pls take a look and say if it's okay so =)

Foundation/include/Poco/File_WINCE.h
```diff
@@ -52,8 +52,8 @@ class Foundation_API FileImpl
 	void setSizeImpl(FileSizeImpl size);
 	void setWriteableImpl(bool flag = true);
 	void setExecutableImpl(bool flag = true);
-	void copyToImpl(const std::string& path) const;
-	void renameToImpl(const std::string& path);
+	void copyToImpl(const std::string& path, bool failOnOverwrite = false) const;
```
obiltschnig: Needs to be fixed.

Foundation/src/File_WINCE.cpp
```diff
@@ -277,18 +277,18 @@ void FileImpl::setExecutableImpl(bool flag)
 }
 
 
-void FileImpl::copyToImpl(const std::string& path) const
+void FileImpl::copyToImpl(const std::string& path, bool failOnOverwrite) const
```
obiltschnig: Needs to be fixed.

Foundation/include/Poco/File_WINCE.h
```diff
@@ -52,8 +52,8 @@ class Foundation_API FileImpl
...
+	void renameToImpl(const std::string& path, bool failOnOverwrite = false);
```
obiltschnig: also here

Foundation/src/File_UNIX.cpp
```diff
@@ -340,8 +340,12 @@ void FileImpl::copyToImpl(const std::string& path) const
 		handleLastErrorImpl(_path);
 	}
 	const long blockSize = st.st_blksize;
-
-	int dd = open(path.c_str(), O_CREAT | O_TRUNC | O_WRONLY, st.st_mode);
+	int dd;
+	if (options == 1) {
```
obiltschnig: better check would be `options & File::OPT_FAIL_ON_OVERWRITE` (in case we add another option)

Foundation/src/File_VX.cpp
```diff
@@ -247,7 +247,13 @@ void FileImpl::copyToImpl(const std::string& path) const
 	}
 	const long blockSize = st.st_blksize;
 
-	int dd = open(path.c_str(), O_CREAT | O_TRUNC | O_WRONLY, st.st_mode & S_IRWXU);
+	int dd;
+	if (options == 1) {
```
obiltschnig: options & File::OPT_FAIL_ON_OVERWRITE

Foundation/src/File_UNIX.cpp
```diff
@@ -376,10 +380,15 @@ void FileImpl::copyToImpl(const std::string& path) const
...
 
 
-void FileImpl::renameToImpl(const std::string& path)
+void FileImpl::renameToImpl(const std::string& path, int options)
 {
 	poco_assert (!_path.empty());
 
+	struct stat st;
+
+	if (stat(path.c_str(), &st) == 0 && options == 1)
```
obiltschnig: (options & File::OPT_FAIL_ON_OVERWRITE)

Foundation/src/File_VX.cpp
```diff
@@ -276,10 +282,15 @@ void FileImpl::copyToImpl(const std::string& path) const
...
 
 
-void FileImpl::renameToImpl(const std::string& path)
+void FileImpl::renameToImpl(const std::string& path, int options)
 {
 	poco_assert (!_path.empty());
 
+	struct stat st;
+
+	if (stat(path.c_str(), &st) == 0 && options == 1)
```
obiltschnig: (options & File::OPT_FAIL_ON_OVERWRITE)

Foundation/src/File_WIN32.cpp
```diff
@@ -288,21 +288,27 @@ void FileImpl::setExecutableImpl(bool flag)
 }
 
 
-void FileImpl::copyToImpl(const std::string& path) const
+void FileImpl::copyToImpl(const std::string& path, int options) const
 {
 	poco_assert (!_path.empty());
 
-	if (CopyFileA(_path.c_str(), path.c_str(), FALSE) == 0)
+	if (CopyFileA(_path.c_str(), path.c_str(), options == 1) == 0)
```
obiltschnig: (options & File::OPT_FAIL_ON_OVERWRITE) != 0

Foundation/src/File_WIN32.cpp
```diff
@@ -288,21 +288,27 @@ void FileImpl::setExecutableImpl(bool flag)
...
 
 
-void FileImpl::renameToImpl(const std::string& path)
+void FileImpl::renameToImpl(const std::string& path, int options)
 {
 	poco_assert (!_path.empty());
 
-	if (MoveFileExA(_path.c_str(), path.c_str(), MOVEFILE_REPLACE_EXISTING) == 0)
-		handleLastErrorImpl(_path);
+	if (options == 1) {
```
obiltschnig: options & File::OPT_FAIL_ON_OVERWRITE

Foundation/src/File_WIN32U.cpp
```diff
@@ -291,25 +291,30 @@ void FileImpl::setExecutableImpl(bool flag)
...
 
-void FileImpl::copyToImpl(const std::string& path) const
+void FileImpl::copyToImpl(const std::string& path, int options) const
 {
 	poco_assert (!_path.empty());
 
 	std::wstring upath;
 	convertPath(path, upath);
-	if (CopyFileW(_upath.c_str(), upath.c_str(), FALSE) == 0)
+	if (CopyFileW(_upath.c_str(), upath.c_str(), options == 1) == 0)
```
obiltschnig: (options & File::OPT_FAIL_ON_OVERWRITE) != 0

Foundation/src/File_WIN32U.cpp
```diff
@@ -291,25 +291,30 @@ void FileImpl::setExecutableImpl(bool flag)
...
-void FileImpl::renameToImpl(const std::string& path)
+void FileImpl::renameToImpl(const std::string& path, int options)
 {
 	poco_assert (!_path.empty());
 
 	std::wstring upath;
 	convertPath(path, upath);
-	if (MoveFileExW(_upath.c_str(), upath.c_str(), MOVEFILE_REPLACE_EXISTING) == 0)
-		handleLastErrorImpl(_path);
+	if (options == 1) {
```
obiltschnig: options & File::OPT_FAIL_ON_OVERWRITE

Foundation/src/File_UNIX.cpp
KevDi: So i have to include the `File.h` Header inside the `File_UNIX.cpp`?

KevDi:
So i have to include the `File.h` Header inside the `File_UNIX.cpp`?
Which seems like a circle include because `File_UNIX` already included inside the `File.h`

obiltschnig:
Yeah, good point. In the other places we have actually defined the enum in the Impl classes (like `FileSize`). I see that for `LinkType`, this has also been implemented the "wrong" way, using a magic number in `FileImpl`, rather than the symbolic value. I'd say, leave it as is for now, I will eventually take care of both. And thank you for your contribution!

obiltschnig:
Btw, an example for how it's done properly is the Thread class.

KevDi:
I have added the remarks like it was implemented inside the Thread Class. Currently i done it for the Unix and Win32U File Implementation. Please take a look if it is ok so. If its ok i will made the changes for the missing ones.

obiltschnig:
Looks good, but please also fix the checks for `options == 1`.

KevDi:
Yes i will fix them i just want to ask if its okay the way i did it now

KevDi:
I have made the changes to the other File Impls

PR #2848, opened 2019-11-27
Added MinGW support & fix ApacheConnector errors

YuriAzathoth:
- Added MinGW support
- Fixed ApacheConnector errors


YuriAzathoth:
More clean pull request based on branch devel will be provided

PR #2855, opened 2019-12-04
setFailureActions Crashes

KevDi:
Currently there are 2 Situations in which the `setFailureActions` could crash.

1. If the `FailureActionVector` has less then 3 Elements. 
2. If there is no `SVC_REBOOT` and `SVC_RUN_COMMAND` is specified. In that case the `ac.lpCommand` and `ac.lpRebootMsg` Parameters are not set.

This commit should fix boths of the Situations. 
The Function now throws an `InvalidArgument` Exception if the Size of the Vector is larger then 3 Elements


KevDi:
Will recreate that pull request based on the poco-1.10.1 branch

PR #2866, opened 2019-12-16
unescape Backslash char in UTF8 unescape method

vfjpl:
This PR fixes UTF8::unescape() method. This method was lacking a unecaping for \ character. I added it and also changed big if else ladder to switch case. testsuite is also updated to test backslash unescaping.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -50,3 +50,4 @@ Jeff Adams
 Martin Osborne
 Björn Schramke
 Francis Andre
+Kacper Piwiński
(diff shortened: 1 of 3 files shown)


obiltschnig:
Please re-target against poco-1.10.0 branch. Thank you.

vfjpl:
@obiltschnig done in #2867 
Also I think It would be a good idea to state clearly in CONTRIBUTING.md that contributors should not create pull requests against master :)

obiltschnig:
merged #2867

PR #2871, opened 2019-12-18
Add support for utf-8 characters in subject

### LINKED ISSUE #2869: X509Certificate does not render UTF-8 characters in subjectName
The current implementation of `X509Certificate::subjectName()` doesn't return return non ASCII UTF-8 characters. Instead it returns them in hex escaped form.

If the subject conains Zählerweg as the locality the expected return would be:
```
/L=Zählerweg
```
However `X509Certificate::subjectName()` returns:

```
/L=Z\xC3\xA4hlerweg
```

Using `subjectName(X509Certificate::NID_LOCALITY_NAME)` properly returns `Zählerweg`

So i think the issue lies in the OpenSSL function [`X509_NAME_oneline`](https://www.openssl.org/docs/man1.0.2/man3/X509_NAME_oneline.html), used [here](https://github.com/pocoproject/poco/blob/master/Crypto/src/X509Certificate.cpp#L202) which according to the documentation is "strongly discouraged in new applications".

## Test case to reproduce issue:
https://github.com/siemens/poco/commit/eb35254610b4c793a13983c5f8482737414d4591

Poco Version:  1.9.4 and master

bachp:
This changes the format produced by OpenSSL from being separated by / to ,

@obiltschnig @aleks-f Is the change in formatting acceptable?

Closes #2869

### DIFF
--- a/Crypto/src/X509Certificate.cpp
+++ b/Crypto/src/X509Certificate.cpp
@@ -38,7 +38,6 @@
 namespace Poco {
 namespace Crypto {
 
-
 X509Certificate::X509Certificate(std::istream& istr):
 	_pCert(0)
 {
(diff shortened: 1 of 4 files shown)


obiltschnig:
Should be okay, despite the format change. Could you please re-submit against poco-1.10.0 branch?

bachp:
@obiltschnig Should I create a second MR against 1.10 or change this one?

obiltschnig:
Whatever is easier for you ;-)

bachp:
@obiltschnig Target branch changed and rebased.

PR #2881, opened 2020-01-13
Add an option to force the use of PollingDirectoryWatcherStrategy

Ludorg:
On Linux, inotfy does not work for network mounted volumes (such as NFS). See https://stackoverflow.com/questions/4231243/inotify-with-nfs

By adding flag POCO_DW_FORCE_POLLING in Foundation/Config.h, the use of PollingDirectoryWatcherStrategy is forced on Linux. This is not the same behavior as flag POCO_NO_INOTIFY which disables compilation of DirectoryWatcher.

### DIFF
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -107,7 +107,7 @@
 //#define POCO_NO_INOTIFY
 
 // Define to force the use of PollingDirectoryWatcher 
-#define POCO_DW_FORCE_POLLING
+// #define POCO_DW_FORCE_POLLING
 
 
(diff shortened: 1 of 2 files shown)


aleks-f:
This behavior (probably) applies to any shared FS, and is (probably) not Linux-specific on top of that. If my guesses are correct, flag should apply to any platform.

Ludorg:
> 
> 
> This behavior (probably) applies to any shared FS, and is (probably) not Linux-specific on top of that. If my guesses are correct, flag should apply to any platform.

OK, I will refactor this and propose a new pull request with the propose behavior.

PR #2883, opened 2020-01-14
Add missing relational operators to VarIterator

vfjpl:
Was not sure if implementation should be put in .cpp or in .h with inline. So I did the same thing as the two operators that was already there.


vfjpl:
Any feedback?

PR #2902, opened 2020-02-02
[FTPClient] Optimize expression

gaurav1086:


obiltschnig:
merged into poco-1.10.1

Issue #2894, opened 2020-01-28
Poco 1.10.0 doesn't build with cmake & POCO_UNBUNDLED


swt2c:
```
/usr/bin/cmake -DCMAKE_C_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_Fortran_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DLIB_SUFFIX=64 -DBUILD_SHARED_LIBS:BOOL=ON -DPOCO_UNBUNDLED=ON -DENABLE_TESTS=ON -DENABLE_REDIS=OFF -DODBC_INCLUDE_DIRECTORIES=/usr/include/libiodbc ..
-- The C compiler identification is GNU 10.0.1
-- The CXX compiler identification is GNU 10.0.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Checking for C++14 compiler
-- Checking for C++14 compiler - available
-- Found OpenSSL: /usr/lib64/libcrypto.so (found version "1.1.1d")  
-- Could NOT find APR (missing: APR_INCLUDE_DIR APR_LIBRARY) 
-- Could NOT find APRUTIL (missing: APRUTIL_INCLUDE_DIR APRUTIL_LIBRARY) 
-- Could NOT find Apache2 (missing: APACHE2_INCLUDE_DIR) 
-- Found MYSQL: /usr/include/mysql  
[...]

PR #2910, opened 2020-02-07
Fix Poco static build on Windows

Bjoe:
Fix following LINK error on Windows: fatal error LNK1104: cannot open file 'PocoFoundationmd.lib'
See https://ci.appveyor.com/project/Bjoe/hunter-t1t06/builds/30611903/job/q9tcn7i4q5xtan4f#L6473

### DIFF
--- a/cmake/DefinePlatformSpecifc.cmake
+++ b/cmake/DefinePlatformSpecifc.cmake
@@ -53,16 +53,16 @@ endif(MSVC)
 
 
 # Add a d postfix to the debug libraries
-if(POCO_STATIC)
-        set(CMAKE_DEBUG_POSTFIX "${STATIC_POSTFIX}d" CACHE STRING "Set Debug library postfix" FORCE)
-        set(CMAKE_RELEASE_POSTFIX "${STATIC_POSTFIX}" CACHE STRING "Set Release library postfix" FORCE)
-        set(CMAKE_MINSIZEREL_POSTFIX "${STATIC_POSTFIX}" CACHE STRING "Set MinSizeRel library postfix" FORCE)
-        set(CMAKE_RELWITHDEBINFO_POSTFIX "${STATIC_POSTFIX}" CACHE STRING "Set RelWithDebInfo library postfix" FORCE)
-else(POCO_STATIC)
-        set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Set Debug library postfix" FORCE)
-        set(CMAKE_RELEASE_POSTFIX "" CACHE STRING "Set Release library postfix" FORCE)
-        set(CMAKE_MINSIZEREL_POSTFIX "" CACHE STRING "Set MinSizeRel library postfix" FORCE)
-        set(CMAKE_RELWITHDEBINFO_POSTFIX "" CACHE STRING "Set RelWithDebInfo library postfix" FORCE)
+if(BUILD_SHARED_LIBS)
+    set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Set Debug library postfix" FORCE)
+    set(CMAKE_RELEASE_POSTFIX "" CACHE STRING "Set Release library postfix" FORCE)
+    set(CMAKE_MINSIZEREL_POSTFIX "" CACHE STRING "Set MinSizeRel library postfix" FORCE)
+    set(CMAKE_RELWITHDEBINFO_POSTFIX "" CACHE STRING "Set RelWithDebInfo library postfix" FORCE)
+else()
+    set(CMAKE_DEBUG_POSTFIX "${STATIC_POSTFIX}d" CACHE STRING "Set Debug library postfix" FORCE)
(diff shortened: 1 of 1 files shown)


Bjoe:
It works here https://ci.appveyor.com/project/Bjoe/hunter-t1t06/builds/30625454/job/tdv3l0q244p04aug but it would be create if some one with a Windows machine can verify on Windows if all are fine ...

bachp:
@Bjoe Sorry I don't have access to a windows machine at the moment.

But I'm not sure if the change is correct. If I remember correctly [`POCO_STATIC`](https://github.com/pocoproject/poco/search?l=CMake&q=POCO_STATIC) is currently the way to tell poco to compile statically.

But I agree the standard CMake variable would be better but it probably requires some more changes.

obiltschnig:
I also don't think the fix is correct. Right now we have the unfortunate situation that we have both POCO_STATIC and BUILD_SHARED_LIBS, and if both settings are on conflicting settings the build fails.

obiltschnig:
Hopefully now properly fixed with 75a86ad868027f2c07b1934960a743d644328710 and e5afb2fa89498235e17d15c3863f5bda12aba9e0

Bjoe:
@bachp Yep, `POCO_STATIC` is one way to tell poco to compile statically. `POCO_STATIC` will [set  `BUILD_SHARED_LIBS`](https://github.com/pocoproject/poco/blob/0557e871172949841a374803ecca87c5682e897a/CMakeLists.txt#L59) on a correct value to specify the type of library to be created via `add_library()`. If on `add_library()` no type is given explicitly the type is `STATIC` or `SHARED` based on whether the current value of the variable `BUILD_SHARED_LIBS` is `ON`. See https://cmake.org/cmake/help/latest/command/add_library.html?highlight=add_library

But it is also possible to set via cmake command for example:
```
cmake -DBUILD_SHARED_LIBS=ON ....
```
to tell poco to compile shared libs. Unfortunately this was not working on Windows because the poco build was only verify the `POCO_STATIC` property not the `BUILD_SHARED_LIBS` property ....

But now ... [`POCO_STATIC` is removed](https://github.com/pocoproject/poco/commit/75a86ad868027f2c07b1934960a743d644328710) ... this was not my idea :-)

@obiltschnig I think we break now some "user builds" :-) when we remove `POCO_STATIC`

Bjoe:
@obiltschnig I currently testing your change ... https://ci.appveyor.com/project/Bjoe/hunter-t1t06 and https://travis-ci.org/Bjoe/hunter/builds/650065693 ....

obiltschnig:
The main issue with having `POCO_STATIC` and `BUILD_SHARED_LIBS` is that we'd have two settings to control the same feature that could potentially be set to conflicting values. As happend with the Windows build (POCO_STATIC set to ON, while BUILD_SHARED_LIBS was also still ON). I found no easy way to fix this (which setting should take precedence?). If BUILD_SHARED_LIBS is the standard CMake way to control this we should use it this way, at the cost of potentially breaking existing builds.

obiltschnig:
In any case I'd like this feature closed before Monday, as I don't have more time to work on the 1.10.1 release.

obiltschnig:
OK, here's some attempt at backwards compatibility:
13d0bd7f0b56bc91d478a60145a044f97c022ecb

With a warning:
6aa1abd8f6db8d6fc5f877d900faec4e6191b7ff

bachp:
I also prefer if we go with BUILD_SHARED_LIBS.

Bjoe:
> The main issue with having `POCO_STATIC` and `BUILD_SHARED_LIBS` is that we'd have two settings to control the same feature that could potentially be set to conflicting values. As happend with the Windows build (POCO_STATIC set to ON, while BUILD_SHARED_LIBS was also still ON). 

Hm good point....

> I found no easy way to fix this (which setting should take precedence?)

I think I have an idea .... let me try

> If BUILD_SHARED_LIBS is the standard CMake way to control this we should use it this way, at the cost of potentially breaking existing builds.

:+1: 

Yep, that's why I changed this to `BUILD_SHARED_LIBS` one year ago ....


At the moment, there is now an other problem .... [Could not find a package configuration file provided by "PocoFoundation"](https://ci.appveyor.com/project/Bjoe/hunter-t1t06/build/job/b45s6qpxvj7paf9h#L6361) .... let me see if I can fix this quickly ...

obiltschnig:
Would be great to have it fixed by Monday, so as not to further delay the release.

Bjoe:
> Would be great to have it fixed by Monday, so as not to further delay the release.

@obiltschnig Yep I know ... I will do my best ... I think I have an idea what are the problem ... depends on change #2894 ...

Bjoe:
@obiltschnig My [comment on)(https://github.com/pocoproject/poco/issues/2894#issuecomment-583516967) #2894 pointed to the problem. I fixed in PR #2925

Bjoe:
@obiltschnig Ok you already merged #2925 :-) ... I close this PR. It is not needed anymore.

Issue #2776, opened 2019-08-30
Shutdown TLS1.3 connection


micheleselea:
I think we have to deal with SecureSocketImple::shutdown() and TLS1.3 (using for example OpenSSL 1.1.1c).
I found some strange behavior that broke connection to the end transfer if the client do not correctly close connection. [Reading this](https://github.com/openssl/openssl/issues/6904) I understand that we should do something more during shutdown or we can disable the session tickets sent after handshake like

```
#if OPENSSL_VERSION_NUMBER >= 0x1010100fL
        /* TLS 1.3 server sends session tickets after a handhake as part of
         * the SSL_accept(). If a client finishes all its job before server
         * sends the tickets, SSL_accept() fails with EPIPE errno. Since we
         * are not interested in a session resumption, we can not to send the
         * tickets. */
        /*if (1 != SSL_set_num_tickets(ssl, 0)) {
            fprintf(stderr, "SSL_set_num_tickets failed\n");
            exit(EXIT_FAILURE);
        }
        Or we can perform two-way shutdown. Client must call SSL_read() before
        the final SSL_shutdown(). */
#endif
```

I'm going to have some try, because I found some strange behavior for example in FTPS connection, after client sent all data and close connection the socket receive on server side throw exception.
If I find more I'll share with you

micheleselea:
[other reference](https://github.com/openssl/openssl/issues/6262 )
if I use the SSL_set_num_tickets(ssl, 0) in acceptSSL after SSL_new I think the problem is solved but we lost session resumption.
I'll try even with the client side walkaround of the shutdown, should be something like

```
 retval = SSL_shutdown(ssl);
    if (retval < 0) {
        int ssl_err = SSL_get_error(ssl, retval);
        fprintf(stderr, "Client SSL_shutdown failed: ssl_err=%d\n", ssl_err);
        kill(server_pid, SIGTERM);
        exit(EXIT_FAILURE);
    }
    printf("Client shut down TLS session.\n");

    if (retval != 1) {
        /* Consume all server's data to access the server's shutdown */
        while (1) {
            processed = SSL_read(ssl, buffer, sizeof(buffer));
            printf("Client SSL_read returned %d\n", processed);
            if (processed <= 0) {
                int ssl_error = SSL_get_error(ssl, processed);
                if (ssl_error == SSL_ERROR_ZERO_RETURN) {
                    printf("Client thinks a server finished sending data\n");
                    break;
                }
[...]

PR #2918, opened 2020-02-12
Disable TLS1.3 session resumption

micheleselea:
In TLS 1.3 session resumption is enabled by default: we can disable to keep shutdown working as is, otherwise we should change the way shutdown close connection.
related issue: [#2776](https://github.com/pocoproject/poco/issues/2776)


aleks-f:
641d29f4a

PR #2926, opened 2020-02-17
Fix clang issue

linquize:
Fix clang issue


obiltschnig:
Please state clearly what your PR is about and send a clean PR against `devel` branch, not one against master that changes 185 files.

PR #2942, opened 2020-02-26
Avoid clang 10 -Wimplicit-int-float-conversion warning/error when con…

Romain-Geissler-1A:
…verting int into float

Example of warning (error when using -Werror) we get with clang 10:

/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:444:14: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-int-float-conversion]
                if (from > std::numeric_limits<T>::max())
                         ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:332:4: note: in instantiation of function template specialization 'Poco::Dynamic::VarHolder::checkUpperLimitFloat<float, int>' requested here
                        checkUpperLimitFloat<F,T>(from);
                        ^
/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:2175:3: note: in instantiation of function template specialization 'Poco::Dynamic::VarHolder::convertToSmaller<float, int>' requested here
                convertToSmaller(_val, val);
                ^

PR #2943, opened 2020-02-26
Avoid clang 10 -Wimplicit-int-float-conversion warning/error when con…

Romain-Geissler-1A:
…verting int into float

Example of warning (error when using -Werror) we get with clang 10:

/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:444:14: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-int-float-conversion]
                if (from > std::numeric_limits<T>::max())
                         ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:332:4: note: in instantiation of function template specialization 'Poco::Dynamic::VarHolder::checkUpperLimitFloat<float, int>' requested here
                        checkUpperLimitFloat<F,T>(from);
                        ^
/remote/intdeliv/components/osp/Poco/Foundation/19-0-0-6/include/Poco/Dynamic/VarHolder.h:2175:3: note: in instantiation of function template specialization 'Poco::Dynamic::VarHolder::convertToSmaller<float, int>' requested here
                convertToSmaller(_val, val);
                ^

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -429,15 +429,33 @@ private:
 	template <typename F, typename T>
 	void checkUpperLimitFloat(const F& from) const
 	{
-		if (from > std::numeric_limits<T>::max())
-			throw RangeException("Value too large.");
+		if (std::is_floating_point<T>::value)
+		{
+			if (from > std::numeric_limits<T>::max())
+				throw RangeException("Value too large.");
+		}
+		else
+		{
+			// Avoid clang -Wimplicit-int-float-conversion warning with an explicit cast.
+			if (from > static_cast<F>(std::numeric_limits<T>::max()))
+				throw RangeException("Value too large.");
+		}
 	}
 
 	template <typename F, typename T>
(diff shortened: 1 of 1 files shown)


Romain-Geissler-1A:
> I don't have time to check right now, but reading it seems to me that this would not fix the problem because offending code would still be compiled - you'd need enable_if to fix it properly.

Well we did patch our own poco like this, and clang warning is now silent. Clang does it only when it's able to do constant propagation, so it will take the side of the "if" and thus not emit the warning. So if the warning is gone like this, is this ok ?

chrisse74:
This PR (or more precisely its predecessor #2942 was mentioned in https://github.com/pocoproject/poco/blob/poco-1.11.0/CHANGELOG but is not merged yet. Maybe someone can have a look... :)

aleks-f:
it is marked for the next [release](https://github.com/pocoproject/poco/milestone/32)

PR #2959, opened 2020-03-18
Fix percent-encoded fragment modification in Poco::URI

dahek:
Before this commit using Poco::URI class to parse specific URIs that had
percent-encoded fragment identifier resulted in the loss of information
concerning the way the fragment identifier was encoded. There could be the cases
when the  result of Poco::URI object serialization to string did not match the
original URI string Poco::URI object was created from.

In this commit we change the internal logic of fragment processing in Poco::URI,
so that the fragment is stored inside the class in raw form (the same way as
query string). The methods getFragment and setFragment work the old way (with
percent-decoded fragment values), new methods getRawFragment and setRawFragment
are added to get access to the original fragment representation.

### DIFF
--- a/Foundation/include/Poco/URI.h
+++ b/Foundation/include/Poco/URI.h
@@ -40,15 +40,18 @@ class Foundation_API URI
 	/// The class automatically performs a few normalizations on
 	/// all URIs and URI parts passed to it:
 	///   * scheme identifiers are converted to lower case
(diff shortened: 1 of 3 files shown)


aleks-f:
@obiltschnig this is a breaking change - `getFragment()` would now returns `const std::string&`, so that would now be a reference to a temporary object for code that was retrieving it as reference

PR #2983, opened 2020-04-19
Handle unexpected EOF when reading HTTP streams

oandrew:
Throw an exception on EOF before stream is complete.
Stream is complete if:
Chunked-Encoding: zero-length chunk received
Content-Length:  N bytes received

### DIFF
--- a/Net/include/Poco/Net/HTTPChunkedStream.h
+++ b/Net/include/Poco/Net/HTTPChunkedStream.h
@@ -48,6 +48,9 @@ protected:
 	int readFromDevice(char* buffer, std::streamsize length);
 	int writeToDevice(const char* buffer, std::streamsize length);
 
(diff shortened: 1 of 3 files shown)


Mironenko:
Hi, @oandrew! Could you please provide the tests so it were easier to understand what bug you solve? There is the other PR concerning the same code: https://github.com/pocoproject/poco/pull/2998 Do you solve the same issue?

Mironenko:
Throwing exception from the StreamBufs looks like a rather bold decision. Are you quite sure this does not break the requirement to leave the stream buffer in valid state (see http://www.cplusplus.com/reference/streambuf/basic_streambuf/underflow/ - Exception safety)? What's going to happen if IO stack above will call readFromDevice once more?

PR #2990, opened 2020-04-28
Fix invalid request parsing.

anton-barabanov:
Problem is in case, when buffer stops reading in the last symbol of the
line. Next time it will read '\r\n' sequence and interpret it as an
empty line and set _end to true.


aleks-f (changes requested):
please write a test for this

obiltschnig:
I don't understand what issue this is solving.

aleks-f:
I guess what @anton-barabanov is saying is that, when `readFromDevice` reads `\r\n` only, it will interpret it as the end prematurely. But I'm not sure this fix is good and how to produce a test reproducing it.

PR #2997, opened 2020-05-14
Ignore Chunked-Body trailer

dahek:
Correct processing of trailer that may exist in Chunked-Body (see 3.6.1
RFC 2616) was added.


obiltschnig:
Properly fixed in #3658

PR #2998, opened 2020-05-14
Fix HTTPChunkedStreamBuf trying to read after eof

dahek:
Before this commit reading from HTTPChunkedStreamBuf could freeze if
HTTPChunkedStreamBuf::readFromDevice was called once more after
receiving EOF. This can happens if reading from HTTPChunkedStreamBuf is
performed using streambuf::sgetn method available through
istream::rdbuf.


Mironenko:
@obiltschnig , is there any chance to get that one merged? It has hugely hit us when request/response stream was accessed from some filtering_stream first and directly after (eg to handle !2977). HTTPChunkedStreamBuf's got to keep a track of its state, as HTTPFixedLengthStreamBuf does. What's to be done to get it merged?

PR #3013, opened 2020-05-27
MongoDB::Array interface improvements

matejk:
Two improvements:

* Use std::size_t instead of int in get function to be consistent with type returned from size()
* Array-specific add functions that automatically set stringified sequence number for element name to simplify adding members to array
* added test function for MongoDB array


obiltschnig:
Thanks, could you please re-target the PR against the `devel` branch?

matejk:
OK, I will create new PR.

Issue #3012, opened 2020-05-26
Application::uninitialize has regression 


gatkinso:
I note that the behavior of Application::uninitialize was changed in commit 1bf40a0cd28be83e6243a0524be19118792c8cd1

Documentation states:

> Uninitializes the application and all registered subsystems. Subsystems are always uninitialized in reverse order in which they have been initialized.
> 
> Overriding implementations must call the base class implementation.

However, this is no longer the case.  Subsystems are now uninitialized **in the order** of which they were registered.

PR #3014, opened 2020-05-27
Revert Application::uninitialize to old behavior.

gatkinso:
Fix for issue:  https://github.com/pocoproject/poco/issues/3012

****

I note that the behavior of Application::uninitialize was changed in commit 1bf40a0

Documentation states:

    Uninitializes the application and all registered subsystems. Subsystems are always uninitialized in reverse order in which they have been initialized.

    Overriding implementations must call the base class implementation.

However, this is no longer the case. Subsystems are now uninitialized in the order of which they were registered.

### DIFF
--- a/Util/src/Application.cpp
+++ b/Util/src/Application.cpp
@@ -191,10 +191,10 @@ void Application::uninitialize()
 {
 	if (_initialized)
 	{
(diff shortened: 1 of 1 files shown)


obiltschnig:
see #3266

PR #3016, opened 2020-05-30
MongoDB::Array interface improvements (on top of 1.11.2)

matejk:
Two improvements:

* Use std::size_t instead of int in get function to be consistent with type returned from size()
* Array-specific add functions that automatically set stringified sequence number for element name to simplify adding members to array
* added test function for MongoDB array

### DIFF
--- a/MongoDB/include/Poco/MongoDB/Array.h
+++ b/MongoDB/include/Poco/MongoDB/Array.h
@@ -39,8 +39,31 @@ public:
 	virtual ~Array();
 		/// Destroys the Array.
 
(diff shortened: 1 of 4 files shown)


matejk:
@obiltschnig: This PR replaces #3013.

matejk:
Rebased to 1.11.2.

aleks-f:
@matejk sorry for being so unresponsive, this should go into 1.12.0 release

matejk:
@aleks-f, no problem. I know that this is not the most important PR.

Issue #2864, opened 2019-12-12
SessionImpl begin


petko:
I want to start a SQLite transaction in `Session` using `BEGIN EXCLUSIVE`. However in `SessionImpl` `begin` is hard-coded to use `BEGIN DEFERRED` and there is currently no way to change that.

Would you accept a pull request with a function, which allows to specify the transaction mode used by the `Session` object? If so, what would be the proposed API for doing it in order it to be consistent with the rest of the POCO library?

aleks-f:
It would be best done using a session property. for an example, see how `connectionTimeout` property is [done](https://github.com/pocoproject/poco/blob/poco-1.10.1/Data/SQLite/src/SessionImpl.cpp#L60)

PR #3018, opened 2020-06-01
Added support for SQLite transaction types using Session property

kitaisreal:
Added support for SQLite transaction types using Session property
@aleks-f  https://github.com/pocoproject/poco/issues/2864

1. Added property key and transaction type enum in SQLite.h
2. Added property details in SessionImpl.cpp
3. Updated begin transaction in SessionImpl.cpp with property type
4. Updated tests

### DIFF
--- a/Data/SQLite/include/Poco/Data/SQLite/SQLite.h
+++ b/Data/SQLite/include/Poco/Data/SQLite/SQLite.h
@@ -58,5 +58,22 @@
 	#endif
 #endif
 
+//
+// SQLite Transaction type
+//
+
+namespace Poco {
+namespace Data {
+namespace SQLite {
+
+	extern const std::string TRANSACTION_TYPE_PROPERTY_KEY;
+
+	enum class TransactionType {
+		deferred,
+		exclusive,
+		immediate
+	};
+
+} } } // namespace Poco::Data::SQLite
 
 #endif // SQLite_SQLite_INCLUDED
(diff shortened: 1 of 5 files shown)


aleks-f:
Thanks for the contribution; however, few things here:

1. pulls should go against `devel` branch, not `master`
2. SQLite transaction support should fit into the [Data library session/transaction framework](https://github.com/pocoproject/poco/blob/master/Data/include/Poco/Data/Session.h#L157)

Sorry that I was not more precise when you inquired first. You can see eg. [ODBC transaction tests](https://github.com/pocoproject/poco/blob/master/Data/ODBC/testsuite/src/SQLExecutor.cpp#L3516-L3906) to get a better idea. Back-end DB engines can differ quite a bit, but we always try to factor out the common denominator but also leave opened the opportunity for any back-end-specific features to be passed through property or feature.

kitaisreal:
Hi, @aleks-f, thanks for review.
1. Updated pull request. But it seems like CI could not start build, could you trigger it manually or there is some problem with branches ?
2. This transaction type feature is specific to SQLite database is not related to transaction isolation level and there are no equivalents for MySQL, Postgre, ODBC sessions, so as I understand we should add it as specific SQLite back-end feature and implement it as property for user to set.
 [SQLite documentation reference to transaction type](https://www.sqlite.org/lang_transaction.html)

bas524:
@kitaisreal 

And what about ```begin concurrent``` for sqlite transaction? It has different behaviour instead of immediate or exclusive transaction.

petko:
Will this be accepted?

obiltschnig:
Will very likely be accepted for 1.12, but not before we actually start work on the 1.12 release. Due to lots of other work we cannot make any predictions when this will happen, but hopefully some time this year.

Issue #1802, opened 2017-07-11
Digest authentication with algorithm SHA-256


BabuGiri:
i see default algorithm configured in latest POCO 1.7.8 is MD5 is tis fixed? or there is a way to configure to latest algorithm SHA-256 as per https://tools.ietf.org/html/rfc7616

aleks-f:
Unfortunately, there is currently no way to change it. Since digest auth is a http feature, it is not a high priority, but if you'd like to see the feature quickly please send a pull (with test).

PR #3026, opened 2020-06-04
HTTPDigestCredentials added support for RFC7616 algorithms

kitaisreal:
HTTPDigestCredentials added support for RFC7617 algorithms

https://tools.ietf.org/html/rfc7616
https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml

1. Added additional algorithms (SHA512/224,SHA 512/256) to SHA2Engine
2. Added different algorithms support in HTTPDigestCredential including (-sess) version of algorithms
3. Added tests

@aleks-f  https://github.com/pocoproject/poco/issues/1802

### DIFF
--- a/Foundation/include/Poco/SHA2Engine.h
+++ b/Foundation/include/Poco/SHA2Engine.h
@@ -38,10 +38,12 @@ class Foundation_API SHA2Engine: public DigestEngine
 public:
 	enum ALGORITHM
 	{
-		SHA_224 = 224,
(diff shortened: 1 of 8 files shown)


ValentinBesnard:
Good job !! I will be happy to use this feature. Do you know if this feature will be part of the next release ? and if yes when ?
Thanks for all the job you do

aleks-f:
it will be in 1.12.0, soon

Issue #3035, opened 2020-06-11
HTTPS proxy support


Jokser:
Hello everyone.

I use Poco HTTP client to work with HTTPS endpoint through an HTTPS proxy but I'm getting timeout errors in such configuration. Looking at HTTP(S)ClientSession source code I see that `proxyConnect()` method establishes a non-secure HTTP connection with a proxy and `proxyRequestPrefix()` returns only "http" proxy prefix:

```
std::string HTTPClientSession::proxyRequestPrefix() const
{
	std::string result("http://");
...
StreamSocket HTTPClientSession::proxyConnect()
{
	ProxyConfig emptyProxyConfig;
	HTTPClientSession proxySession(getProxyHost(), getProxyPort(), emptyProxyConfig);
...
```

As I can understand Poco doesn't support HTTPS proxy.
Is there any way to modify source code to get HTTPS proxy support in the case when both endpoint and proxy work with the same certificates?

obiltschnig:
It should be possible by changing the `HTTPClientSession::proxyConnect()` method to use `Poco::Net::HTTPSessionFactory` to create the `Poco::Net::HTTPClientSession or `Poco::Net::HTTPSClientSession` object instead of a hard-coded `Poco::Net::HTTPClientSession`. 
An additional `protocol` member should be added to `ProxyConfig`, denoting the protocol to use (`http` or `https`). The protocol value can then be given to `Poco::Net::HTTPSessionFactory::createClientSession()` to obtain the session for the proxy request. You'll also need to manage a separate `HTTPSessionFactory` for use by `HTTPClientSession`, as the default one may itself have proxy parameters configured.

Jokser:
@obiltschnig Thank you very much for your feedback. I'll try to fix it in a fork and after success will commit to upstream.

obiltschnig:
What's likely not going to work with this approach is sending HTTP requests through a HTTPS proxy. But that's kind of pointless anyway.

Jokser:
@obiltschnig I've done. Could you please review the change about HTTPS proxy? https://github.com/pocoproject/poco/pull/3040

aleks-f:
#3691

PR #3040, opened 2020-06-19
HTTPS proxy support

Jokser:
https://github.com/pocoproject/poco/issues/3035
Added possibility to connect to HTTPS proxy.
Not all proxies support CONNECT request, so it's possible to send a request directly to HTTPS proxy without CONNECT request. This behavior is controlled by the proxy tunnel option which is 'true' by default for backward-compatibility.
In the case of HTTP endpoint and HTTP proxy, there is no tunneling despite the tunnel option is true by default. This was done also for backward-compatibility.


Akazz:
@obiltschnig Could you please find a moment to check the changes?

obiltschnig:
So far looks good, but will have a more detailed look when I'll merge.

aleks-f:
@Jokser there are conflicts. also, is it possible to have a testcase?

Jokser:
@aleks-f The branch was too stale to make successful rebase and I recreated the new one https://github.com/pocoproject/poco/pull/3691

PR #3058, opened 2020-07-24
Output to stderr in Poco::WindowsConsoleChannel

### LINKED ISSUE #3056: Inconsistent behavior ConsoleChannel vs. WindowsConsoleChannel
Current behavior:

Poco::ConsoleChannel uses std::clog by default, which implies output to **stderr**.
Poco::WindowsConsoleChannel writes (hardcoded) to **stdout** instead.

Correct behavior:

WindowsConsoleChannel should write to stderr by default as well, because this does not interfer with "regular" output to stdout.

Should be trivial to fix (just replacing 2 occurrences of STANDARD_OUTPUT_HANDLE by STANDARD_ERROR_HANDLE in WindowsConsoleChannel.cpp). Do you want a PR for that?

Cheers
 -Fritz

felfert:
- Fixes #3056

### DIFF
--- a/Foundation/src/WindowsConsoleChannel.cpp
+++ b/Foundation/src/WindowsConsoleChannel.cpp
@@ -26,7 +26,7 @@ WindowsConsoleChannel::WindowsConsoleChannel():
 	_isFile(false),
 	_hConsole(INVALID_HANDLE_VALUE)
 {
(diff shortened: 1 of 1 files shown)


felfert:
The arm64 build on travis seems to be stuck in booting the VM.

PR #3067, opened 2020-08-08
Fix pkg-name in find_package_handle_standard_args

### LINKED ISSUE #3066: CMake warning about MYSQL casing
When including Poco in our cmake project we get
```
CMake Warning (dev) at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:273 (message):
  The package name passed to `find_package_handle_standard_args` (MYSQL) does
  not match the name of the calling package (MySQL).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  3rdparty/poco/cmake/FindMySQL.cmake:107 (find_package_handle_standard_args)
  3rdparty/poco/CMakeLists.txt:117 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Krzmbrzl:
For the MySQL package find_package_handle_standard_args got passed "MYSQL" as the package name, which differs from
the actual package name. Due to this cmake (3.18) emitted a warning.
This commit fixes this issue by also passing "MySQL" to find_package_handle_standard_args.

Fixes #3066


Bjoe:
But still the question is, should we also "camel-case" the other Variable Names?
See https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names

But this will break others build system. Maybe rename `FindMySQL.cmake` to `FindMYSQL.cmake` and add a `FindMySQL.cmake` and give a warning that this is deprecated and they should use `FindMYSQL.cmake` ?

Krzmbrzl:
> But still the question is, should we also "camel-case" the other Variable Names?
See cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names

Yeah that'd probably have to be done as well :thinking: 

> But this will break others build system. Maybe rename FindMySQL.cmake to FindMYSQL.cmake and add a FindMySQL.cmake and give a warning that this is deprecated and they should use FindMYSQL.cmake ?

By renaming the FindMYSQL, it should not interfere with any project using poco, should it? In that case I don't even see the need for providing a FindMySQL at all :eyes:

obiltschnig:
Leaving it at FindMySQL.cmake for now, as we have the same issue with PostgreSQL.

Issue #3079, opened 2020-08-20
Multiple issues regarding ApacheConnector


felfert:
Current state as in devel:

1. It is never built due to a typo in the toplevel CMakeList.txt
2. If (1) is fixed, It does build on ubuntu but not on windows for
    multiple reasons. (Needs fixes in FindAPR.cmake, FindAPRUTIL.cmake and FindApache2.cmake)
3. The signature of ap_log_error is wrong. It spits out a warning only, but looking into that, it appears
    that the existing code is meant for some ancient apache version which nobody uses anymore.
4. The handler for the AddPocoRequestHandler command cannot cope with whitespace in the SharedLib
    argument (no string quoting supported)
5. The handler for the AddPocoConfig command has no problem with whitespace in the name, but does not
    support string quoting either.
6. On windows, the module should be named mod_poco.so (just like on linux). Apache itself on windows uses
    the same naming and considers the .dll extension deprecated for modules.
7. Chunked transfer encoding in responses is not supported.
8. URI filter behaves wrong, E.g.: declaring `/foobar` enables `/f /fo /foo /foob` and `/fooba` which it shouldn't.

I have worked on all but (7) and have a working version [here](https://github.com/felfert/poco/tree/fix-cmake-apacheconnector). (Tested on windows and ubuntu 18.04 LTS)
Would you accept a PR, even if (7) is not yet implemented?

Cheers
 -Fritz

PR #3082, opened 2020-08-21
Various fixes for ApacheConnector

felfert:
- Fix typo in CMakeLists.txt
 - Adapt ApacheConnector to apache httpd 2.4
 - Fix target naming (no lib prefix) and install target
 - Fix build on windows
 - Use recommended .so suffix  for mod_poco even on Windows
 - Handle quoted dll name in AddPocoRequestHandler command
 - Handle quoted string in AddPocoConfig command
 - Disable chunked transfer encoding in samples and doc

Fixes (all except chunked transfer encoding) #3079 

Note:
  Build tested on windows against the [ApacheLounge](https://www.apachelounge.com/download/VC15/) binary apache distribution (the one from ApacheHaus unfortunately is missing some headers) using the following options:

````
-DAPACHE2_ROOT_DIR=/Apache24 -DAPR_ROOT_DIR=/Apache24 -DAPRUTIL_ROOT_DIR=/Apache24 -DENABLE_APACHECONNECTOR=ON
````
Cheers
 -Fritz

### CODE REVIEW

File: ApacheConnector/CMakeLists.txt [hunk 1]
```diff
@@ -19,7 +21,14 @@ target_include_directories(mod_poco
 		${APACHE2_INCLUDE_DIRS}
 		${CMAKE_CURRENT_SOURCE_DIR}/src
 )
-target_link_libraries(mod_poco PUBLIC Poco::Util Poco::Net Apache::Apr Apache::Aprutil)
+set(DEPLIBS Poco::Util Poco::Net Apache::Apr Apache::Aprutil)
+if(WIN32)
+	list(APPEND DEPLIBS Apache::httpd)
+endif(WIN32)
+target_link_libraries(mod_poco PUBLIC ${DEPLIBS})
```
Bjoe, line 28:
```suggestion
target_link_libraries(mod_poco PUBLIC Poco::Util Poco::Net Apache::Apr Apache::Aprutil)
if(WIN32)
	target_link_libraries(mod_poco PUBLIC Apache::httpd)
endif(WIN32)
```
Keep it simple :-)

ApacheConnector/CMakeLists.txt
```diff
@@ -19,7 +21,14 @@ target_include_directories(mod_poco
...
+
+file(MAKE_DIRECTORY "include/Poco")
```
Bjoe, line 30: Not sure why we need this ? Should be created with this `POCO_INSTALL` macro

Bjoe (changes requested):
Please fix findings

ApacheConnector/CMakeLists.txt
felfert: I think this suggestion is wrong.
AFAIK, target_link_libraries() is **NOT** cumulative. So: your suggestion would result in missing link libraries
Poco::Util Poco::Net,  Apache::Apr and Apache::Aprutil
if building on Windows?!

ApacheConnector/CMakeLists.txt
aleks-f: @felfert what is the purpose of this line?

PR #3101, opened 2020-09-18
FileChannel: make it possible to override ArchiveStrategy and RotateStrategy

oktal:
Currently, the `FileChannel` offer limited variants of `ArchiveStrategy` and `RotateStrategy`, based on the properties of the channel.

However, it might be useful for the user to provide its own `ArchiveStrategy` or `RotateStrategy`.
Such a use case would be, for example, to add a header every-time a new log file is created (or rotated).

Without access to the `ArchiveStrategy`, this is currently not possible.

This Pull Request just makes it possible for the user to inherit from `FileChannel` and override `createArchiveStrategy` or `createRotationStrategy` virtual functions.

Note that this PR also adds a new `open()` function to the `ArchiveStrategy` interface so that the user can know from the strategy that a log file has been opened by the `FileChannel` and implement custom behavior from there (such as writing a header to the file).

### DIFF
--- a/Foundation/include/Poco/ArchiveStrategy.h
+++ b/Foundation/include/Poco/ArchiveStrategy.h
@@ -42,6 +42,7 @@ public:
 	ArchiveStrategy();
 	virtual ~ArchiveStrategy();
 
+	virtual LogFile* open(LogFile* pFile) = 0;
 	virtual LogFile* archive(LogFile* pFile) = 0;
 		/// Renames the given log file for archiving
 		/// and creates and returns a new log file.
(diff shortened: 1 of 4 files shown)


obiltschnig:
Thanks, but please clean-up the pull request as it currently has 178 (!) modified files.

oktal:
Yes, I initially branched out from `master`. I rebased my branch on top of `devel`, the PR should be clean now.

obiltschnig:
Thank you!

aleks-f:
@oktal can you please write tests for this feature?

aleks-f:
https://github.com/pocoproject/poco/pull/3810

PR #3102, opened 2020-09-18
Made it possible to use lowercase hex numbers, also when encoding JSON

Archipel:
I needed this feature for a project, perhaps it can be of help to someone else as well.


aleks-f (changes requested):
please write some tests for this

PR #3017, opened 2020-05-30
Fixes from develop experimental (for devel)

matejk:
This pull request contains a few fixes and improvements that were originally merged to develop but then did not make it to releases. Changes are cherry picks or reimplementations of the following:

* #2470 (Can't use Poco::MongDB::Cursor on aggregation curcor), PR #2471
* #860 (Poco::MongoDB connection pool and multithreading), PR #1146

The improvement to get numeric thread ID and have the possibility to use it in Logger could not be cherry picked anymore, so I re-implemented it on master. Changes are based on:

* #2465 (Operating system specific thread ID is not available any more in Logger/Formatter), PR #2466
* PR #846 (Posix OS thread ID, added %O (thread ID) to logger)

PR #3103, opened 2020-09-18
Fix log thread identifier being 0 when not using Poco::Thread

oktal:
`Poco::Message` now falls back to system's thread id when the underlying `Poco::Thread` is null.

### DIFF
--- a/Foundation/src/Message.cpp
+++ b/Foundation/src/Message.cpp
@@ -17,6 +17,10 @@
 #if !defined(POCO_VXWORKS)
 #include "Poco/Process.h"
 #endif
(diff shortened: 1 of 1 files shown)


obiltschnig:
That's a bit problematic as this only works on Windows and Linux (and will actually cause the build to fail on other Unix systems like macOS). Also, the system's thread ID is from an different number space than the one by Thread::id(), which could cause confusion.

matejk:
Please see also pull request #3017, where similar functionality is implemented slightly differently.

oktal:
Thanks, closing in favor of #3017

PR #3104, opened 2020-09-18
Publicly expose Poco::Message parameters

oktal:
It is now possible to retrieve all the parameters from a `Poco::Message`.


obiltschnig:
Needs some formatting fixes, but otherwise looks good.

PR #3105, opened 2020-09-18
cmake: use GNUInstallDirs

lopsided98:
POCO currently hardcodes the installation directories relative to the installation prefix. This makes it impossible to split the installation; for example to install the libraries and headers to different prefixes.

This PR makes POCO use the standard [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) mechanism to specify the installation directories.

The only user visible impact of this should be that anyone who explicitly passes `LIB_SUFFIX` to the build will have to use `CMAKE_INSTALL_LIBDIR` instead.


lopsided98:
I updated the patch to fix another hardcoded reference to `include`.

gocarlos (approved):

bachp:
@obiltschnig I think it makes sense to merge this.

gocarlos:
great

PR #3114, opened 2020-10-01
Added JSON Array::empty() method

pkl97:
To be in sync with the underlying std::vector and to allow more concise code being written I added an empty() method to Poco::JSON::Array.


Mironenko:
There is the test `testEmptyArray` in JSON testsuite. Looks like it's a good place to test the new method.

pkl97:
The existing test cases have been extended to use the new empty() method.

PR #3115, opened 2020-10-03
Change EventHandlerMap key from Socket to poco_socket_t

micheleselea:
the EventHandlerMap key now is Socket: that could end up in some coding problem if you for example have to accept a socket and than make it ssl handshake. That way we make the same key used in PollSet


aleks-f:
@micheleselea ok, but this pull should go to [devel](https://github.com/pocoproject/poco/tree/devel) branch, not master

micheleselea:
OK right, my bad, I resubmit on monday

PR #3116, opened 2020-10-05
Changed EventHandlerMap key

micheleselea:
Changed EventHandlerMap key from Socket to poco_socket_t to avoid errors in removing/access EventHandlerMap when for example we make an SSL handshake. This way we have the same behavior PollSet has

### DIFF
--- a/Net/src/SocketReactor.cpp
+++ b/Net/src/SocketReactor.cpp
@@ -179,12 +179,14 @@ bool SocketReactor::hasEventHandler(const Socket& socket, const Poco::AbstractOb
 
 SocketReactor::NotifierPtr SocketReactor::getNotifier(const Socket& socket, bool makeNew)
 {
-	if (socket.impl() == nullptr) return 0;
+	const SocketImpl* pImpl = socket.impl();
+	if (pImpl == nullptr) return 0;
+	poco_socket_t sockfd = pImpl->sockfd();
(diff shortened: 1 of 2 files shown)


Net/src/SocketReactor.cpp
```diff
@@ -179,26 +179,28 @@ bool SocketReactor::hasEventHandler(const Socket& socket, const Poco::AbstractOb
 
 SocketReactor::NotifierPtr SocketReactor::getNotifier(const Socket& socket, bool makeNew)
 {
+	if (socket.impl() == nullptr) return 0;
```
aleks-f, line 182: @micheleselea let's cache the `SocketImpl*` here and also the `poco_socket_t` below to avoid repeated function calls. I know they are inlines but inlining is not guaranteed

aleks-f (changes requested):

Net/src/SocketReactor.cpp
micheleselea: Ok correct point, how can I make change, let me know if it's ok because I don't remember if I use correct procedure

aleks-f (approved):

micheleselea:
About the handler key change, be careful when you add or remove reactor's observers that the sockfd has to be a valid fd: so you cannot for example do addObserver than close socket and than removeObserver because you will end up with a no remove action. I don't know if it's better to explain that

PR #3136, opened 2020-11-10
Fixed null character issue when parsing a JSON

msi22:
Consider the length of string values to ensure that strings, containing a null character, get not truncated at the null character.

### DIFF
--- a/JSON/src/ParserImpl.cpp
+++ b/JSON/src/ParserImpl.cpp
@@ -204,8 +204,15 @@ void ParserImpl::handle()
 			break;
 		}
 		case JSON_STRING:
-			if (_pHandler) _pHandler->value(std::string(json_get_string(_pJSON, NULL)));
+		{
+			if (_pHandler) 
+			{
+				std::size_t length = 0;
+				const char* val = json_get_string(_pJSON, &length);
+				_pHandler->value(std::string(val, length == 0 ? 0 : --length)); // Decrease the length by 1 because it also contains the terminating null character
+			}
 			break;
+		}
 		case JSON_OBJECT:
 			if (_pHandler) _pHandler->startObject();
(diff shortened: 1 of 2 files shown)


Mironenko:
How can these functionality be tested?

msi22:
The conversion to JSON worked correctly before this fix but the parsing failed, when the null character was part of the content. To test this issue a null character can be added to a value. The null character should still be there after creating the JSON and parsing the JSON.

Mironenko:
> 
> 
> To test this issue a null character can be added to a value. The null character should still be there after creating the JSON and parsing the JSON.

I admit that would be better PR if you added an unittest for this case.

Mironenko:
For me it looks good. CM in imperative mood would be conventional. Not having unnecessary side effect from `--length` could be more accurate too, but in general PR is good.

msi22:
json_get_string() returns the buffer length including the null character and I think it would be no good idea to change this behavior in this fix. For the string the length is needed without the null character. Therefore this transformation is needed. Side effects of future changes should be covered by the test case.

Are there any further changes needed for the PR?

Mironenko:
> json_get_string() returns the buffer length including the null character and I think it would be no good idea to change this behavior in this fix. For the string the length is needed without the null character. Therefore this transformation is needed. Side effects of future changes should be covered by the test case.

I suggested replacing `--length` with `length-1`, that's where the side effect is. But it does not matter much

I would also suggest changing CMs to:

```
Fix parsing JSON having string containing null character

Before this commit strings containing `\u0000` character got truncated.
This commit fixes the issue.
```

```
Add test on parsing JSON having string containing null character
```
What is left is to wait for @obiltschnig or @aleks-f to notice the PR.

PR #3146, opened 2020-11-19
WIP: Feat/data db encoding

aleks-f:
Adds `dbEncoding` property to session (when different from `UTF-8` used by extractor to convert strings)


Data/ODBC/include/Poco/Data/ODBC/Extractor.h
```diff
@@ -53,7 +55,8 @@ class ODBC_API Extractor: public Poco::Data::AbstractExtractor
 	typedef Preparator::Ptr PreparatorPtr;
 
 	Extractor(const StatementHandle& rStmt, 
-		Preparator::Ptr pPreparator);
+		Preparator::Ptr pPreparator,
+		const std::string& dbEncoding = "UTF-8");
```
obiltschnig, line 59: I would pass in the Encoding object (Ptr) directly, to avoid having to look it up every time. Encoding objects are immutable, so this is safe.

Data/ODBC/include/Poco/Data/ODBC/Extractor.h
```diff
@@ -574,6 +577,50 @@ class ODBC_API Extractor: public Poco::Data::AbstractExtractor
...
+			throw InvalidAccessException("Direct container extraction only allowed for bound mode.");
+		val.clear();
+		if (ret)
+		{
+			val.resize(result.size());
+			C::iterator vIt = val.begin();
+			C::iterator it = result.begin();
+			for (; it != result.end(); ++it, ++vIt)
+			{
+				Poco::TextEncoding::Ptr pDataEncoding = Poco::TextEncoding::find(_dbEncoding);
```
obiltschnig, line 597: Can avoid the lookup if the Poco::TextEncoding::Ptr is passed via the constructor. Check for `nullptr` to avoid transcoding, which has a performance penalty.

obiltschnig (changes requested):

Data/ODBC/include/Poco/Data/ODBC/Extractor.h
aleks-f: I've done something along those lines, but on the constructor side, see https://github.com/pocoproject/poco/pull/3146/commits/cf2d7c75c10a639b3ebf17e44285ec2ee3ed8c32

aleks-f: probably also a good idea to verify that the encoding is valid early (ie. immediately when it is set), I'll add that check

PR #3149, opened 2020-11-20
Add fuzzer

AdamKorcz:
I have been working on setting up continuous fuzzing of Poco by way of OSS-fuzz.
Fuzzing is a way of testing applications, whereby pseudo-random data is passed to the target - in this case the json parser - with the goal of finding bugs and vulnerabilities.

This PR adds a fuzzer implemented by way of LibFuzzer. I have also integrated this fuzzer into OSS-fuzz's infrastructure, and I will be happy to integrate Poco into the OSS-fuzz project. This will allow Google to run this fuzzer and all future fuzzers in the Poco project and notify maintainers with detailed reports in case bugs are found. It is a service that is offered free of charge to open source projects with an implied expectation that bugs are fixed, so that the resources spent on fuzzing Poco are put to good use.

To finalize the integration into OSS-fuzz, at least one maintainer email address is needed. This can be shared here, via email to me or on the OSS-fuzz side of things when I submit Poco. All maintainer email addresses will be added to a public list that can be changed at any time.

OSS-fuzz has a 90 day public disclosure policy.

### DIFF
--- a/JSON/testsuite/fuzz/json_parse_fuzzer.cc
+++ b/JSON/testsuite/fuzz/json_parse_fuzzer.cc
@@ -1,0 +1,19 @@
+#include "../src/JSONTest.h"
+
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+        std::string json(reinterpret_cast<const char*>(data), size);
+        Poco::JSON::Parser parser;
+
+        Poco::Dynamic::Var result;
+        try
+        {
+                result = parser.parse(json);
+        }
+        catch(const std::exception& e)
+        {
+                return 0;
+        }
+        return 0;
+}


obiltschnig:
Thanks, looks interesting. Might also make sense for other parts, e.g. HTTP request/response parser, XML parser, etc.

AdamKorcz:
I agree. There are definitely other obvious targets like the ones you mention. For the record, I would be glad to write fuzzers for those as well.

AdamKorcz:
If there is anything I can do from my self please let me know.

obiltschnig:
Plan to look into this in February, when I'll work on the next release. Too busy with other projects at the moment.

aleks-f:
@obiltschnig merge this for 1.13?

aleks-f:
@AdamKorcz does [this pull](https://github.com/pocoproject/poco/pull/3882) supersedes this PR? If not, what exactly is required from our end to merge this one? My email is at(alex, dot(pocoproject, org)). I am lead and maintainer, but have no experience with OSS-fuzz.

PR #3151, opened 2020-11-23
fix(JSON::Object): crash when a key is removed from object with JSON_PRESERVE_KEY_ORDER

fenghao119:
If you remove a key from the JSON object with option JSON_PRESERVE_KEY_ORDER, it will cause a crash.


Mironenko:
Hi! Fix looks great! Hope it won't be neglected. Couldn't you please rephrase CM in more [regular way](https://chris.beams.io/posts/git-commit/)?

PR #3155, opened 2020-11-29
Fixed typo in overridden

pkl97:
The term "overridden" was missing one of the "d" characters a few times.


obiltschnig:
Fixed in 1.10.2

PR #3192, opened 2021-01-27
[WIP] add SSL to MongoDB

knoguchi:
This patch adds SSL support for MongoDB.


knoguchi:
I was wondering why Travis is failing.  I didn't know many builds use `configure` script.  Why not cmake for all?

PR #3195, opened 2021-01-29
MinGW also defines __GNUC__.

iSLC:
`POCO_COMPILER_MINGW` is never defined because MinGW also defines `__GNUC__`. The solution is to either check for `__MINGW32__`/`__MINGW64__` before GCC but then `POCO_COMPILER_GCC` is not defined. Or duplicate the MinGW branch inside the GCC branch since they would be interchangeable.


aleks-f:
please redirect this pull to devel branch

PR #3208, opened 2021-02-11
add github actions for CI

bachp:
In this first step we build only for Linux via CMake.

Open Points:

- [ ] Execute tests, the are not runable via ctest right now
- [ ] Add more configurations via [matrix build](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix)
  - [x] Different OS
  - [x] Different architectures
  - [x] Different versions

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,46 @@
+name: CMake
+
+on: [push]
+
+env:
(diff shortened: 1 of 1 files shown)


aleks-f:
@bachp before merging, this should be redirected to [devel](https://github.com/pocoproject/poco/tree/devel)

bachp:
@aleks-f Sure, for now this is just me trying if github actions are suitable for our usecase.

PR #3228, opened 2021-03-24
Fix GlobTest testcase.

arrowd:
The current CMake code causes directories like "Crypto/testsuite/..." to appear, which gets picked up by "globtest/../*/" glob.


obiltschnig:
CMake builds and CTest seem to work fine. Postponing this issue for later.

arrowd:
Seems to be not relevant anymore.

PR #3233, opened 2021-04-05
Feat/hash range

aleks-f:
port of boost::hash_range


Foundation/include/Poco/Hash.h
```diff
@@ -99,6 +110,90 @@ inline std::size_t hash(UInt64 n)
 }
 
 
+namespace Impl {
+
+
+template <typename SizeT>
+inline void HashCombine(SizeT& seed, SizeT value)
```
obiltschnig, line 117: Function names should be `camelCase`.

obiltschnig:
@aleks-f can you re-target against poco-1.11.0 branch? After the release we'll rebase devel to poco-1.11.0.

Issue #3215, opened 2021-03-08
XML parser returns item from different element in a array


Erbengi:
Example:

```c++
std::stringstream sstream;
sstream << R"(
<array>
  <item>
    <string>foo</string>
    <bool>true</bool>
  </item>
  <item>
    <string>bar</string>
    <bool>false</bool>
    <float>0.4</float>
  </item>
  <item>
    <string>test</string>
    <bool>false</bool>
  </item>
</array>
)";

Poco::XML::InputSource src(sstream);
Poco::XML::DOMParser parser;
auto doc = parser.parse(&src);

auto node = doc->getNodeByPath("/array/item[0]/float");
std::cout << (node ? node->innerText() : "[not found]") << std::endl;

node = doc->getNodeByPath("/array/item[1]/float");
std::cout << (node ? node->innerText() : "[not found]") << std::endl;

node = doc->getNodeByPath("/array/item[2]/float");
std::cout << (node ? node->innerText() : "[not found]") << std::endl;
```

Output:
```
0.4
0.4
[not found]
```

Poco returns `float` element for `item[0]` even the first element doesn't contain it. Tested with Poco 1.10.1.

obiltschnig:
Required a different fix than proposed.

PR #3239, opened 2021-04-10
XML parser returns item from different element in a array #3215

xvqxy:
Fixes issue where parser is returning node located after the valid one
if targetted one does not meet criteria and any subsequent does.


obiltschnig:
Needs different fix, this may break behavior in other cases.

Issue #3245, opened 2021-04-19
find_package(Poco REQUIRED COMPONENTS ... NetSSL) requires an aditional find_package(OpenSSL) since poco-10


strahlc:
After switching from poco-1.9.4 to poco-10.0.1, my library doesn't compile any more, because cmake couldn't find the target for OpenSSL.

With the old version I was using:

```cmake
find_package(Poco REQUIRED COMPONENTS Data DataSQLite Foundation Net NetSSL Util)

...

include_directories(
        Poco::Data
        Poco::DataSQLite
        Poco::Foundation
        Poco::Net
        Poco::NetSSL
        Poco::Util
)

add_library(myLib SHARED
        ${mySrcs}
        )

target_link_libraries(myLib PUBLIC
        Poco::Data
        Poco::DataSQLite
        Poco::Foundation
        Poco::Net
        Poco::NetSSL
        Poco::Util
        )
```

with the new poco version I get this error from cmake:

```
CMake Error at src/CMakeLists.txt:107 (add_library):
  Target "myLib" links to target "OpenSSL::SSL" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

CMake Error at src/CMakeLists.txt:107 (add_library):
  Target "myLib" links to target "OpenSSL::Crypto" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

If I manually add ```find_package(OpenSSL REQUIRED)``` to my `CMakeList.txt`, everything works like before, but ```find_package(Poco REQUIRED COMPONENTS ... NetSSL)``` should already find this required dependency like before.

strahlc:
Here is the diff from the generated cmake files

```patch
--- /home/strahlc/Poco-1.9.4/PocoNetSSLTargets.cmake
+++ /usr/lib64/cmake/Poco/PocoNetSSLTargets.cmake
@@ -63,9 +63,8 @@
 add_library(Poco::NetSSL SHARED IMPORTED)
 
 set_target_properties(Poco::NetSSL PROPERTIES
-  INTERFACE_COMPILE_DEFINITIONS "POCO_NO_AUTOMATIC_LIBS"
   INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include"
[...]

PR #3157, opened 2020-12-02
fix(openssl): add missing dependency to OpenSSL components

andy9a9:
```
| -- Configuring done
| CMake Error at src/server/CMakeLists.txt:21 (add_executable):
|   Target "test" links to target "OpenSSL::SSL" but the target was not
|   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
|   an ALIAS target is missing?
| 
| 
| CMake Error at src/server/CMakeLists.txt:21 (add_executable):
|   Target "test" links to target "OpenSSL::Crypto" but the target was
|   not found.  Perhaps a find_package() call is missing for an IMPORTED
|   target, or an ALIAS target is missing?
| 
| 
| -- Generating done
| CMake Warning:
|   Manually-specified variables were not used by the project:
| 
|     CMAKE_INSTALL_BINDIR
|     CMAKE_INSTALL_DATAROOTDIR
|     CMAKE_INSTALL_INCLUDEDIR
|     CMAKE_INSTALL_LIBDIR
|     CMAKE_INSTALL_LIBEXECDIR
|     CMAKE_INSTALL_LOCALSTATEDIR
|     CMAKE_INSTALL_SBINDIR
|     CMAKE_INSTALL_SHAREDSTATEDIR
|     CMAKE_INSTALL_SYSCONFDIR
|     PYTHON_EXECUTABLE
|     Python3_EXECUTABLE
|     Python_EXECUTABLE
| 
| 
| CMake Generate step failed.  Build files cannot be regenerated correctly.
```
Even if `find_package(Poco REQUIRED COMPONENTS Util Foundation Net NetSSL)` were used, OpenSSL was not found. It has to be done automatically, when NetSSL/Crypto component is used.

PR #3254, opened 2021-04-21
Fix missing OpenSSL dependency

Bjoe:
Should fix issue #3245


Bjoe:
This should also merge into `poco-1.10.2` and `devel` ...

aleks-f:
> This should also merge into `poco-1.10.2` and `devel` ...

but why to 1.10.1? That is already released, so nothing goes there anymore

andy9a9:
Duplicate of #3157

strahlc:
@aleks-f , @Bjoe please merge #3157 instead

PR #3333, opened 2021-06-29
Add NativeThreadInfo and blend in Message

osmeest:
+ test with PatternFormatter

### DIFF
--- a/Foundation/src/Message.cpp
+++ b/Foundation/src/Message.cpp
@@ -18,6 +18,7 @@
 #include "Poco/Process.h"
 #endif
 #include "Poco/Thread.h"
+#include "Poco/NativeThreadInfo.h"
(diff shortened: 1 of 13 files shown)


aleks-f:
few things missing here:
- [x]  add new files to Makefile
- [ ] ~~add new files to VS90 project and regenerate all VS projects/solutions using Progen~~

Foundation/src/Message.cpp
```diff
@@ -133,6 +134,11 @@ void Message::init()
 		_tid    = pThread->id();
 		_thread = pThread->name();
 	}
+	else {
```
bas524, line 137: it should be formatted as
```cpp
}
else
{
```

aleks-f:
#5135

PR #3360, opened 2021-07-27
Add POCO_PGSQL_{INCLUDE,LIB} variables

ffontaine:
Add `POCO_PGSQL_{INCLUDE,LIB}` variables to avoid the following build failure when cross-compiling with buildroot:

`arm-linux-g++: ERROR: unsafe header/library path used in cross-compilation: '-L/usr/lib/postgresql'`

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


obiltschnig:
see 931143df8

PR #3363, opened 2021-08-08
Fixed compilation error with MongoDB::Connection and Util::Application

rudolfheszele:
MongoDB/Connection.h had to be included before Util/Application.h
on Windows with VS-2019, otherwise the compilation failed.
The root cause of this was that including Util/Application.h
introduced a Message type under the Poco namespace, and
the MongoDB::MessageHeader class was friended with that type,
instead of MongoDB::Message. A simple solution for this is to
explicitly forward declare the Message class in the MongoDB namespace.

I am not sure if the changes in the tests are needed, I do not like adding
a new dependency on Poco::Util just to test this, but it is up for discussion.


obiltschnig:
See f71cc32f0 - did not add dependency to Util to testsuite.

PR #3384, opened 2021-08-24
Always set thread names on POSIX platforms

carun:
I'm not sure about the original intent to hide it under a DEBUG macro.

Naming the threads in release mode makes it easier to see runtime
application and know which thread pool uses how many threads and what
their names are. Firefox, Chromium and many other apps do this on Linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,10 @@ cmake-build/
 *.bak
 stage/
 releases/
+
+# vim #
+#######
+*.orig
+*.swp
+*.vim
+tags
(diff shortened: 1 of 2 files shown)


aleks-f:
@carun please redirect this to devel branch

aleks-f (changes requested):
this pull needs to go against the devel branch

aleks-f (changes requested):
this pull needs to go against the devel branch

aleks-f (approved):

PR #3392, opened 2021-08-26
Modify mysql testsuite solution to be compatible with mysql8.0:

qusuo:
1、transaction_isolation was added in MySQL 5.7.20 as an alias for tx_isolation, which is now deprecated and is removed in MySQL 8.0;
2、mysql8.0 Default character set is utf8mb4;


aleks-f:
fixed in https://github.com/pocoproject/poco/pull/3490

Issue #3221, opened 2021-03-15
Crash reported on Windows in X509Certificate verification


jdee:
We've received reports of an enterprise app using an SDK built on Poco that often experiences crashes at this location:

```
00 (Inline Function) --------`--------     EnterpriseApp!std::_WChar_traits<wchar_t>::length+0xa [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.26.28801\include\xstring @ 168]
01 (Inline Function) --------`--------     EnterpriseApp!std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >::assign+0xa [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.26.28801\include\xstring @ 2992]
02 (Inline Function) --------`--------     EnterpriseApp!std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >::{ctor}+0x1b [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.26.28801\include\xstring @ 2344]
03 000000bc`387fe430 00007ff7`e983ae6b     EnterpriseApp!Poco::Net::X509Certificate::extractNames+0x2a4 [D:\path\to\poco\NetSSL_Win\src\X509Certificate.cpp @ 262]
04 000000bc`387fe590 00007ff7`e983eee2     EnterpriseApp!Poco::Net::X509Certificate::verify+0x9b [D:\path\to\poco\NetSSL_Win\src\X509Certificate.cpp @ 478]
05 000000bc`387fe6c0 00007ff7`e983eb91     EnterpriseApp!Poco::Net::SecureSocketImpl::clientVerifyCertificate+0x82 [D:\path\to\poco\NetSSL_Win\src\SecureSocketImpl.cpp @ 1173]
```

This happens with Poco 1.9.4 in production releases using `HTTPSClientSession` for API calls, but apparently only for a limited subset of users. For them, it is consistent. It seems to be directly correlated with Kaspersky. When users disable Kaspersky, the issue is resolved. We cannot directly reproduce in testing. The app involved is a Win32 app, packaged without MSIX, running on Windows 10.

https://github.com/pocoproject/poco/blob/poco-1.9.4/NetSSL_Win/src/X509Certificate.cpp#L262

[...]

PR #3397, opened 2021-08-31
Fix crash due to X.509 certificates with Subject Alternative Name other than DNS Host

echo-branch:
This PR fixes https://github.com/pocoproject/poco/issues/3221. 

Modified `X509Certificate.cpp` to skip Subject Alternative Name entries other than DNS Host. This fixes a crash.

Modified `SecureSocketImpl.cpp` to fix an error check issue. The comment indicates the code should have ignored errors other than CRYPT_E_REVOKED, however `revStat.dwIndex < certs.size() - 1` is true when you have any error.

### DIFF
--- a/NetSSL_Win/src/Context.cpp
+++ b/NetSSL_Win/src/Context.cpp
@@ -275,11 +275,11 @@ void Context::acquireSchannelCredentials(CredHandle& credHandle) const
 
 	schannelCred.grbitEnabledProtocols = proto();
 
-    // Windows NT and Windows Me/98/95: revocation checking not supported via flags
-    if (_options & Context::OPT_PERFORM_REVOCATION_CHECK)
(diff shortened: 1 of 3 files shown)


echo-branch:
@obiltschnig 
Hi, just curious if the 1.11.1 release is going to hit its target release date. Would rather use an official release, rather than patched version.

echo-branch:
Just want to share detailed repro steps with Kaspersky. Though any X.509 certificate with SAN entries other than DNS Host will also repro the issue.

1. Install Kaspersky Total Security for Windows. The California binary is fine.
2. Open the Kaspersky "Settings" and navigate to the "Network settings".
3. Under "Encrypted connection scanning" select "Always scan encrypted connections"
4. Network calls crash due to the SAN "IP Address" entry in Kaspersky's certificate.

obiltschnig:
See 270c2643c

PR #3406, opened 2021-09-07
buildwin.ps1 updated to support the new vcxproj files.  Fixes pocoproject/poco/#3405

Mike4Online:
buildwin.ps1 updated to support the new vcxproj files which contain both Win32 and x64 platform build configurations.  Successfully tested with devenv.  Fixes pocoproject/poco/#3405

### DIFF
--- a/buildwin.ps1
+++ b/buildwin.ps1
@@ -315,7 +315,7 @@ function Exec-Devenv([string] $projectConfig, [string] $vsProject)
 }
 
 
-function Build-Devenv([string] $vsProject, [switch] $skipStatic)
+function Build-Devenv([string] $vsProject, [string]  $Platform, [switch] $skipStatic)
 {
   if ($linkmode -contains "static" -and $skipStatic) { Return }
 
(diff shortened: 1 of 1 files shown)


Mike4Online:
> Shouldn't this go into the devel branch?

@hmartinez82 , @aleks-f -

Which branch this pull request should be merged with is up to the Poco maintainers.  I have no opinion, other than I think the next release of Poco should incorporate these changes to the PowerShell build script.  Further edits to Poco's PowerShell build script may also be warranted:

I've been updating my build of Visual Studio 2019 every few months, and I noticed a change last August.  The `devenv.exe` program used to be able to perform a build of a `.sln` solution or `.vcxproj` project, and then return control to the build script when that solution or project build was done.  But in the newest versions of Visual Studio, `devenv.exe` immediately returns control to the build script while the solution/project build proceeds.  No `ERRORLEVEL` is returned from `devenv.exe`.  This plays havoc with build scripts.

The normal solution to the above issue is to launch `devenv.com` instead of `devenv.exe`.  `devenv.com` waits until the build is done before returning control to the build script.  It also returns an `ERRORLEVEL` and allows console output to be redirected to a file (`devenv.exe` used to behave this way, but no longer).  But I had difficulties getting my builds to succeed with `devenv.com`.  So ultimately I rewrote my application's Windows build script to build entirely with `MSBuild.exe`.  

I have not attempted to rewrite the Poco PowerShell build script to use `MSBuild.exe` rather than `devenv.exe`.  Doing so might help ensure reliable builds of Poco on Windows using the latest versions of Visual Studio.  One requirement might be needed: to open a Visual Studio Native Tools x86 or x64 Command Prompt before running the PowerShell build script.  This will ensure the various environment variables needed by `MSBuild.exe` are set.

I suggest testing Poco's PowerShell build script using the latest versions of Visual Studio 2019 and 2022.

hmartinez82:
Thank you for the hard work. I'm waiting for your changes to be merged because I want to add the Arm64 changes after.

aleks-f:
> @Mike4Online @aleks-f I'm not sure about master vs devel branches. Shouldn't this go into the devel branch?

please redirect this to **devel** branch

PR #3448, opened 2021-10-29
Basic support for OpenSSL 3.0.0

### LINKED ISSUE #3415: OpenSSL 3.0 support
OpenSSL 3.0 was recently released and Fedora is already adopting it for F36.

Poco seems to build okay (with a few warnings), but some tests are failing:

```
The following tests passed:
	Encodings
	XML
	JWT
	JSON
	ActiveRecord
	Zip
	Data
	Util
	DataSQLite
	Net
	Foundation
85% tests passed, 2 tests failed out of 13
Total Test time (real) = 113.11 sec
The following tests FAILED:
	  8 - NetSSL (Failed)
	  9 - Crypto (SEGFAULT)
```

cheese:
Fixes #3415


micheleselea:
I think this
```
#if OPENSSL_VERSION_NUMBER >= 0x30000000L
	OSSL_PROVIDER_load(NULL, "default");
	OSSL_PROVIDER_load(NULL, "legacy");
#endif
```
should be done just once in the openssl inizializer. And we shold get the returned pointer to
`OSSL_PROVIDER_unload`
on exit

cheese:
> I think this
> 
> ```
> #if OPENSSL_VERSION_NUMBER >= 0x30000000L
> 	OSSL_PROVIDER_load(NULL, "default");
> 	OSSL_PROVIDER_load(NULL, "legacy");
> #endif
> ```
> 
> should be done just once in the openssl inizializer. And we shold get the returned pointer to `OSSL_PROVIDER_unload` on exit

This is not quite necessary if Poco doesn't have its own OpenSSL library context. There are several design choices for how to deal with OpenSSL library context. Just use the default one? Create a private one? Let the user prepare and pass one? ...

micheleselea:
Beware to "Implement SSL abort handling on OpenSSL 3"
I'm trying that error handling but I found an error when I use WebSockets server side. I think that somehow an error is triggered, but should not, and throw me an unexpected exception: unexpected because with the prev error handler version does not happen. I revert have to revert the change

andy9a9:
Fixes #3504

obiltschnig:
@micheleselea can you check the poco-1.11.2 branch - I've made some minor changes to the abort handling code"

micheleselea:
ok I'll check tomorrow

micheleselea:
@obiltschnig I'm going to check the 1.11.2 changes.
I found that with the Openssl 3 I added a `setSecurityLevel(int)` function that do something like

```
void Context::setSecurityLevel(int level)
{
	if (level < 0)
		return;
#if OPENSSL_VERSION_NUMBER >= 0x030000000L
	SSL_CTX_set_security_level(_pSSLContext, level);
#endif
}
```

Called inside `void Context::init(const Params& params)` is useful for deal with old webservers or clients, because some ciphersuite is disabled by default.



another change for security reason I made to FTPSClientSession::establishDataConnection

```
StreamSocket FTPSClientSession::establishDataConnection(const std::string& command, const std::string& arg)
{
	beforeCreateDataSocket();

	StreamSocket ss = FTPClientSession::establishDataConnection(command, arg);
	ss.setNoDelay(true);

	// SSL nogotiating of data socket
	if (_secureDataConnection && _pControlSocket->secure())
	{
		// We need to reuse the control socket SSL session so the server can ensure that client that opened control socket is the same using data socket
		Poco::Net::SecureStreamSocketImpl* pSecure = dynamic_cast<Poco::Net::SecureStreamSocketImpl*>(_pControlSocket->impl());
		if (pSecure != nullptr)
		{
			Poco::Net::SecureStreamSocket sss(Poco::Net::SecureStreamSocket::attach(ss, pSecure->context(), pSecure->currentSession()));
			ss = sss;
			if (_forceSessionReuse) {
				sss.setLazyHandshake(false);
				if (sss.completeHandshake() != 1) {
					throw Poco::Net::FTPException("SSL Session HANDSHAKE error");
				}
				if (!sss.sessionWasReused()) {
					throw Poco::Net::FTPException("SSL Session for data connection was not reused");
				}
			}
		}
	}
	return ss;
}
```

micheleselea:
@obiltschnig I think the function now is working, WebSocket seems connect correctly, I'll do some more tests

micheleselea:
Hi @obiltschnig I end up with a "strange" behavior with some websockets on unreliable network.
Usually the old implementation when you did 
`int n=ws.receiveFrame()`
you can have 3 responses:
1)n=0 (socket closed)
2)n>0 data received
3)exception

with the new implementation I have a 4th reply
4)n=-1 without exception

I'm telling you because some code I use did not consider n=-1. I'm trying to check which is the real return error network code and let you know.

micheleselea:
Hi @obiltschnig I think it's the TIMEOUT (POCO_ETIMEDOUT 10060 for windows) that no more triggers exception, but return -1

micheleselea:
Another strange behavior I found with ssl connections, this time tcp ssl connections, I found that sometime the server side tcp ssl connection on receiveBytes call it returns me -1 (on windows 10 or 2012) without throw exception and specific error is
WSAECONNRESET 10054
I can't figure out what is trigger the problem but is frequent on some environment, I'm trying to debug

micheleselea:
I found that seems to be related to the SO_LINGER options. On windows the default seems to be false,0
when you call `getLinger(bool& on, int& seconds)` on a new accepted socket. I found that in some environment doing `setLinger(true,1)` can solve the WSAECONNRESET  problem even I can't figure out why, and I can't figure out why not always solve the problem. Probably I have to set in both client and server socket...I'll try
do you ever bump into this problem?

micheleselea:
Ok I setup a very small environment and I reproduced the problem, I'll write here in metacode

Thread A:
create a SecureServerSocket on port xxxx
accept connection
read from socket until receiveBytes returns >0

Thread B:
Create a StreamSocket connect to xxxx
than switch in secure socket with
Poco::Net::SecureStreamSocket sss(Poco::Net::SecureStreamSocket::attach(ss, Poco::Net::SSLManager::instance().defaultClientContext()));
ss=sss;
ss.sendByte (64000 bytes)
ss.sendByte (64000 bytes)
ss.sendByte (64000 bytes)
ss.sendByte (64000 bytes)
ss.sendByte (64000 bytes)
ss.close()

you will end up with server side receive some bytes than on client ss.close() he will receive -1 with partial data received

micheleselea:
Ok I probably found the problem and the solution.
I had to change the function
```
void SecureSocketImpl::shutdown()
{
	if (_pSSL)
	{
        // Don't shut down the socket more than once.
        int shutdownState = SSL_get_shutdown(_pSSL);
        bool shutdownSent = (shutdownState & SSL_SENT_SHUTDOWN) == SSL_SENT_SHUTDOWN;
        if (!shutdownSent)
        {
			// A proper clean shutdown would require us to
			// retry the shutdown if we get a zero return
			// value, until SSL_shutdown() returns 1.
			// However, this will lead to problems with
			// most web browsers, so we just set the shutdown
			// flag by calling SSL_shutdown() once and be
			// done with it.
			int rc = SSL_shutdown(_pSSL);
			if (rc < 0) handleError(rc);
			if (_pSocket->getBlocking())
			{
				_pSocket->shutdown();
			}
		}
	}
}
```

in

```
void SecureSocketImpl::shutdown()
{
	if (_pSSL)
	{
        // Don't shut down the socket more than once.
        int shutdownState = SSL_get_shutdown(_pSSL);
        bool shutdownSent = (shutdownState & SSL_SENT_SHUTDOWN) == SSL_SENT_SHUTDOWN;
        if (!shutdownSent)
        {
			// A proper clean shutdown would require us to
			// retry the shutdown if we get a zero return
			// value, until SSL_shutdown() returns 1.
			// However, this will lead to problems with
			// most web browsers, so we just set the shutdown
			// flag by calling SSL_shutdown() once and be
			// done with it.
			int rc = SSL_shutdown(_pSSL);
			if (rc == 0)
				rc = SSL_shutdown(_pSSL);
			if (rc < 0) handleError(rc);
			if (_pSocket->getBlocking())
			{
				_pSocket->shutdown();
			}
		}
	}
}
```

Otherwise the client does a brutal close of the socket. In this environment the problem is the client that send data, I don't know if can be even when a server send data, but probably yes

micheleselea:
Probably I'm going off topic here, but I think that the problem described is related to TLS1.3 session ticket and behavior of TCP stack when you have a pattern FTP like where che client connect to the server for sending data without receive nothing
https://github.com/openssl/openssl/issues/7948

micheleselea:
Ok so after reading about I write down some point:

1. The problem described is related to TLS1.3 with session tickets enabled
2. When we have one way TCP communication (like data socket of FTP where client connect send data and close without sending nothing) you can have data lost/truncation if the client send data and lose immediately after send
3. The problem seems to be about the ticket sent by the server and never read/processed by the client
4. A bidirectional SSL_shutdown() can fix the problem, but I don't know if it works in all the environment and if it's good for all connections
5. Disabling session ticket seems to fix the problem, but obviously you lost that TLS.13 feature
It just a specific case but it affects a lot of other software (like curl) that can do FTP or FTP data connection style 

I don't know which is the best way to deal with in poco lib, do you have suggestion? @obiltschnig 

CURL:
https://github.com/curl/curl/compare/master...jay:ossl_ftps_blocking_shutdown_test?expand=1
other link:
https://github.com/syslog-ng/syslog-ng/pull/3082

PR #3460, opened 2021-11-23
LoongArch support

loongson-zn:
Add LoongArch support, please review, thanks!

### DIFF
--- a/Foundation/include/Poco/Platform.h
+++ b/Foundation/include/Poco/Platform.h
@@ -136,6 +136,7 @@
 #define POCO_ARCH_ARM64   0x0f // same as POCO_ARCH_AARCH64
 #define POCO_ARCH_RISCV64 0x10
 #define POCO_ARCH_RISCV32 0x11
+#define POCO_ARCH_LOONGARCH64 0x12
 
 
 #if defined(__ALPHA) || defined(__alpha) || defined(__alpha__) || defined(_M_ALPHA)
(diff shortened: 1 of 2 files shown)


loongson-zn:
LoongArch is a RISC style ISA which is independently designed by Loongson Technology in China. Now we have a loongnix os base on debian 10.4, we hope this PR can be merged，thanks！
Documentations:
 - ISA:
   https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
 - ABI:
   https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
 - More docs can be found at:
   https://loongson.github.io/LoongArch-Documentation/README-EN.html

If you have any questions, please let me know.

aleks-f:
@loongson-zn thanks for the contribution
for future reference: please do not send pulls against master, but devel branch
also, the double conversion header was patched, but if you want that change to be safe from being overwritten in the future, you should send it upstream to the original library - we can't afford to keep track of such things

loongson-zn:
> @loongson-zn thanks for the contribution for future reference: please do not send pulls against master, but devel branch also, the double conversion header was patched, but if you want that change to be safe from being overwritten in the future, you should send it upstream to the original library - we can't afford to keep track of such things

Thanks @aleks-f , where isupstream to the original librar

loongson-zn:
> @loongson-zn thanks for the contribution for future reference: please do not send pulls against master, but devel branch also, the double conversion header was patched, but if you want that change to be safe from being overwritten in the future, you should send it upstream to the original library - we can't afford to keep track of such things

Thanks @aleks-f, there isn't poco's upstream?  Where is the upstream of the original poco library？

aleks-f:
I was referring to [double_conversion](https://github.com/google/double-conversion) header: [utils.h](https://github.com/pocoproject/poco/blob/master/Foundation/src/utils.h)

loongson-zn:
Thanks @aleks-f

PR #3462, opened 2021-11-24
improve performance for toLower/toUpper

dtiukalov:
applied loop unrolling for clang and MS compilers. Use new Ascii::toLower and Ascii::toUpper implementations with bitwise operations. Also, S::reserve and S::operator+ replaced to copy ctor (tested many times, and copy ctor works faster on average).


aleks-f (changes requested):
what are the compiler versions needed for these pragmas? also, don't compilers (at least clang, starting at -O3, I think) already do this at certain optimization level?

Issue #2569, opened 2018-12-18
sql timestamp


mediawiz61:
#### Steps to reproduce the problem
(please make this a [SSCCE](http://www.sscce.org/), if applicable and reasonable)

mediawiz61:
Has the issue of no support for the "Timestamp" data type in MySQL been resolved in the latest build???

PR #3471, opened 2021-12-08
#2569: Added TIMESTAMP data type support on MySQL

hectots:
Hi,

I noticed that the only thing that was missing for supporting TIMESTAMP columns on MySQL was to add the proper enum to a couple of switch statements. I also added a unit test to go with it. I hope this helps. I'm not sure if TIMESTAMP should get special treatment since, according to MySQL docs, it is virtually identical to DATETIME. The only differences are on the ranges and the way they work with time zones upon storage and retrieval. I'm new to this whole contribution to FOSS. Let me know if I'm breaking any guidelines.


obiltschnig:
Thank you, looks fine at first glance.

Issue #2755, opened 2019-07-24
MYSQL LONGTEXT


micheleselea:
#### Expected behavior
LONGTEXT data type correctly supported

#### Actual behavior
Using LONGTEXT cause the select statement throw exception in function 
` void assignRaw(const T* ptr, std::size_t count)
/// Assigns raw content to internal storage.
{
poco_assert_dbg (ptr);
LOB tmp(ptr, count);
swap(tmp);
}`
in LOB.h
just because the _metadata column description has 0 len. TEXT and MEDIUMTEXT is working

#### Steps to reproduce the problem
(please make this a [SSCCE](http://www.sscce.org/), if applicable and reasonable)

#### POCO version
latest 

#### Compiler and version
VS2015

#### Operating system and version
Windows 10 64bit

obiltschnig:
@aleks-f could you have a look at it please?

Issue #174, opened 2013-05-02
MySQL: 4GB allocated when reading any largetext or largeblob field


aleks-f:
From https://sourceforge.net/p/poco/bugs/608/

When a result set contains a largeblob or largetext field, mysql_fetch_fields() returns 2^32 as length of that field. ResultMetadata::init() adds this huge number when calculating the row buffer, and actually allocates this memory!!!
Proposed solution: When a 2^32 value of length is detected, assign a NULL buffer for that field. This will result in mysql_stmt_fetch() returning MYSQL_DATA_TRUNCATED which currently is handled as error. So, the error check should accept this return code as normal. Ideally, it should be checked which field was truncated and whether this field was assigned a NULL buffer. If not, then the data truncation is due to something else and should be reported as error. But currently this is not trivial because the StatementExecutor does not know about the ResultMetadata instance.
The actual length querying and buffer allocation is already implemented properly in Extractor.cpp, but that code was not executed the way it was meant.
As I had problems generating a diff (seems the indentation of my version was changed), I am pasting the code snippets here:

In ResultMetadata.cpp - 

```
ResultMetadata::init()
    for (size_t i = 0; i < count; i++)
    {
 +          unsigned long size = fieldSize(fields[i]);
 +          if (size == 0xFFFFFFFF)
 +             size = 0;

            _columns.push_back(MetaColumn(
                    i,                               // position
                    fields[i].name,                  // name
                    fieldType(fields[i]),            // type
 +                  size,                            // length
                    0,                               // TODO: precision (Now I dont know how to get it)
                    !IS_NOT_NULL(fields[i].flags)    // nullable
                    ));

            commonSize += size;
    }
.....
for (size_t i = 0; i < count; i++)
{
std::memset(&_row[i], 0, sizeof(MYSQL_BIND));
 +           unsigned int len = static_cast<unsigned int>(_columns[i].length());
            _row[i].buffer_type   = fields[i].type;
            _row[i].buffer_length = len;
 +          _row[i].buffer        = (len>0)?(&_buffer[0] + offset):NULL;
            _row[i].length        = &_lengths[i];
            _row[i].is_null       = &_isNull[i];
            _row[i].is_unsigned   = (my_bool) ((fields[i].flags & UNSIGNED_FLAG) == UNSIGNED_FLAG);
            offset += _row[i].buffer_length;
    }
```

In StatementExecutor.cpp, StatementExecutor::fetch()

```
[...]

PR #3474, opened 2021-12-09
#2755: Fix MySQL's LONGBLOB/LONGTEXT not allocating enough space

hectots:
Hi,

This is my take on a fix for bug #2755. I found that for LONGBLOB and LONGTEXT columns the size of the buffer was being set to zero. This was due to a fix for a memory leak reported on #174, where 4GB of space were being allocated for this buffers. At first I thought of just removing the condition that set the buffer size to zero. However, after reading the [MySQL doc's recommendations for fetching large values](https://dev.mysql.com/doc/c-api/8.0/en/mysql-stmt-fetch.html), I decided to implement a way to adjust the buffer sizes to fit the length of the data. At extraction time I would check if the size allocated for that column is zero. If it is then I will adjust the buffer size and fetch that column again. Hope this helps. Let me know if there is something I can improve or if there is a better solution.

### DIFF
--- a/Data/MySQL/include/Poco/Data/MySQL/Extractor.h
+++ b/Data/MySQL/include/Poco/Data/MySQL/Extractor.h
@@ -320,6 +320,8 @@ public:
 private:
 	bool realExtractFixed(std::size_t pos, enum_field_types type, void* buffer, bool isUnsigned = false);
 
+	bool extractLongLOB(std::size_t pos);
+
 	// Prevent VC8 warning "operator= could not be generated"
 	Extractor& operator=(const Extractor&);
 
--- a/Data/MySQL/include/Poco/Data/MySQL/ResultMetadata.h
+++ b/Data/MySQL/include/Poco/Data/MySQL/ResultMetadata.h
@@ -40,6 +40,9 @@ class ResultMetadata
 	/// MySQL result metadata
 {
 public:
+	~ResultMetadata();
+		/// Destroys the ResultMetadata.
+
 	void reset();
 		/// Resets the metadata.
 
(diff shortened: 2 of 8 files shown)


hmartinez82:
@obiltschnig I just saw this. This cover 80% of the work in my other PR (for #3481) that was I about to create. Because the buffer being set to 0 was indeed getting in the way . @hectots  Good job with the extra call to `fetchColumn`, that was also the route I took. I'll rebase my changes on top of this branch.

hmartinez82:
Just for completion. The unit test is failing with an internal MySQL assertion.
```
testBLOBStmt:
testLongBLOB: Assertion failed: param->buffer_length != 0, file G:\ade\build\sb_0-24199034-1500304113.07\mysql-connector-c-6.1.11-src\libmysql\libmysql.c, line 4069
```
Stack trace:
```
PocoDataMySQLd.dll!setup_one_fetch_function(st_mysql_bind * param, st_mysql_field * field) Line 4069
	at g:\ade\build\sb_0-24199034-1500304113.07\mysql-connector-c-6.1.11-src\libmysql\libmysql.c(4069)
PocoDataMySQLd.dll!mysql_stmt_bind_result(st_mysql_stmt * stmt, st_mysql_bind * my_bind) Line 4217
	at g:\ade\build\sb_0-24199034-1500304113.07\mysql-connector-c-6.1.11-src\libmysql\libmysql.c(4217)
PocoDataMySQLd.dll!Poco::Data::MySQL::StatementExecutor::bindResult(MYSQL_BIND * result) Line 93
	at D:\Dev\Github\poco\Data\MySQL\src\StatementExecutor.cpp(93)
PocoDataMySQLd.dll!Poco::Data::MySQL::MySQLStatementImpl::compileImpl() Line 129
	at D:\Dev\Github\poco\Data\MySQL\src\MySQLStatementImpl.cpp(129)
PocoDatad.dll!Poco::Data::StatementImpl::compile() Line 185
	at D:\Dev\Github\poco\Data\src\StatementImpl.cpp(185)
PocoDatad.dll!Poco::Data::StatementImpl::execute(const bool & reset) Line 87
	at D:\Dev\Github\poco\Data\src\StatementImpl.cpp(87)
PocoDatad.dll!Poco::Data::Statement::execute(bool reset) Line 128
	at D:\Dev\Github\poco\Data\src\Statement.cpp(128)
DataMySQL-testrunner.exe!SQLExecutor::longBlob() Line 1434
	at D:\Dev\Github\poco\Data\MySQL\testsuite\src\SQLExecutor.cpp(1434)
```
I'm almost confident this only happens in debug builds of the library (https://github.com/mysql/mysql-server/blob/8.0/libmysql/libmysql.cc#L3525) which means we can't run these unit tests with the debug libraries

hmartinez82:
But otherwise, all good.

PR #3476, opened 2021-12-13
add separate accessors and mutators for connect, send and receive tim…

fosterbt:
…eouts


fosterbt:
I work on several large projects that use Poco. Anyway, these are mostly for convenience, as several places in the code like to change these timouts without having to access the other values and use the setTimeout method.  Mainly I'm just tired of porting this change every time we update Poco. :)

aleks-f:
@fosterbt this requires some code style corrections - function definitions should go separate in the header below the class declaration

fosterbt:
I've made the requested changes. Thanks and have a great day.

PR #3491, opened 2021-12-29
Extract JSON type as string

### LINKED ISSUE #3481: JSON DataType support for MySQL
We have a table in our DB , which has a column with Datatype JSON for MySQL. I am not able to Query this particular column using Poco::DataMySQL library. 
Following exception is thrown "MySQL: [MySQL]: unknown field type"

Does Poco::DataMysql not support JSON Data type?
If yes , could you please suggest a workaround for this.

we are using this version of software:
https://github.com/pocoproject/poco/archive/refs/tags/poco-1.11.0-release.zip

Cheers!

hmartinez82:
Fixes #3481 

Requires #3474 being merged first

### DIFF
--- a/Data/MySQL/include/Poco/Data/MySQL/Extractor.h
+++ b/Data/MySQL/include/Poco/Data/MySQL/Extractor.h
@@ -325,6 +325,8 @@ private:
 
 	bool extractLongLOB(std::size_t pos);
 
+	bool extractJSON(std::size_t pos);
+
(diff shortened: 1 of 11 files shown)


aleks-f:
@hmartinez82 when will you rebase? If you want it in 1.12, we plan to release next week

hmartinez82:
@aleks-f  Done . Sorry for the delay. I was traveling, but once I got home I forgot about. And then it took me a while to rebuild and re-run the unit tests. All passing now

PR #3500, opened 2022-01-12
Sandbox all iFrames in PocoDoc

josharenson:
While these pages are typically served locally, this is a best security
practice.


josharenson:
This is a nitpicky patch but brings Poco into compliance with my organization's security policies.

Issue #3507, opened 2022-01-31
Reference counting for bound configuration in Util::Option is broken


kblaschke:
**Issue:**
In commit 1bf40a0, the type of the `_pConfig` member was changed from a raw pointer to `Poco::AutoPtr`, but the `duplicate()`/ `release()` calls in the implementation were not adapted accordingly. In most cases, the usage counter is now incremented twice, but not in a call to `binding()`, because the function still takes a raw pointer and assignment of a raw pointer to AutoPtr does not increment the counter. Since the Option destructor will call `release()` twice (once manually, once in the AutoPtr destructor), counter is decremented by 2.

If more than one option is bound to the same configuration object, this will result in an early free of the associated AbstractConfiguration object, and subsequent destructors will dereference a deleted pointer and cause a crash.

**How to reproduce:**
- Create a class based on `Poco::Util::Application`.
- Add a member of type `Poco::Util::MapConfiguration::Ptr` to the app and initialize it.
- Define two options in `defineOptions()` bound to the configuration above.
- Run the application.

**Expected result:**
The application should terminate without error.

**Actual result:**
Depending on the compiler and OS, the application will crash with a C library error or segmentation fault. It's not totally reliable though, depending on memory layout.

Debugging with a breakpoint in `Poco::Util::Option::~Option()` and watching pConfig will always show the behaviour.

obiltschnig:
Fixed slightly different in [010cf8b](https://github.com/pocoproject/poco/commit/010cf8b7d0a33694168af3cd85354720622de440) by changing the `binding()` argument type to an `AbstractConfiguration::Ptr`, which relieves us from manual reference counting.

kblaschke:
Was my first idea too, but was not sure about introducing a potentially breaking API/ABI change. Glad to see this is an option!

obiltschnig:
(Potentially) breaking API changes are always a concern, but in this case it seems like the better option overall, as it simplifies things, is consistent with the rest of the Configuration APIs and should not break any existing code. Furthermore, it should have been changed back in 1.10.0, when the Configuration APIs have been updated to use `Poco::AutoPtr`.

stuaxo:
This seems to be the version of poco I have with ubuntu, is there an upstream bug to get this version bumped there ?

kblaschke:
> This seems to be the version of poco I have with ubuntu, is there an upstream bug to get this version bumped there ?

[...]

PR #3508, opened 2022-01-31
#3507: Fix bound configuration reference counting in Poco::Util::Option.

kblaschke:
Type of _pConfig was changed from raw pointer to AutoPtr in commit 1bf40a0, but reference counting was not properly adapted. In most instances, refcount was incremented by 2, but not in "binding()". In the destructor, refcount was always decremented twice, leading to early deletion of the bound config and later possibly an invalid pointer access.

### DIFF
--- a/Util/src/Option.cpp
+++ b/Util/src/Option.cpp
@@ -54,7 +54,6 @@ Option::Option(const Option& option):
 {
 	if (_pValidator) _pValidator->duplicate();
 	if (_pCallback) _pCallback = _pCallback->clone();
-	if (_pConfig) _pConfig->duplicate();
 }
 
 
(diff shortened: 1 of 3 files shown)


kblaschke:
Kept the `duplicate()` call in `binding()`, so the `_pConfig` refcount is always correctly incremented by 1 on all assignments.

kblaschke:
Amended commit to remove the line `if (_pConfig) _pConfig->release();` in `binding()`, as `AutoPtr` will do that for us (but not calling `duplicate()`). Would otherwise cause the same issue if `binding()` is called multiple times on an Option.

obiltschnig:
Fixed in [010cf8b](https://github.com/pocoproject/poco/commit/010cf8b7d0a33694168af3cd85354720622de440)

PR #3509, opened 2022-01-31
fix dst and utcOffset handling for Dublin time zone

radimdvorak:


radimdvorak:
The pull request is fixing issue #3510 - LocalDateTime returns wrong tzd for Ireland/Dublin time zone.

andy9a9:
Closes #3519

andy9a9:
Closes #3510

PR #3519, opened 2022-02-13
Add convertation to string in Redis Command#set

102Rudy:
There is a problem in `Poco::Redis::Command::set(...)`

If I want to set expire time for the value, I can't just write:
```cpp
Poco::Redis::Command::set(key, value, true, Poco::Timespan(60, 0));
```
That doesn't work because of incorrect implementation of the `set` function: when `expireTime` is added to `cmd` it's not converted from `Int64` to `std::string`.

This PR fixes the issue.


102Rudy:
@obiltschnig Hey! Will we merge that?

PR #3528, opened 2022-02-19
Added LocalConfigurationView 

KevDi:
With the `LocalConfigurationView` it should be possible to only search inside the Viewed Scope of the Configuration.
With the `ConfigurationView` at the moment there are some Problems possible when looking for Properties in a Viewed Scope and also the outer Scope of the Configuration is searched.


KevDi:
I'm not quite sure why the github actions failed. The `cmake` ones and also the windows one just completes fine but the `make` ones failed

KevDi:
Accidently does a rebase i will reopen the PR again only with the changes i made

PR #3529, opened 2022-02-19
Added LocalConfigurationView to only search inside the viewed scope

KevDi:
`LocalConfigurationView` only searches in the viewed scope. The current existing `ConfigurationView` causes sometimes errors when properties are also declared in the outside scope.


aleks-f:
@KevDi please, for future contributions, follow the [coding style](https://www.appinf.com/download/CppCodingStyleGuide.pdf) - it is the first item listed in [Code Contributions](url)

PR #3552, opened 2022-03-30
Handle OpenSSL bidirectional shutdown

micheleselea:
Handle bidirectional SSL shutdown when a FTPS data like client that never read from socket has to close connection. If it shutdown properly you can end up with data truncation problem server side


aleks-f:
fixed in https://github.com/pocoproject/poco/commit/3fb001f39742c6cf7659f679bb8bbe51403fdb99

PR #3567, opened 2022-04-19
fix(openssl-initializer): check legacy provider existence for legacy exception

petermarko:
This fixes copy+paste error in openssl3 implementation.
This condition could never be true because default provider failure would throw couple lines above this code.

Note that this fix implies that the library is installed.
I'd suggest to include legacy provider optionally in future version of poco.

### DIFF
--- a/Crypto/src/OpenSSLInitializer.cpp
+++ b/Crypto/src/OpenSSLInitializer.cpp
@@ -137,7 +137,7 @@ void OpenSSLInitializer::initialize()
 		if (!_legacyProvider)
 		{
 			_legacyProvider  = OSSL_PROVIDER_load(NULL, "legacy");
-			if (!_defaultProvider) throw CryptoException("Failed to load OpenSSL legacy provider");
+			if (!_legacyProvider) throw CryptoException("Failed to load OpenSSL legacy provider");
 		}
 #endif
 	}


obiltschnig:
Thinking about removing the check altogether, so that absence of the legacy provider does not lead to an exception.

petermarko:
Just let me know when you decide and I'll update this PR.
I guess then I would also need to add an inline method to check if it is loaded...

aleks-f:
[3b3747e](https://github.com/pocoproject/poco/commit/3b3747eb405eb459217963411c10f8751336d1bd) fixes leak reported by valgrind due to not unloading providers.

[484eed7](https://github.com/pocoproject/poco/commit/484eed7e9b32cd9bd45d7b965ebe79676a1e9509) fixes race on unloading

There should probably be the load-side race-prevention equivalent, to prevent multiple calls to `OSSL_PROVIDER_load`, but I'm not sure if it's doable with atomic only - maybe there is a need for mutex here, after all?

obiltschnig:
Unloading providers is problematic as OpenSSL also unloads them when it shuts down, unless you completely switch OpenSSL to fully manual initialization. However, this will break interoperability with other libraries that also need OpenSSL. That's the reason why I removed unloading providers before the 1.11.2 release.

obiltschnig:
Also, if you manually initialize and uninitialize OpenSSL, you can only do it once. Re-initializing OpenSSL after it has already been uninitialized once will not work. This is stated in the OpenSSL documentation.

aleks-f:
In regards to providers load/unload, there are leaks reported when unload is not called, so OpenSSL 3 is definitely not unloading them, at least not in our implementation and on the platforms I tested; I can't find mention of it in the [documentation](https://www.openssl.org/docs/man3.0/man3/OSSL_PROVIDER_load.html), either. 

As for initializing after uninitialize, I agree with that - maybe another flag there.

Or remove all the unloading I added, and ignore the leaks.

aleks-f:
ok, I found it [here](https://github.com/openssl/openssl/issues/15915), I guess leaks were reported because valgrind did not catch the atexit openssl cleanup.

aleks-f:
@obiltschnig I have removed the cleanup in 1.11.3 branch, but on a more detailed read of the [openssl github issue](https://github.com/openssl/openssl/issues/15915#issuecomment-868384876), those provider unloads should actually be called, and the leaks I'm seeing when they are not are legitimate:

"if the application owns references to the OSSL_PROVIDER and those references never get freed then the ref count never gets to 0 and a memleak will occur"

I guess not critical, since it happens on shutdown, but it's still nice to have a clean leak report

petermarko:
Is there anything I can do here? I don't see any decision in the discussion above...

aleks-f:
@petermarko your fix from this pull is in, so nothing to worry about there.

As for unloading the loaded providers, we currently still have a disagreement there - my intertpretation of [this](https://github.com/openssl/openssl/issues/15915#issuecomment-868384876) is that all providers explicitly loaded by the library user should also be explicitly unloaded

obiltschnig:
The above means we'll have to change Crypto/NetSSL to maintain its own `OSSL_LIB_CTX`.

obiltschnig:
Closing this for now.

PR #952, opened 2015-09-28
Named substring support

david-moran:
Hello,

Regular expressions are very limited in C++11 using ECMAScript syntax. POCO is more powerful using PCRE but it has not support for substrings names.

See man pages:
pcre_get_named_substring
pcre_fullinfo

I have added a field in Match structure to support group name when calling match() method with a MatchVec parameter:

# Code example:
```c++
#include <iostream>
#include <string>
#include <cstring>
#include <cstdlib>
#include <Poco/RegularExpression.h>

using namespace std;
using namespace Poco;

int main(int argc, char* argv[])
{
	if (argc != 3)
		return -1;

	RegularExpression::MatchVec matches;
	RegularExpression regex(argv[1]);

	regex.match(argv[2], 0, matches);
	if (matches.size() > 0)
	{
		cout << "String found" << endl;

		for (auto match : matches)
		{
			char* matchedStr = strndup(argv[2] + match.offset, match.length);
			cout << "    Group name: " << match.name << " => " << matchedStr << endl;
			free(matchedStr);
		}
	}
	else
	{
		cout << "String not found" << endl;
	}

	return 0;
}
```
## Execution:
```
david@ubuntu:~/workspace/test$ g++ -o regexp -std=c++11 regexp.cpp -lPocoFoundation
david@ubuntu:~/workspace/test$ ./regexp "Hello (?P<name>\S+) (?P<surname>\S+)" "Hello David Moran"
String found
    Group name:  => Hello David Moran
    Group name: name => David
    Group name: surname => Moran
david@ubuntu:~/workspace/test$ 
```

Regards

PR #3569, opened 2022-04-21
Readded named substring support for regexes

ccarrasc:
Original merged PR #952 was merged in 2015, but has gone missing. This is adding it back in.


aleks-f:
yeah, releases are not coming from devel branches anymore - please resubmit this to the [poco-1.12.0](https://github.com/pocoproject/poco/tree/poco-1.12.0) branch and it will be in the next release

Issue #3486, opened 2021-12-22
Windows on ARM64 support


hmartinez82:
Windows has been available for ARM64 on laptops since the end of 2018, but the Surface Pro X released in the end of 2019 made it even more popular

Right now Poco fails to build even the basic components, like Foundation with the `amd64_arm64` environment from Visual Studio 2019 or 2022.

PR #3606, opened 2022-05-26
Add ARM64 to ProGen for vs170

hmartinez82:
Implements #3486


ProGen/src/ProGen.cpp
```diff
@@ -585,6 +585,17 @@ class ProGenApp: public Application
...
+		for (unsigned long i = 0; i < pLinkList->length(); i++)
+		{
+			Poco::XML::Element* pLinkElem = static_cast<Poco::XML::Element*>(pLinkList->item(i));
+			Poco::XML::Element* pItemDefinitionGroupElem = static_cast<Poco::XML::Element*>(pLinkElem->parentNode());
+			Poco::XML::XMLString condition = pItemDefinitionGroupElem->getAttribute("Condition");
+			if (Poco::endsWith(condition, Poco::XML::XMLString("ARM64'")))
+			{
+				appendElement(pLinkElem, "TargetMachine", "MachineARM64");
+			}
+		}
```
hmartinez82, line 598: This was required because, although the templates used `TargetMachine="18"`, the `DevEnv /update` call is not honoring that value and leaving an empty `TargetMachine` in the created `.vcxproj`

PR #3617, opened 2022-05-30
Fix/tsan

aleks-f:
fixes for tsan warnings

- [x] Foundation (`ActiveDispatcherTest` disabled)
- [x] Net
- [x] JSON
- [x] XML
- [x] Util
- [x] Crypto
- [x] NetSSL_OpenSSL
- [x] Data
- [x] Data/SQLite
- [x] JWT
- [x] Prometheus
- [x] Zip


aleks-f:
@obiltschnig please look at `Thread_POSIX` changes in https://github.com/pocoproject/poco/pull/3617/commits/4426f4708c7bd7ae74bba5e6d2ba72729e8a82fc

the rest of it is i think quite straightforward (except for that one warning that I don't know what to do about)

aleks-f:
FWIW, all `ActiveDispatcher` tsan warnings stem from the fact that the parent object `run()` is started in a tread from the parent constructor (ie. before the inheriting class is fully constructed); since run truly only starts executing once there are queued notifications, this is not a legitimate issue; unfortunately, tsan does not understand the semantics behind it and issues a warning, and for some reason one of those warnings only shows in release run. the solution will likely be not to run that test in release tsan build

Issue #3591, opened 2022-05-11
` format ` function is called recursively in file `Format.h`  when enable  msvc/c++ 20


shelllet:
test code:
```

#include <iostream>
#include <vector>
#include <any>


void format(std::string& result, const std::string& fmt, const std::vector<std::any>& values)
{
	std::cout << "invoked." << std::endl;
}


template <
	typename T,
	typename... Args>
void format(std::string& result, const char* fmt, T arg1, Args... args)
/// Appends the formatted string to result.
{
	std::vector<std::any> values;
	values.reserve(sizeof...(Args) + 1);
	values.emplace_back(arg1);
	values.insert(values.end(), { args... });

	format(result, fmt, values); // warning C4717: 'format<std::vector<std::any,std::allocator<std::any> > >': recursive on all control paths, function will cause runtime stack overflow

}


template <
	typename T,
	typename... Args>
std::string format(const char* fmt, T arg1, Args... args)
/// Returns the formatted string.
{
	std::vector<std::any> values;
	values.reserve(sizeof...(Args) + 1);
	values.emplace_back(arg1);
	values.insert(values.end(), { args... });
	std::string result;
	::format(result, fmt, values);
	return result;
}

int main()
{
	std::string connector = "sqlite", connectionString="c:\\sqlite.db";

	format("%s:///%s", connector, connectionString);
}

```

shelllet:
`https://github.com/pocoproject/poco/blob/master/Foundation/include/Poco/Format.h`

```
template <
	typename T,
	typename... Args>
void format(std::string& result, const std::string& fmt, T arg1, Args... args) // why add additional  template param arg1?
```

aleks-f:
https://github.com/pocoproject/poco/pull/3640

PR #3625, opened 2022-06-02
Update Format.h

shelllet:
#3591
fix warning C4717: 'format<std::vector<std::any,std::allocator<std::any> > >': recursive on all control paths, function will cause runtime stack overflow


shelllet:
`std::remove_cvref_t` defined since C++20

PR #3626, opened 2022-06-02
Update double-conversion to 3.2.0 

Spaky:
* from https://github.com/google/double-conversion/tree/v3.2.0
* adopted call of std::min, std::max from original code
* reapplied the fixes for riscv and nios
* build for win32/x64 and linux
* tests okay


aleks-f:
thanks, but this is already done in https://github.com/pocoproject/poco/pull/3617/commits/15a4c60062213762605eef1df316394a1a02ca11

PR #3640, opened 2022-06-18
fix warning C4717: 'format<std::vector<std::any,std::allocatorstd::an…

shelllet:
…y > >': recursive on all control paths, function will cause runtime stack overflow

### DIFF
--- a/Foundation/include/Poco/Format.h
+++ b/Foundation/include/Poco/Format.h
@@ -27,7 +27,7 @@
 namespace Poco {
 
 
-std::string Foundation_API format(const std::string& fmt, const Any& value);
+	std::string Foundation_API format(const std::string& fmt, const Any& value);
 	/// This function implements sprintf-style formatting in a typesafe way.
 	/// Various variants of the function are available, supporting a
 	/// different number of arguments (up to six).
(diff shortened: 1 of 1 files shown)


Foundation/include/Poco/Format.h
```diff
@@ -27,7 +27,7 @@
 namespace Poco {
 
 
-std::string Foundation_API format(const std::string& fmt, const Any& value);
+	std::string Foundation_API format(const std::string& fmt, const Any& value);
```
aleks-f, line 30: why this indentation?

aleks-f (changes requested):
please revert the indentation

Foundation/include/Poco/Format.h
shelllet: this is the result of vscode auto-formatting

aleks-f (approved):

aleks-f:
this breaks g++, will be reverted and redone
https://github.com/pocoproject/poco/pull/3732

shelllet:
I don't have any experience with g++, anyone has an idea?

aleks-f:
There's no need for any more ideas regarding this, it has bern taken care of in 1.12.2 release

PR #3641, opened 2022-06-20
FifoBuffer.advance method not throw exception when length==0

JackyWoo:
FifoBuffer.advance method not throw exception when length==0 and buffer is full.

### DIFF
--- a/Foundation/include/Poco/FIFOBuffer.h
+++ b/Foundation/include/Poco/FIFOBuffer.h
@@ -334,7 +334,7 @@ public:
 		/// Should be called AFTER the data 
 		/// was copied into the buffer.
 	{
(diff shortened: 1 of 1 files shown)


Foundation/include/Poco/FIFOBuffer.h
```diff
@@ -334,6 +334,7 @@ class BasicFIFOBuffer
 		/// Should be called AFTER the data 
 		/// was copied into the buffer.
 	{
+        if (0 == length) return;
```
aleks-f, line 337: why not use tabs and indent this like the rest of the code?

aleks-f (changes requested):

aleks-f (approved):

PR #3642, opened 2022-06-24
Make ParallelSocketReactor thread namable

JackyWoo:
Make ParallelSocketReactor thread namable. It is helpful for debuging.


aleks-f (changes requested):
please pay attention to the code style - we use tabs, not spaces indentation

PR #3644, opened 2022-06-24
Make poll set interruptable

JackyWoo:
Make poll set interruptable by a pipe. When wakeUp invoked write a char to pipe.

### DIFF
--- a/Net/src/PollSet.cpp
+++ b/Net/src/PollSet.cpp
@@ -24,6 +24,7 @@
 #elif defined(POCO_HAVE_FD_POLL)
 	#ifndef _WIN32
 		#include <poll.h>
+		#include "Poco/Pipe.h"
 	#endif
 #endif
 
@@ -226,6 +227,17 @@ private:
 class PollSetImpl
 {
 public:
+	PollSetImpl()
+		{
+		pollfd fd{_pipe.readHandle(), POLLIN, 0};
+		_pollfds[0] = fd;
+	}
+
+	~PollSetImpl()
+	{
+		_pipe.close();
+	}
+
 	void add(const Socket& socket, int mode)
 	{
 		Poco::FastMutex::ScopedLock lock(_mutex);
@@ -280,7 +292,7 @@ public:
 		_socketMap.clear();
 		_addMap.clear();
 		_removeSet.clear();
-		_pollfds.clear();
+		_pollfds.reserve(1);
 	}
 
 	PollSet::SocketModeMap poll(const Poco::Timespan& timeout)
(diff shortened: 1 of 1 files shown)


Net/src/PollSet.cpp
```diff
@@ -226,6 +227,17 @@ class PollSetImpl
 class PollSetImpl
 {
 public:
+	PollSetImpl()
+		{
```
aleks-f, line 231: double indent

aleks-f (changes requested):
testcase must be enabled and pass:
https://github.com/pocoproject/poco/blob/cbe738d5e8390881d7f2de15e0073b0ab0caa290/Net/testsuite/src/PollSetTest.cpp#L305-L307

aleks-f:
mac test segfaults

JackyWoo:
@aleks-f  There is a CI error, perhaps not cost by this pr, can you take a look at it?

Net/src/PollSet.cpp
JackyWoo: I reconfigured my IDE with tap as indent. Thanks you.

aleks-f:
> @aleks-f There is a CI error, perhaps not caused by this pr, can you take a look at it?

It is not, but you need to look better at what you're doing here; there are essentially 4 implementations: epoll, poll, `select` with multiple `fd_set`s and the fallback of ordinary `select`, with poll implementation not used by default on any platform. I don't mind you adding interruptibility to poll implementation, but the tests must reflect the fact that select implementations are not interruptible.

JackyWoo:
Got it, thx

JackyWoo:
gcc build error ：maybe caused by abi version compatibility . 
 https://stackoverflow.com/questions/48149323/what-does-the-gcc-warning-project-parameter-passing-for-x-changed-in-gcc-7-1-m
```
** Compiling src/Format.cpp (debug, shared)
In file included from src/Format.cpp:15:
include/Poco/Format.h: In function ‘void Poco::format(std::string&, const string&, T, Args ...) [with T = Poco::Any; Args = {}]’:
include/Poco/Format.h:129:2: note: parameter passing for argument of type ‘std::vector<Poco::Any>::const_iterator’ {aka ‘__gnu_cxx::__normal_iterator<const Poco::Any*, std::vector<Poco::Any> >’} changed in GCC 7.1
  129 |  values.insert(values.end(), { args... });
      |  ^~~~~~
In file included from /usr/arm-linux-gnueabihf/include/c++/9/vector:72,
                 from include/Poco/Format.h:23,
                 from src/Format.cpp:15:
/usr/arm-linux-gnueabihf/include/c++/9/bits/vector.tcc: In member function ‘void std::vector<_Tp, _Alloc>::emplace_back(_Args&& ...) [with _Args = {Poco::Any&}; _Tp = Poco::Any; _Alloc = std::allocator<Poco::Any>]’:
```

PR #3657, opened 2022-07-01
Add lower case format for `NumberFormatter`

JackyWoo:
Add lower case format for `NumberFormatter`

### DIFF
--- a/Foundation/include/Poco/NumberFormatter.h
+++ b/Foundation/include/Poco/NumberFormatter.h
@@ -59,13 +59,13 @@ public:
 		/// right justified and zero-padded in a field
 		/// having at least the specified width.
 
-	static std::string formatHex(int value, bool prefix = false);
+	static std::string formatHex(int value, bool prefix = false, bool upper = true);
 		/// Formats an int value in hexadecimal notation.
 		/// If prefix is true, "0x" prefix is prepended to the
 		/// resulting string.
 		/// The value is treated as unsigned.
 
-	static std::string formatHex(int value, int width, bool prefix = false);
+	static std::string formatHex(int value, int width, bool prefix = false, bool upper = true);
 		/// Formats a int value in hexadecimal notation,
 		/// right justified and zero-padded in
 		/// a field having at least the specified width.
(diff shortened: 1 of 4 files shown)


obiltschnig:
Instead of multiple bool arguments we should have an option argument (int) accepting a bitmask, e.g.:
```
enum Options
{
    OPT_PREFIX = 1,
    OPT_UPPERCASE = 2
};

static std::string formatHex(int value, int options = 0);
```

JackyWoo:
@obiltschnig It isa good fashion, but I wonder maybe it will cause compatibility issue.

aleks-f:
@JackyWoo no compatibility issues - `bool` is guaranteed to be `1`, so `formatHex(val, true)` will work exactly as before. Can you please modify and send another pull?

aleks-f:
@JackyWoo please focus on fixing this first, we want to release it in 1.12. The thread name is scheduled for 1.13

aleks-f:
It is not a drop-in replacement because it introduces ambiguity. I'll revert this for now
```src/Logger.cpp: In static member function ‘static void Poco::Logger::formatDump(std::__cxx11::string&, const void*, std::size_t)’:
src/Logger.cpp:252:52: error: call of overloaded ‘formatHex(int&, int)’ is ambiguous
   message.append(NumberFormatter::formatHex(addr, 4));
                                                    ^
In file included from src/Logger.cpp:19:
include/Poco/NumberFormatter.h:541:20: note: candidate: ‘static std::__cxx11::string Poco::NumberFormatter::formatHex(int, int)’
 inline std::string NumberFormatter::formatHex(int value, int options)
                    ^~~~~~~~~~~~~~~
include/Poco/NumberFormatter.h:549:20: note: candidate: ‘static std::__cxx11::string Poco::NumberFormatter::formatHex(int, int, int)’
 inline std::string NumberFormatter::formatHex(int value, int width, int options)

**Comment by obiltschnig** [comment]
We need to keep the old methods taking a `bool` for backwards compatibility, but mark them as deprecated.

**Comment by JackyWoo** [comment]
> @JackyWoo please focus on fixing this first, we want to release it in 1.12. The thread name is scheduled for 1.13

@aleks-f sorry for missing your message.  

I tried as @obiltschnig menthioned but found ambiguous issue
```
static std::string formatHex(int value, int options = 0);
static std::string formatHex(int value, int width, bool prefix = false);
```

So I suggest:

1. the fashion in this pr
2. add a new method such as `formatHexNew`


what is your opinion？

/cc @obiltschnig

Issue #3678, opened 2022-07-11
Make socket reactor more easy to use


JackyWoo:
I use `SocketReactor` in Poco.  And I found that user must do resource management in `UDF Handler` which can be improved by handling it in `SocketReactor`.

Resource management means:
1. remove all  events in `NObserver`
2. delete `UDF Handler`

In my case there is a uncaught error in my `UDF Handler` and it lead to 100% CPU usage.

Possible fix:

1. Clear  events `SocketReactor::dispatch` reveive error.
2. Create a new `NObserver` whose `_pObject`  is a `RefCountedObject`

This change may lead to compatibility issue.

JackyWoo:
SSCCE ：

Use `Net/samples/EchoServer/src/EchoServer.cpp` just modify `onSocketWritable` which throw something.

```
        void onSocketWritable(const AutoPtr<WritableNotification>& pNf)
	{
		try
		{
			_socket.sendBytes(_fifoOut);
			throw 1;
		}
		catch (Poco::Exception& exc)
		{
			Application& app = Application::instance();
			app.logger().log(exc);
			delete this;
		}
	}
```

JackyWoo:
@aleks-f How about the idea `Make socket reactor more easy to use`? Expecting for reply.

aleks-f:
It's quite a big diff, did not have time to look at it yet. There are conflicts in the pull now, fix and I'll review as soon as I get a chance.

JackyWoo:
@aleks-f Conflicts are fixed. Pls review when you have a change.

PR #3685, opened 2022-07-14
Make socket reactor more easy to use.

JackyWoo:
This pr try to fix issue #3678 . If the idea is ok I will fix the tests.


JackyWoo:
close it temporarily

Issue #2976, opened 2020-04-10
SharedMemoryImpl x64 size error


trublemaker:
i use SharedMemory with x64,
while i set size > 4G , then error,
        std::size_t size = 1024LL * 1024 * 1024 * 4;
	SharedMemory mem("MySharedMemory", size, SharedMemory::AM_WRITE); 

and i find the define of _size is DWORD, while size_t is 
#ifdef _WIN64
    typedef unsigned __int64 size_t;
....
and _size(static_cast<DWORD>(size)) may lose high

trublemaker:
and 	  
	_memHandle = CreateFileMappingW(INVALID_HANDLE_VALUE, NULL, _mode, 0, _size, utf16name.c_str());

dwMaximumSizeHigh always set 0, and   it should  be _size>>32

Spaky:
Hey @aleks-f,
i created an improvement that allows SharedMemory greater than 4GB on x64 for win32.
Currently the win32 jobs on GitHub are failing because the paging file is too small,
but on my local machine the new test is successful.

Do you think it is okay, when i create a PR for this issue?

pavledragisic:
@aleks-f  This commit https://github.com/pocoproject/poco/commit/df656fc51d254ffe83781d713d5875b99d4aeeee ended up here by mistake. It was supposed to go to the issue [2978](https://github.com/pocoproject/poco/issues/2978)

aleks-f:
> @aleks-f This commit [df656fc](https://github.com/pocoproject/poco/commit/df656fc51d254ffe83781d713d5875b99d4aeeee) ended up here by mistake. It was supposed to go to the issue [2978](https://github.com/pocoproject/poco/issues/2978)

@pavledragisic you should [--amend](https://www.atlassian.com/git/tutorials/rewriting-history) it

PR #3707, opened 2022-07-21
SharedMemoryImpl x64 size error #2976

Spaky:
* allow more then 4GB on x64 Windows systems
* added new test for large SharedMemory objects


aleks-f (changes requested):
Well, in order to merge, it first has to pass CI (I have some intuitive doubts about changing `DWORD` to `size_t`).
And then, please clean this pull up - rebase or create a new, clean pull - it won't be merged before all these unnecessary and duplicate commits are cleaned up.

PR #3731, opened 2022-08-03
Changes so app which uses c++20 can compile and Poco works.

sildeag:
Format.h created many compile errors which are resolved, at least on Ubuntu 22.04 with g++ 11.2, and the app works.  The changes include recognizing that Any was in the Poco namespace.  Any.h was changed because #define was out of scope.  A typehandler pointer was added.  And cmake find for ODBC did not function on Ubuntu.

### DIFF
--- a/Data/include/Poco/Data/Binding.h
+++ b/Data/include/Poco/Data/Binding.h
@@ -1387,7 +1387,7 @@ inline AbstractBinding::Ptr use(T& t, const std::string& name = "")
 	// This can be resolved by either (a) using bind (which will copy the value),
 	// or (b) if the const ref is guaranteed to exist when execute is called
 	// (which can be much later!), by using the "useRef" keyword instead
-	poco_static_assert (!IsConst<T>::VALUE);
+	//poco_static_assert (!IsConst<T>::VALUE);
 	return new Binding<T>(t, name, AbstractBinding::PD_IN);
 }
 
--- a/Data/include/Poco/Data/TypeHandler.h
+++ b/Data/include/Poco/Data/TypeHandler.h
@@ -5674,6 +5674,7 @@ class TypeHandler<Poco::AutoPtr<T>>: public AbstractTypeHandler
 	/// Specialization of type handler for Poco::AutoPtr
 {
 public:
+        typedef AutoPtr<T> TPtr; // new code here
 	static void bind(std::size_t pos, const Poco::AutoPtr<T>& obj, AbstractBinder::Ptr pBinder, AbstractBinder::Direction dir)
 	{
 		// *obj will trigger a nullpointer exception if empty: this is on purpose
--- a/Foundation/include/Poco/Any.h
+++ b/Foundation/include/Poco/Any.h
@@ -28,6 +28,8 @@
 
 namespace Poco {
 
+#define POCO_SMALL_OBJECT_SIZE 64
+
 class Any;
 
 namespace Dynamic {
--- a/Foundation/include/Poco/Format.h
+++ b/Foundation/include/Poco/Format.h
@@ -22,12 +22,15 @@
 #include "Poco/Any.h"
 #include <vector>
 #include <type_traits>
-
+#include <charconv>
+#include <string>
+#include <iostream>
 
 namespace Poco {
 
+//using Poco::Any;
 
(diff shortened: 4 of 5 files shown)


Data/include/Poco/Data/TypeHandler.h
```diff
@@ -5674,6 +5674,7 @@ class TypeHandler<Poco::AutoPtr<T>>: public AbstractTypeHandler
 	/// Specialization of type handler for Poco::AutoPtr
 {
 public:
+        typedef AutoPtr<T> TPtr; // new code here
```
aleks-f, line 5677: please do not send unnecessary comments in PRs - it is obvious from the diff that his is new code

Foundation/include/Poco/Any.h
```diff
@@ -28,6 +28,8 @@
 
 namespace Poco {
 
+#define POCO_SMALL_OBJECT_SIZE 64
```
aleks-f, line 31: this is defined in the Config.h, not needed here

aleks-f (changes requested):
I started reviewing, but this pull is totally unacceptable as is. please clean it it up, we don't need unnecessary comments in the code, duplicated defines, and necessary things removed. last but not least, send pull to devel branch, NOT master

Foundation/include/Poco/Any.h
sildeag: I am not experienced with c++.  The compiler generated an error `out of scope`, I am merely fixing the problem.

Data/include/Poco/Data/TypeHandler.h
sildeag: will remove the obvious comment.

Data/include/Poco/Data/Binding.h
```diff
@@ -1387,7 +1387,7 @@ inline AbstractBinding::Ptr use(T& t, const std::string& name = "")
 	// This can be resolved by either (a) using bind (which will copy the value),
 	// or (b) if the const ref is guaranteed to exist when execute is called
 	// (which can be much later!), by using the "useRef" keyword instead
-	poco_static_assert (!IsConst<T>::VALUE);
+	//poco_static_assert (!IsConst<T>::VALUE);
```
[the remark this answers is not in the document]
sildeag: this was a problematic assert.  the app openBliSSART would not compile without error, without this being removed.

sildeag:
I am not familiar enough with GitHub to know how to send my pull to the development branch.  I thought I had.  The code changes all seemed to work but I realize have not been tested multi-platform, just Ubuntu 22.04.  I hope they will help in some way which is why I submitted them.  The app openBliSSART compiles with c++20 and seems to work with these modifications.  I fully intend to send more requests as I am debugging all memory leaks and uninitialized variables in openBliSSART.

aleks-f:
That application states poco 1.3.6 as a dependency, which is something like 10 years old so please do not try to modify poco to fit the application peculiarities. As for c++20, we'll look into that eventually.

sildeag:
@aleks-f I followed your advice, closed #3731 and opened #3732 which merges now with your 'devel'.  I would appreciate it if some of my suggested changes were made as Format.h isn't an application peculiarity - in its current form it generates numerous compiler errors when using c++20 on Ubuntu 22.04.  My changes don't.  The new version of openBliSSART is a leap ahead of the old unmaintained app.  It specifies: Modified to use Ubuntu 22.04 using https://github.com/pocoproject/poco built with cmake, SDL2 and SDL2_sound.  Dependencies required are libsvm (version 3.25 is used), doxygen, pdflatex, fft3w, and possibly others.  Qt6.3.0.  Qt5 will be added later and CUDA 11.6 not tested and not used.  CUDA stills generates linker errors.  SDL2 is implemented not SDL1.  The new build process of openBliSSART uses cmake and c++20.  I was hoping that even though this is just one app that compiler issues from the poco application framework on Ubuntu 22.04 might be addressed.  I am using the app to improve c++ skills but at least one person, interested in Music and Sound, feels that having a working openBliSSART would be useful as it is a fairly unique app.  The ODBC cmake change for example reflects how anyone using unixODBC might find things.  I don't use it in this app but may one day.

PR #3732, opened 2022-08-03
Proposed changes to Format.h and others to allow compiling with c++20.

sildeag:
On Ubuntu 22.04 the current Format.h generates numerous compile time errors and these changes work.  Any.h has been modified due to a "scope error" to which Config.h did not provide resolution.  The other changes were to accommodate the app and the build environment and I have tried to avoid changing anything that might affect others.


aleks-f:
format issue with c++20 is fixed and will be released in [1.12.2](https://github.com/pocoproject/poco/milestone/44).
The rest in this pull is not accepted.

PR #3735, opened 2022-08-07
Make socket non-blocking in socket acceptor

JackyWoo:
Make socket non-blocking in socket acceptor

### DIFF
--- a/Net/include/Poco/Net/ParallelSocketAcceptor.h
+++ b/Net/include/Poco/Net/ParallelSocketAcceptor.h
@@ -144,6 +144,7 @@ public:
 	{
 		pNotification->release();
 		StreamSocket sock = _socket.acceptConnection();
+		sock.setBlocking(false);
 		_pReactor->wakeUp();
 		createServiceHandler(sock);
 	}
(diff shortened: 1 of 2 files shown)


aleks-f:
I don't think we should do this - users who want non-blocking sockets can (and probably already do) make them so. Perhaps it does make sense to be non-blocking by default, but with this change we'd only be breaking the existing user applications that actually want blocking sockets.

JackyWoo:
hi aleks-f,  thanks for response, should we add constructor param to `SocketAcceptor`?   And perhaps we could add possability to set some commonly used socket options in `SocketAcceptor`.

aleks-f:
We could do that, but it seems unnecessary because the end user has access to the socket and can modify its behavior any way they want in the handler. 

A more worthy acceptor improvement, in my opinion, would be to somehow (variadic template parameters? `enable_if`?) allow the customer's service handler to have additional constructor arguments:

https://github.com/pocoproject/poco/blob/be19dc4a2f30eb97cc9bdd7551460db11cc27353/Net/include/Poco/Net/SocketAcceptor.h#L160

Right now any service handler property or (co)ownership of an external object can not be set at the handler construction time.

JackyWoo:
@aleks-f ok, I will close this pr and I posted another pr #3741  try to add support of additional constructor parameters.

PR #3741, opened 2022-08-11
SocketAcceptor service handler additional constructor parameters

JackyWoo:
This pr try to fix https://github.com/pocoproject/poco/pull/3735#issuecomment-1209038877

### Change log
Support additional constructor parameters to socket acceptor serivce handler.

### DIFF
--- a/Net/include/Poco/Net/ParallelSocketAcceptor.h
+++ b/Net/include/Poco/Net/ParallelSocketAcceptor.h
@@ -79,8 +79,8 @@ public:
 		unsigned threads = Poco::Environment::processorCount(), const std::string& threadName = ""):
 		_threadName(threadName),
 		_socket(socket),
-		_parameters(std::forward_as_tuple(parameters...)),
 		_pReactor(&reactor),
+		_parameters(std::forward_as_tuple(parameters...)),
 		_threads(threads),
 		_next(0)
 		/// Creates a ParallelSocketAcceptor using the given ServerSocket, sets the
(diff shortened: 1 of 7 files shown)


aleks-f (changes requested):
`SocketAcceptorTest` should be added to `ReactorTestSuite`

https://github.com/pocoproject/poco/blob/46d5d58bc808f7fceefdd86ba55a6afe17ffa5c6/Net/testsuite/src/ReactorTestSuite.cpp#L16

JackyWoo:
@aleks-f  I have no experience with MSVC, could you give me a quick help for regenerating VS 150, 160, 170 projects configurations?

aleks-f (approved):

aleks-f:
@JackyWoo [Generating VS Projects using Progen](https://github.com/pocoproject/poco/wiki/Generating-VS-Projects-using-Progen)

aleks-f:
@JackyWoo I have generated files and moved this to [feat/acceptor-service-handler-args](https://github.com/pocoproject/poco/tree/feat/acceptor-service-handler-args)

There is compile error in VS 150, likely some ifdefs needed for older compilers (we must be c++14-compliant, so parts that are not must be appropriately `ifdef`'d):

```
2>c:\poco\net\include\poco\net\parallelsocketacceptor.h(76): error C2440: 'default argument': cannot convert from 'unsigned int' to 'const std::string &' (compiling source file src\SocketReactorTest.cpp)
2>c:\poco\net\include\poco\net\parallelsocketacceptor.h(76): note: Reason: cannot convert from 'unsigned int' to 'const std::string' (compiling source file src\SocketReactorTest.cpp)
2>c:\poco\net\include\poco\net\parallelsocketacceptor.h(76): note: No constructor could take the source type, or constructor overload resolution was ambiguous (compiling source file src\SocketReactorTest.cpp)
```

Issue #3740, opened 2022-08-11
buildwin.ps1 failed to build x64 


KevDi:
**Describe the bug**
Building the Poco Libraries with the `buildWin.ps1` script failed if  `-platform x64`  is used.

**To Reproduce**
This is the command that was used: `.\BuildWin.ps1' -poco_base . -vs 160 -action build -linkmode shared -config both -platform x64 -samples -tests  -omit 'Crypto;NetSSL_OpenSSL;Data/ODBC;Data/MySQL;Data/PostgreSQL;JWT;MongoDB'`

**Expected behavior**
Complete build of the Poco Library.

**Logs**
This is the Output from the Command above:

```
VS160COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\Common7\Tools\
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.11.15
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
Build configuration:
--------------------
Poco Base:               .
Version:                          160
Action:                          build
Link Mode:                shared
Configuration: both
Platform:                        x64
Tests:                            True
Samples:                         True
Build Tool: msbuild
Omit:                                Crypto;NetSSL_OpenSSL;Data/ODBC;Data/MySQL;Data/PostgreSQL;JWT;MongoDB
OpenSSL:                        $poco_base\openssl
+------------------------------------------------------------------
| VS project .\CppUnit_x64_vs160.vcxproj not found, skipping.
+------------------------------------------------------------------
```

The Issue is that in the end of the Powershell Script these lines are executed:

```
if ($platform -eq 'x64')           { $platformName = '_x64' }
elseif ($platform -eq 'WinCE') { $platformName = '_CE' } 
```

If these two lines are commented out the build works without any problems.
It seems this is a left over from the time when there where different `vcproj` files for `x64` and `win32`.
If you want i can create a PR and remove these two lines.

**Please add relevant environment information:**
 - Windows 10 & Visual Studio 2019 Professional
 - Poco Version 1.12.1

aleks-f:
Yes, create a PR to devel branch, please. We can also try to add a CI build

KevDi:
I just stumbled upon another problem. I have for example Visual Studio 2022 installed and also the Build Tools for Visual Studio 2019. If i run the script it gave me an error because it is unable to detect the 2019 Build tools.
[...]

PR #3749, opened 2022-08-14
buildwin.ps1 script error building the x64 version of Poco

KevDi:
If build x64 the scripts looks for the _x64.vcproj Files which no longer exists. By removing this the Build is possible.
If only the Build Tools for Visual Studio are available the script does not run because those are not recognized by the Select-VSSetupInstance Script.
Adding the parameter `-product *` also recognizes the Build Tools.

#3740

### DIFF
--- a/buildwin.ps1
+++ b/buildwin.ps1
@@ -409,10 +409,7 @@ function Build-Exec([string] $tool, [string] $vsProject, [switch] $skipStatic)
 
 
 function Build-Components([string] $extension, [string] $type)
-{
-	$dummy = Get-Content "$poco_base\components"
-	Write-Host "$dummy"
-	
+{	
 	Get-Content "$poco_base\components" | Foreach-Object {
 
 		$component = $_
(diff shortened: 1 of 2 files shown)


buildwin.ps1
```diff
@@ -505,8 +505,6 @@ function Build
 	else										 { $extension = 'vcxproj' }
 
 	$platformName = ''
```
aleks-f, line 507: It seems that `$platformName` can be completely removed?

aleks-f (changes requested):

buildwin.ps1
KevDi: mmh yes i think so i will remove it, test it and then push the changes

aleks-f (approved):

aleks-f:
@KevDi can you also try to add a CI entry for this?
It would be another entry like this, just using the `buildwin.ps1` script instead of cmake:
https://github.com/pocoproject/poco/blob/be19dc4a2f30eb97cc9bdd7551460db11cc27353/.github/workflows/ci.yml#L150-L160

KevDi:
Ok i will try to add the entry

KevDi:
@aleks-f  i tried to add the CI Section but it seems on the Actions Job Run the `-omit` does not work as expected. If i run the same command on my local Windows System it works. But on the Github Actions it still tries to build the Crypto Library.

KevDi:
@aleks-f  i test it again with the `Encodings` Project and it seems like on the github actions the check does not work. On my local system the Projects are skipped. But in the github actions no project is skipped and i don't know why

KevDi:
@aleks-f  ok i solved that issue. It seems like in the github actions the `split` method should only take exactly one character. I changed it from `.split(",;")`  to `.split(",")` and now the skipping of Projects works.
Should we keep the `,` or the `;` as the seperator for the omit list? I would go with the `;`because it is already mentioned in the help section.

aleks-f:
> @aleks-f ok i solved that issue. It seems like in the github actions the `split` method should only take exactly one character. I changed it from `.split(",;")` to `.split(",")` and now the skipping of Projects works. Should we keep the `,` or the `;` as the seperator for the omit list? I would go with the `;`because it is already mentioned in the help section.

I'd go with `,` because it is more intuitive and in line with `make` build. just change the doc, no big deal

aleks-f:
now you have some joy with VS2022. try to comment out other CI entries so that a push triggers run only of the entries you are working on, not the whole world

PR #3757, opened 2022-08-17
fix(CppUint) : RepeatedTest compile error

HR1025:
617ebfcc1b06756471287c0ba29039d
4e5055d9d53dfe6873799436c5fcc53

PR #3759, opened 2022-08-18
fix(CppUint) : RepeatedTest compile error

HR1025:


HR1025:
https://github.com/pocoproject/poco/pull/3757
Is it modified like this???

aleks-f:
I don't know, I can't make sense of it - the previous pull is gone, and there's no `RepeatedTest.cpp` in this one. In any case, if you want that cpp file added, the it will have to be added to the `CppUnit` Makefile and all VS projects.

HR1025:
i think the RepeatedTest.cpp is little redundant, so remove it , but include the header and delete forward declare

HR1025:
previous pull disappear may be i use the git push -f to update my commit

PR #3787, opened 2022-08-30
fix(Crypto) Update method to extract friendlyName from certificate

trzeciak:
Hey,
in single codebase I need to use webrtc with BoringSSL and Poco.
I ran into a few problems compiling poco with BoringSSL, I'll try to suggest some changes to make it easier for others.

First things first: the current `extractFriendlyName` method doesn't compile with BoringSSL, I asked the BoringSSL team for advice and I have a tip on that:
https://bugs.chromium.org/p/boringssl/issues/detail?id=515

The discussion shows that this is a compatible change with a positive effect on the current code.

I run the Crypto test suite and it returned:
```
...
OK (44 tests)
```

BTW: These changes are not `BoringSSL` specific.

### DIFF
--- a/Crypto/src/PKCS12Container.cpp
+++ b/Crypto/src/PKCS12Container.cpp
@@ -125,18 +125,18 @@ PKCS12Container::~PKCS12Container()
 
 std::string PKCS12Container::extractFriendlyName(X509* pCert)
 {
-    std::string friendlyName;
-    if(pCert)
-    {
-        int len = 0;
-        char* pBuffer = reinterpret_cast<char *>(X509_alias_get0(pCert, &len));
-        if (pBuffer) {
-            friendlyName.append(pBuffer, len);
-        }
-    }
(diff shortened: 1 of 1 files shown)


Crypto/src/PKCS12Container.cpp
```diff
@@ -128,19 +128,11 @@ std::string PKCS12Container::extractFriendlyName(X509* pCert)
...
-			char* pBuffer = PKCS12_get_friendlyname(pBag);
-			if(pBuffer)
-			{
-				friendlyName = pBuffer;
-				OPENSSL_free(pBuffer);
-			}
-			if(pBags) sk_PKCS12_SAFEBAG_pop_free(pBags, PKCS12_SAFEBAG_free);
+		int length = 0;
+		char* pBuffer = reinterpret_cast<char*>(X509_alias_get0(pCert, &length));
+		if (pBuffer) {
```
trzeciak, line 133:
```suggestion
		if (pBuffer)
		{
```

PR #3810, opened 2022-09-22
Custom rotate, archive and purge strategies for FileChannel

andrewauclair:
This is an attempt at making an improved version of #3101 that has tests.

`setRotationStrategy`, `setArchiveStrategy` and `setPurgeStrategy` allow setting custom strategies to be used by the `FileChannel`.

These new functions take a non-null pointer and `FileChannel` owns the pointer after the call and will delete it when necessary.

Also changed `FileChannel` to use a new `NullRotationStrategy` and `NullPurgeStrategy` instead of null pointers.

One thing to note about these changes right now is that setting a property will still internally create a new strategy in `FileChannel`, overwriting the one provided to the set strategy functions. I think this is ok and I don't have any better ideas right now.

Personally, I've wanted this ability with Poco at work for a while so that we can start each log file with the version of software. It's currently difficult to know what version a customer is using when they send us log files.

### DIFF
--- a/Foundation/include/Poco/ArchiveStrategy.h
+++ b/Foundation/include/Poco/ArchiveStrategy.h
@@ -42,6 +42,9 @@ public:
 	ArchiveStrategy();
 	virtual ~ArchiveStrategy();
 
+	virtual LogFile* open(LogFile* pFile) = 0;
+		/// Open a new log file and return it.
+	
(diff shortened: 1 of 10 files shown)


aleks-f:
@andrewauclair for some reason, this did not trigger CI builds. can you please push a dummy commit to trigger it so we can see it green before merge. I'll mark it for 1.13.0

Issue #2331, opened 2018-05-15
Improve implementation of logging macros.


obiltschnig:
Currently, the logging macros are defined as:

```
#define poco_error(logger, msg) \
   if ((logger).error()) (logger).error(msg, _FILE_, __LINE__); else (void) 0
```

This can cause compiler warnings if used in an `if` statement.

Possible alternative implementations:

```
// works but generates warning
#define poco_error(logger, msg) \
   do { if ((logger).error()) (logger).error(msg, _FILE_, __LINE__); } while(0)

#define poco_error(logger, msg) \
   for( ; ; ) { if ((logger).error()) (logger).error(msg, _FILE_, __LINE__); break;}
```

PR #3862, opened 2022-11-07
Improve implementation of logging macros. #2331

bas524:
The GNU compiler emits a warning if nested "if" statements are followed by an "else" statement and braces are not used to explicitly disambiguate the "else" binding.  This leads to problems with code like:

   if (gate)
     ASSERT_*(condition) << "Some message";

The "switch (0) case 0:" idiom is used to suppress this.


bas524:
tests fail reason fixed in #3863

aleks-f:
@bas524 wouldn't it be easier to just add curly braces and avoid the conditionally defined switch/case gymnastics?

bas524:
@aleks-f, You are right, I was wrong, I did some checks https://godbolt.org/z/W5nnYrzx6
I made new PR.

PR #3866, opened 2022-11-08
Socket::isConnected()

### LINKED ISSUE #3799: Poco::Redis reconnect issue
Hello, looks like redis client does not support reconnect. I connect to redis via vpn. If vpn connection will be dropped and reconnected again redis client might be broken.
For example

```c++
Poco::Redis::Client client("127.0.0.1:6379");
if (Redis.isConnected())
{
//// OK
logger().log(Poco::Message("app", "Redis connected", Poco::Message::PRIO_INFORMATION, __FILE__, __LINE__));
}
// here can be some commands
// drop vpn
// reconnect vpn
if (Redis.isConnected())
{
//// still connected (but really not)
logger().log(Poco::Message("app", "Redis connected", Poco::Message::PRIO_INFORMATION, __FILE__, __LINE__));
}
// try to send any another command
catch IOException("No buffer space available", 10054);
```
изображение

Windows 10 x64 with poco 2b11cb50bf0ff84cd08acf755faac0e4170337a9

Jihadist:
Add check the current state of socket
Fixes #3799

### DIFF
--- a/Net/include/Poco/Net/Socket.h
+++ b/Net/include/Poco/Net/Socket.h
@@ -127,6 +127,9 @@ public:
 		/// Returns true if socket is a raw socket,
 		/// false otherwise.
 
+	bool isConnected() const;
+		/// Returns true if socket connected.
+
 	void close();
 		/// Closes the socket.
 
--- a/Net/src/Socket.cpp
+++ b/Net/src/Socket.cpp
@@ -95,6 +95,13 @@ Socket::~Socket()
 	if (_pImpl) _pImpl->release();
 }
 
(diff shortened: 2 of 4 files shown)


Jihadist:
[@obiltschnig](https://github.com/obiltschnig) you are right, of course. It should be only for tcp sockets. Maybe it's not correct for tcp sockets too. But I tried it with winsock and unix sock, and it works. I tried connection, disconnection and connection failure.
I have no idea how to find an appropriate solution for [#3799](https://github.com/pocoproject/poco/issues/3799). Can you suggest something?

PR #3868, opened 2022-11-11
MongoDB replica set uri support

artem-yadr:
While working with Poco, I noticed that there is no way to connect to a MongoDB cluster using a replica set uri. Here I made basic support for mongodb replica set uri(uri with host:port enumiration) and readPreference option.

### DIFF
--- a/MongoDB/src/Connection.cpp
+++ b/MongoDB/src/Connection.cpp
@@ -221,7 +221,7 @@ void Connection::connect(const std::string& uri, SocketFactory& socketFactory)
         else if (it->first == "readPreference")
         {
             readPreference= it->second;
-            if (readPreference != "primary" or readPreference != "secondary"){
+            if (readPreference != "primary" and readPreference != "secondary"){
                 throw Poco::InvalidArgumentException("read preference", readPreference);
             }
         }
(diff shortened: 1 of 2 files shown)


MongoDB/src/Connection.cpp
```diff
@@ -145,68 +145,159 @@ void Connection::connect(const Poco::Net::StreamSocket& socket)
...
-			password.assign(userInfo, pos, userInfo.size() - pos);
-		}
-		else username = userInfo;
-
-		Database database(databaseName);
-		if (!database.authenticate(*this, username, password, authMechanism))
-			throw Poco::NoPermissionException(Poco::format("Access to MongoDB database %s denied for user %s", databaseName, username));
-	}
+    std::vector<std::string> str_addresses;
+    std::string new_uri;
```
aleks-f, line 149: naming convention

aleks-f (changes requested):

MongoDB/src/Connection.cpp
artem-yadr: fixed it

artem-yadr:
@aleks-f any comments about pr? I made the requested changes but have not received a response.

PR #3882, opened 2022-11-22
add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where POCO already runs (see https://github.com/pocoproject/poco/pull/3149 and https://github.com/google/oss-fuzz/pull/4736). CIFuzz can help detect regressions and fuzzing catch build issues early, and has a variety of features (see the URL above). It can also help to get the fuzzing closer into the development workflow. In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


aleks-f:
#4302

PR #3885, opened 2022-11-23
Use map from key to count instead of multiset

davenger:
We noticed slow performance of XMLConfiguration::enumerate when a node had lots of children with the same name. In this case std::multiset is not needed for assigning consequent indexes to children, std::map is better in terms of number allocations and memory consumption.


davenger:
I added test XMLConfigurationTest::testManyKeys that reproduces the problem by enumerating 100K elements with the same name.
Without the fix it takes around 2 minutes on my machine in release and with the fix it take 0.1 seconds.

Issue #3907, opened 2023-01-02
NetSSL_Win certificate verification failure when intermediate certificates are not available in the Windows trusted store


bhatkarthik:
https://github.com/pocoproject/poco/blob/69d15c5dddedf5d9adb86ade0dcee8254bb3ede5/NetSSL_Win/src/SecureSocketImpl.cpp#L1222

- Poco is passing the `hAdditionalStore` parameter as null when calling `CertGetCertificateChain`.
- When the intermediate certificates are not installed on the machine, looks like `hCertStore` of the `CERT_CONTEXT` member needs to be passed as `hAdditionalStore` parameter to `CertGetCertificateChain`, for it to be able to build the entire chain.
- Without this, the created certificate chain seems to contain only the leaf certificate and thus the certificate validation fails
  - It fails to find the back of the chain (in this case leaf itself), in the certificate store. 
  - https://github.com/pocoproject/poco/blob/69d15c5dddedf5d9adb86ade0dcee8254bb3ede5/NetSSL_Win/src/SecureSocketImpl.cpp#L1288
- Windows (Schannel) gives the handle to the store containing intermediate certificates under `hCertStore` memeber of `CERT_CONTEXT` certificate obtained from `QueryContextAttributes` (Schannel)

Note: 
- Curl and Chromium seems to be passing this `hAdditionalStore` argument
  - https://chromium.googlesource.com/chromium/src/net/+/master/cert/cert_verify_proc_win.cc#1208
  - https://github.com/curl/curl/blob/master/lib/vtls/schannel_verify.c#L669

grmcdorman:
There is a related issue: if Windows downloads the root CA to the Trusted Root store (LocalMachine\ROOT) when the chain is built, the subsequent check for the root CA fails despite the certificate being present. It appears that there is some sort of caching going in Windows when Poco keeps the root store open; opening the root store _again_ and adding it to the context certificate will allow the verification to pass, i.e. this code as a workaround in the `onInvalidCertificate` callback:
```
            auto &manager{ Poco::Net::SSLManager::instance() };
            auto contextPointer{ manager.defaultClientContext() };

            static HCERTSTORE store{ nullptr };
            if (store != nullptr)
            {
                CertRemoveStoreFromCollection(contextPointer->certificateStore(), store);
                CertCloseStore(store, 0);
            }

[...]

PR #3910, opened 2023-01-04
Pass `pServerCert->hCertStore` for `hAdditionalStore` argument in `CertGetCertificateChain` call

bhatkarthik:
Fix for #3907 
- Passing `hCertStore` member of the CERT_CONTEXT as `hAdditionalStore` parameter to `CertGetCertificateChain` call
- When the intermediate certificates are not installed on the machine, this will be able to build the entire chain.


matejk:
Fixed #3907 .

PR #3911, opened 2023-01-04
Fix: define NOMINMAX only if not defined yet

matejk:
Fixes compiler warnings when NOMINMAX is already defined in a project using Poco.


matejk:
Closing because a PR (#3906) already exists.

Issue #3739, opened 2022-08-10
Drop support for OpenSSL < 1.1.1 (Poco 1.12.x compile error on systems with OpenSSL 1.0.0)


gyee-penguin:
**Describe the bug**
I get a compile error for Poco 1.12.0 and 1.12.2 on CentOS 7 (with GCC 10) or Ubuntu 16 (both have OpenSSL 1.0.0) in the Crypto component.  Compilation works for 1.11.1, 1.11.2, and 1.11.3.

**To Reproduce**
CentOS 7, Use devtoolset-10, Download Poco 1.12.2 source.  Compile.

**Expected behavior**
I didn't see anything in the change log to suggest that OpenSSL 1.0.0 support had been dropped for Poco 1.12.x or that OpenSSL 1.1.1 is required.  I thought it would build.

**Logs**
```
/opt/rh/devtoolset-10/root/usr/bin/g++  -DPOCO_ENABLE_CPP11 -DPOCO_ENABLE_CPP14 -DPOCO_HAVE_FD_EPOLL -DPOCO_OS_FAMILY_UNIX -DPOCO_STATIC -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -D_REENTRANT -D_THREAD_SAFE -D_XOPEN_SOURCE=500 -I/home/builder/src/poco_container/poco/Crypto/include -I/home/builder/src/poco_container/poco/Crypto/src -I/home/builder/src/poco_container/poco/Foundation/include -O2 -g -DNDEBUG   -std=gnu++14 -MD -MT Crypto/CMakeFiles/Crypto.dir/src/Envelope.cpp.o -MF Crypto/CMakeFiles/Crypto.dir/src/Envelope.cpp.o.d -o Crypto/CMakeFiles/Crypto.dir/src/Envelope.cpp.o -c /home/builder/src/poco_container/poco/Crypto/src/Envelope.cpp
/home/builder/src/poco_container/poco/Crypto/src/Envelope.cpp: In constructor 'Poco::Crypto::Envelope::Envelope(int)':
/home/builder/src/poco_container/poco/Crypto/src/Envelope.cpp:27:8: error: invalid operands of types 'int' and 'void' to binary 'operator!='
   27 |  if (1 != EVP_CIPHER_CTX_init(_pCtx))
      |      ~ ^~ ~~~~~~~~~~~~~~~~~~~~~~~~~~
      |      |                       |
      |      int                     void
```

```
[...]

PR #3912, opened 2023-01-04
Fixed compile error for OpenSSL 1.0 systems (#3739)

gyee-penguin:
Trying again -- not sure why it failed checks last time.

### DIFF
--- a/Crypto/src/Envelope.cpp
+++ b/Crypto/src/Envelope.cpp
@@ -29,8 +29,12 @@ Envelope::Envelope(int cipherNID): _pCipher(EVP_get_cipherbynid(cipherNID)),
 {
 	poco_check_ptr(_pCipher);
 	poco_check_ptr(_pCtx);
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
 	if (1 != EVP_CIPHER_CTX_init(_pCtx))
 		handleErrors(std::string("Envelope():EVP_CIPHER_CTX_init()"));
+#else
+	EVP_CIPHER_CTX_init(_pCtx);
+#endif
 	_iv.resize(ivSize(), 0);
 }
 


aleks-f:
Pull requests should go to develop branch

gyee-penguin:
Oops, my apologies.  I've changed the Pull Request to base off of devel.

gyee-penguin:
Hello @aleks-f I don't think our patch is the reason for the checks failing.  I think the checks in the devel branch were failing before we committed.

gyee-penguin:
Rebased with latest devel to see if that helps it pass checks

gyee-penguin:
There seems to be a bug with the github checks.

bas524:
Hi! I think, that problem a little bit more complicated than this PR. Please look at #3939

gyee-penguin:
Hello!  Our patch does address the problem you mention.  We use a C macro to test the OpenSSL version and handle both cases (one where the function returns void and one that returns an int).

gyee-penguin:
Hello @aleks-f Do you know if our patch is acceptable for the 1.12.5 release?

bas524:
> Hello! Our patch does address the problem you mention. We use a C macro to test the OpenSSL version and handle both cases (one where the function returns void and one that returns an int).

Please, try build with openssl version 1.0.2. And you will see that in this version EVP_CIPHER_CTX returns void, and it will be a compile time error.
I have gotten it on solaris with openssl 1.0.2f

gyee-penguin:
Hello @bas524, 

We successfully compiled against OpenSSL 1.0.2k on CentOS 7.  I agree that EVP_CIPHER_CTX_init returns void.  I can't explain why you are getting compile errors on your Solaris system, but below I explain why our patch works:

```
#if OPENSSL_VERSION_NUMBER >= 0x10100000L
	if (1 != EVP_CIPHER_CTX_init(_pCtx))
		handleErrors(std::string("Envelope():EVP_CIPHER_CTX_init()"));
#else
	EVP_CIPHER_CTX_init(_pCtx);
#endif
```

The first line ```#if OPENSSL_VERSION_NUMBER >= 0x10100000L``` tests if the OpenSSL version is >= 1.1.0.  If this is true, then we compile the version of EVP_CIPHER_CTX_init that returns an int.  

Otherwise, we compile the version that returns void.  This is true if OpenSSL version is 1.0.2, for example.

bas524:
> Hello @bas524,
> 
> We successfully compiled against OpenSSL 1.0.2k on CentOS 7. I agree that EVP_CIPHER_CTX_init returns void. I can't explain why you are getting compile errors on your Solaris system, but below I explain why our patch works:
> 
> ```
> #if OPENSSL_VERSION_NUMBER >= 0x10100000L
> 	if (1 != EVP_CIPHER_CTX_init(_pCtx))
> 		handleErrors(std::string("Envelope():EVP_CIPHER_CTX_init()"));
> #else
> 	EVP_CIPHER_CTX_init(_pCtx);
> #endif
> ```
> 
> The first line `#if OPENSSL_VERSION_NUMBER >= 0x10100000L` tests if the OpenSSL version is >= 1.1.0. If this is true, then we compile the version of EVP_CIPHER_CTX_init that returns an int.
> 
> Otherwise, we compile the version that returns void. This is true if OpenSSL version is 1.0.2, for example.

I'm sorry. I was wrong. It was my inattention.
Your fix is correct. I have reused it for PR with fixes for solaris

Issue #3923, opened 2023-01-23
UDP Multicast : `leaveGroup()` method always throw an exception


Crivaledaz:
I am trying to implement an UDP multicast socket on Windows 10 with POCO. I can create the multicast socket and join a multicast group without issue. When I want to close the socket, I try to leave the multicast group previously joined, but it raises an `NotFoundException` from [`firstAddress()`](https://github.com/pocoproject/poco/blob/devel/Net/src/NetworkInterface.cpp#L331) method.

## Analysis

The [`NotFoundException`](https://github.com/pocoproject/poco/blob/devel/Net/src/NetworkInterface.cpp#L331) is raised because `_addressList` is empty.

 In fact, the method [`firstAddress()`](https://github.com/pocoproject/poco/blob/devel/Net/src/NetworkInterface.cpp#L321-L332) is called on object `interfc` from method [`leaveGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L271), but this object is empty since it is created just before in method  [`leaveGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L260) with the default constructor. This [constructor](https://github.com/pocoproject/poco/blob/devel/Net/src/NetworkInterface.cpp#L163-L171) does not initialize the attribute `_addressList` which remains empty.

So the problem come from the [line 260 of `MulticastSocket.cpp`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L260), and I don't know how this can work.

## Solution

I think the method [`leaveGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L258) should use `findFirstInterface(groupAddress)` instead of create a new `NetworkInterface`, as it is done in the method [`joinGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L197) which works well.

I can do a merge request if needed.

## Code example

Below an C++ implementation to reproduce the bug : 

```cpp
#include <Poco/Net/DatagramSocket.h>
#include <Poco/Net/SocketAddress.h>
#include <Poco/Net/MulticastSocket.h>

using Poco::Net::DatagramSocket;
using Poco::Net::SocketAddress;
using Poco::Net::MulticastSocket;

int main()
{
    const std::string multicastAddress = "234.2.2.2";
    const Poco::UInt16 multicastPort = 4040;
[...]

PR #3929, opened 2023-01-24
Fix multicast leave group

Crivaledaz:
As exposed in the issue #3923, using the method [`leaveGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L261) on a multicast socket raises an [`NotFoundException`](https://github.com/pocoproject/poco/blob/devel/Net/src/NetworkInterface.cpp#L331). This patch allows to use the same interface than the [`joinGroup()`](https://github.com/pocoproject/poco/blob/devel/Net/src/MulticastSocket.cpp#L197) method. Thus, the interface exists and the exception is not raised.

I have successfully tested my patch on Fedora 37 and Windows 10, using the example given in the issue.

This pull request solves issue #3923.

### DIFF
--- a/Net/src/MulticastSocket.cpp
+++ b/Net/src/MulticastSocket.cpp
@@ -257,8 +257,7 @@ NetworkInterface MulticastSocket::findFirstInterface(const IPAddress& groupAddre
 
 void MulticastSocket::leaveGroup(const IPAddress& groupAddress)
 {
-	NetworkInterface intf;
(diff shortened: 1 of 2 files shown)


aleks-f:
@Crivaledaz can you please add a test case for this [here](https://github.com/pocoproject/poco/blob/devel/Net/testsuite/src/MulticastSocketTest.cpp)?

Issue #3918, opened 2023-01-11
Static FastMutex fails to lock when issued from another thread on linux


JoKre17:
Hello,
as the title says, we are getting an error and undefined behavior after logging 2-3 messages when using the LoggingConfigurator with the following logging properties file:
```
logging.formatters.pretty.class = PatternFormatter
logging.formatters.pretty.pattern = %Y-%m-%d %H:%M:%S.%i [%z] [%p] %s: %t
logging.formatters.pretty.times = local

logging.channels.console.class = ColorConsoleChannel
logging.channels.console.formatter = pretty

logging.channels.file.class = FileChannel
logging.channels.file.formatter = pretty
logging.channels.file.path = my_logging_file.log

logging.channels.splitter.class = SplitterChannel
logging.channels.splitter.channel = console,file

logging.channels.async.class = AsyncChannel
logging.channels.async.channel = splitter

logging.loggers.root.channel = async
logging.loggers.root.level = debug
```
We are using the AsyncChannel which runs another Thread. Further down the logging channel chain we are reaching the channel named "console" which is wrapped with the formatter named "pretty" which is calling `PatternFormatter::format`. 
As we include `"[%z]"` in the format, the function in return calls `Timezone::utcOffset()`.

Under linux the Timezone implementation contains the following (I've stripped out the unrelevant parts):
```
class TZInfo {
 public:
  TZInfo() { tzset(); }

  int timeZone() {
    Poco::FastMutex::ScopedLock lock(_mutex);  // <-- HERE we are getting an exception/undefined behavior

    tzset();
    return -timezone;
  }

 private:
  Poco::FastMutex _mutex;
};
static TZInfo tzInfo;

int Timezone::utcOffset() { return tzInfo.timeZone(); }
```

As `TZInfo tzInfo` including the FastMutex member is static, it will be created by the main thread. Now the logging contains an AsyncChannel running on a different thread which then fails to lock the FastMutex. (Line 27 in `Foundation/src/Timezone_UNIX.cpp`)

The code running `pthread_mutex_lock` (in this case `MutexImpl::lockImpl`) throws an exception without any catch handler which is likely the cause for the process hanging.
It seems on `pthread_mutex_lock(&_mutex)` we are getting the error code for EINVAL indicating that our logging thread has a higher priority than the main thread or the mutex handle `_mutex` is invalid. 

[Linux documentation](https://linux.die.net/man/3/pthread_mutex_lock) says:
> EINVAL
[...]

PR #3941, opened 2023-02-08
Regarding #3918: Fixing Mutex / FastMutex by replacement with std library implementations

JoKre17:
As described in #3918 , `Poco::Mutex` and `Poco::FastMutex` being used in contexts (e.g. `static`) where they are accessed by multiple threads leads to `EINVAL` errors during access of the lock.

This does not happen with the standard implementation of `mutex` and `recursive_mutex` which is what was used to fix the very occurrence of this issue mentioned in #3918 .
Thus this pullrequest prepares the Poco codebase for the deprecation of `Poco::Mutex` and `Poco::FastMutex` classes and replaces all occurrences with standard implementations.

### DIFF
--- a/ApacheConnector/include/ApacheApplication.h
+++ b/ApacheConnector/include/ApacheApplication.h
@@ -40,7 +40,7 @@ public:
 private:
 	bool _ready;
 	ApacheRequestHandlerFactory _factory;
-	Poco::FastMutex _mutex;
+	std::mutex _mutex;
 };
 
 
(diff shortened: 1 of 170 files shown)


aleks-f:
tentatively marked for 1.13.0, @obiltschnig please confirm. I'm not sure why there's no CI run on this pull

aleks-f:
I'm thinking if we do this, it will be done like [this](https://github.com/pocoproject/poco/pull/3954)

PR #3954, opened 2023-02-22
Complimentary to #3918 (std::*mutex wrapper)

bas524:
Complimentary to #3918
I think that we can use Poco::Mutex and Poco::FastMutex as wrappers for std::recursive_mutex and std::mutex instead of replacing

For using std::*mutexes switch on cmake-option POCO_ENABLE_STD_MUTEX


aleks-f (changes requested):
please also add this option to [Config.h](https://github.com/pocoproject/poco/blob/devel/Foundation/include/Poco/Config.h)

PR #3964, opened 2023-03-07
Bugfix/bulk insert vector strings

ostropik:
- fix(ODBC): BULK INSERT vector of strings. When using SQL_NTS and the string has length as maximum for the column, into the cumulative buffer must be present zero-terminates


aleks-f:
@ostropik please redirect this pull to the `devel` branch

ostropik:
this is correct? https://github.com/pocoproject/poco/pull/3972

PR #1886, opened 2017-09-12
POCO_UNBUNDLED fine tuning

proller:
Our project uses zlib-ng library, and we need to unbundle only zlib.

Added cmake options:
POCO_UNBUNDLED_PCRE
POCO_UNBUNDLED_ZLIB
POCO_UNBUNDLED_EXPAT
POCO_UNBUNDLED_SQLITE

all defaults to POCO_UNBUNDLED

PR #3988, opened 2023-03-25
Fine grained flags for POCO_UNBUNDLED

Smarre:
Allows building only certain parts with external libraries, instead of forcing everything to be built either against in-source dependencies or against external dependencies.

Supersedes #1886.


Smarre:
@aleks-f Both Makefile and CMake compilations should work just with `POCO_UNBUNDLED` set; it will define the more fine-grained flags. If that is necessary for code level, I think some kind of `or` condition for the if is necessary. Or did I miss something?

aleks-f:
> If that is necessary for code level, I think some kind of `or` condition for the if is necessary.

yes; we can not assume that the proposed default constraints (ie. `POCO_UNBUNDLED` defined => all individual `*_UNBUNDLED` defined) will always be true - they are easy to work around, eg. one could set only `POCO_UNBUNDLED` on the command line when building and expect everything to be unbundled, but it won't be, because only individual defines are checked everywhere.

Long story short, an individual 3rd party library (eg. Expat) should be considered unbundled like this:
```c++
#if defined(POCO_UNBUNDLED) || defined(POCO_UNBUNDLED_EXPAT)
```

aleks-f (changes requested):
- [ ] check both "global" and "individual" defines throughout build system and code
- [ ] search returns 63 occurrences of `POCO_UNBUNDLED`; make sure all are addressed
- [ ] update [documentation](https://github.com/pocoproject/poco/blob/5c3f827c8639a996dd0d27a77f48d343ed6fe524/doc/00200-GettingStarted.page#L393)

PR #3992, opened 2023-04-01
Fix thread counter leak

alex65536:
I noticed that Poco HTTP server becomes slow after some uptime, especially when the load is high. The reason is that there is a race condition when decrementing `_currentThreads`. Because of this, `_currentThreads` always becomes bigger than the current number of threads.

The fix was already proposed before in the Clickhouse fork, see https://github.com/ClickHouse/poco/pull/28. I have just cherry-picked it to be merged into upstream Poco.

### DIFF
--- a/Net/include/Poco/Net/TCPServerDispatcher.h
+++ b/Net/include/Poco/Net/TCPServerDispatcher.h
@@ -101,30 +101,6 @@ private:
 	TCPServerDispatcher(const TCPServerDispatcher&);
 	TCPServerDispatcher& operator = (const TCPServerDispatcher&);
 
-	class ThreadCountWatcher
-	{
-	public:
-		ThreadCountWatcher(TCPServerDispatcher* pDisp) : _pDisp(pDisp)
-		{
-		}
-
-		~ThreadCountWatcher()
-		{
-			FastMutex::ScopedLock lock(_pDisp->_mutex);
-			if (_pDisp->_currentThreads > 1 && _pDisp->_queue.empty())
-			{
-				--_pDisp->_currentThreads;
-			}
-		}
-
-		private:
-			ThreadCountWatcher();
-			ThreadCountWatcher(const ThreadCountWatcher&);
-			ThreadCountWatcher& operator=(const ThreadCountWatcher&);
-
-			TCPServerDispatcher* _pDisp;
-	};
-
 	std::atomic<int> _rc;
 	TCPServerParams::Ptr _pParams;
 	std::atomic<int>  _currentThreads;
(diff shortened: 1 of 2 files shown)


aleks-f:
@alex65536 wouldn't just adding check for `_stopped` fix the problem?
```c++
if (_pDisp->_stopped || (_pDisp->_currentThreads > 1 && _pDisp->_queue.empty()))
```

alex65536:
> @alex65536 wouldn't just adding check for _stopped fix the problem?

I don't think so. The bug is that the server is slowly degrading, and the server doesn't need to be stopped in order to reproduce it. So I don't believe that just adding `stopped_` will help.

aleks-f:
> > @alex65536 wouldn't just adding check for _stopped fix the problem?
> 
> I don't think so. The bug is that the server is slowly degrading, and the server doesn't need to be stopped in order to reproduce it. So I don't believe that just adding `stopped_` will help.

OK, but then the fix you submitted won't fix it either, because it's doing the same thing as the destructor would if modified as I suggested. If I'm missing something, please explain in detail, "slowly degrading" does not make sense to me; I admit I have scarce time to spend on this, but I do not want to again merge fix which is not fixing the problem.

alex65536:
> OK, but then the fix you submitted won't fix it either, because it's doing the same thing as the destructor would if modified as I suggested. If I'm missing something, please explain in detail, "slowly degrading" does not make sense to me; I admit I have scarce time to spend on this, but I do not want to again merge fix which is not fixing the problem.

The difference is that both the current code and your proposed change perform the checks twice, and both checks don't happen under the same lock.

Thus, it can happen that the value of `_currentThreads` variable will change between those checks. First, on this line

```
if (_stopped || (_currentThreads > 1 && _queue.empty())) break;
```

we have `_currentThreads == 2`, and decrease it by one shortly afterwards.

Then we have

```
if (_pDisp->_currentThreads > 1 && _pDisp->_queue.empty())
{
	--_pDisp->_currentThreads;
}
```

Now, we have `_currentThreads == 1`, so we don't decrement it. This leads us to the situation when the `_currentThreads` variable has greater value than the number of current threads.

After each such race as described above, the difference between `_currentThreads` and the real number of current threads will become greater by one. It basically means that we run less threads for handling the requests than expected. In the end, after much time we may end up with just one thread handling the requests, and we cannot start any more threads because `_currentThreads` is already equal to `_pParams->getMaxThreads()`.

I have also tested the fix on my internal codebase (which I won't disclose, unfortunately), and the patch does indeed fix the issue.

alex65536:
Any progress on the review?

PR #3994, opened 2023-04-02
Revert "Fix dead lock on Timer destructor (#3987)"

### LINKED ISSUE #3986: Fix dead lock on Timer destructor

vojinilic:
This reverts commit 39a8b9a7c7a03e7a8c46d4f6531d216ec3ed95da.
It actually brakes cancel since dequeueNotification only returns notification which is due to execute now.
And there is no test which checks if notifications is executed after cancel.

fixes #3986

### DIFF
--- a/Util/src/Timer.cpp
+++ b/Util/src/Timer.cpp
@@ -101,6 +101,7 @@ public:
 			pNf = static_cast<TimerNotification*>(queue().dequeueNotification());
 		}
 
+		queue().clear();
 		_finished.set();
 		return true;
 	}


aleks-f:
The original fix actually made sense because I've experienced some problems in the past that could be because of that; it seems that what is missing here is a way to shut down the notification queue to prevent it from accepting further notifications. Or return `false` at the end of `CancelNotification::execute()`

In any case, if you want to revert this, send the 1.12.5 revert pul also. I would prefer it being properly fixed, but I don't have time for it myself.

vojinilic:
I've already sent a revert for devel.
Here is a pull request for 1.12.5 as well.
https://github.com/pocoproject/poco/pull/3995

As for proper fix in the current state of the code I have an idea.
Instead of returning false or blocking we can do the following.
Add a new method in NotificationQueue similar to dequeueNotification but
instead of returning a task which needs to be executed now that it returns
the next task in queue regardless of when it needs to be executed. Let's
call that method dequeueNextNotification, I am not stuck on naming it's
just for reference now.

Execution of Cancel notification would turn to something like this then.

size_t pendingSize = queue().dequeueNotification().size();
Poco::AutoPtr<TimerNotification> pNf =
static_cast<TimerNotification*>(queue().dequeueNextNotification());
while (pNf && pendingSize > 0)
{
    pendingSize--;
    ...
    pNf =
static_cast<TimerNotification*>(queue().dequeueNextNotification());
}

So we would get the size of pending tasks when we start to execute Cancel.
We want to dequeue that many notifications regardless of when they need to
be executed. We flush one by one taking special consideration of stop and
cancel tasks as before. After we leave the while loop we will again remove
the clear() method.
Does that make sense?
I can also write all of that down in the issue itself, and possibly take
care of that in the coming weeks.
Vojin Ilic

aleks-f:
@vojinilic make a new pull with your proposal and we'll look into it. I know testing is often not easy with async and timed stuff, but having tests would definitely be beneficial.

PR #3998, opened 2023-04-04
Fix: MongoDB::OpMsgCursor fail with "Batch size for getMore must be positive, but received: 0"

matejk:
Zero batch size for the first batch is performance optimisation to get errors fast from the server without much processing. Implementation of OpMsgCursor did not handle it properly and server responded with:

```
{
	"ok": 0,
	"errmsg": "Batch size for getMore must be positive, but received: 0",
	"code": 2,
	"codeName": "BadValue"
}

```

This pull request contains two changes:

* Fix: MongoDB::OpMsgCursor did not handle zero batch size properly: cursor requests failed.

* Improvement: Add emptyFirstBatch to indicate that the size of the first batch shall be zero for performance to get potential error ASAP from the server.

@aleks-f: This PR fixes new MongoDB wire protocol that was merged to branch devel, but is not yet released.

### DIFF
--- a/MongoDB/include/Poco/MongoDB/OpMsgCursor.h
+++ b/MongoDB/include/Poco/MongoDB/OpMsgCursor.h
@@ -36,11 +36,16 @@ public:
 	virtual ~OpMsgCursor();
 		/// Destroys the OpMsgCursor.
 
+	void setEmptyFirstBatch(bool empty);
+		/// Empty first batch is used to get error response faster with little server processing
+
+	bool emptyFirstBatch() const;
+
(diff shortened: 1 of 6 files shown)


aleks-f:
@matejk are all these contained in [971a7cc](https://github.com/pocoproject/poco/commit/971a7cc670c881964423e85ea9b9a91088c14142)? If so, then we should just close this

matejk:
Yes, the changes are contained there.

matejk:
Already contained in PR #4004.

PR #3999, opened 2023-04-04
Fix hang in destructor

### LINKED ISSUE #3986: Fix dead lock on Timer destructor

vojinilic:
Consider following situation. A class owns a timer. In destructor of that class we call .cancel() asynchronous on timer before it's destruction. Now timer is executing cancel in it's own internal thread, while it's doing that destructor of timer is called from owner's destructor. Timer destructor enqueues stop notification. If that enqueue is happening just after while loop from cancel notification, stop notification is gonna be dropped and timer will never stop. Fix: Add new method in TimedNotificationQueue which will return a notification regardless of the time it needs to be executed. Get number of pending tasks in the queue. Flush out that many notifications from queue while taking special consideration of pending Stop and Cancel notifications. Add test for new method in TimedNotificationQueue and fix cancel all tests to actually check if notification got executed.
fixes #3986

### DIFF
--- a/Foundation/include/Poco/TimedNotificationQueue.h
+++ b/Foundation/include/Poco/TimedNotificationQueue.h
@@ -89,6 +89,16 @@ public:
 		/// assigned to a Notification::Ptr, to avoid potential
 		/// memory management issues.
 
+	Notification* dequeueNextNotification();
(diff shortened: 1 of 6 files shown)


obiltschnig:
looks good to me

PR #4012, opened 2023-04-19
fix #3979 dangled pointer in AutoPtr assign

### LINKED ISSUE #3979: Check _ptr != ptr is not enough in AutoPtr.assign
https://github.com/pocoproject/poco/blob/38380701469d28c4275b5ac192c8f2afc5c9b65d/Foundation/include/Poco/AutoPtr.h#L122-L127

New object could be one referenced by old with it's internal AutoPtr.
If we `release` old pointer before `duplicate` new we could have dangled pointer problem already.

infdahai:
1. fixes #3979 
2. use `ptr.get() != _ptr` not `&ptr != this`, and avoid the situation where internal pointers are the same

### DIFF
--- a/Foundation/include/Poco/AutoPtr.h
+++ b/Foundation/include/Poco/AutoPtr.h
@@ -119,7 +119,7 @@ public:
 
 	AutoPtr& assign(const AutoPtr& ptr)
 	{
-		if (&ptr != this)
+		if (ptr.get() != _ptr)
 		{
 			if (_ptr) _ptr->release();
 			_ptr = ptr._ptr;


funny-falcon:
To correctly fix #3979 code should call `duplicate()` BEFORE `release().`
See https://github.com/pocoproject/poco/issues/3979#issuecomment-1585545791

aleks-f:
> To correctly fix #3979 code should call `duplicate()` BEFORE `release().` See [#3979 (comment)](https://github.com/pocoproject/poco/issues/3979#issuecomment-1585545791)

@funny-falcon please provide such a fix which is not causing a memory leak and we'll consider it. Otherwise, I'm inclined to reject and close this PR.

funny-falcon:
@aleks-f , I just opened #4068

infdahai:
OK， I close the pr.

PR #4021, opened 2023-04-28
Fix compile with `-DPOCO_NET_NO_IPv6`

gelldur:
Fixed compile with `-DPOCO_NET_NO_IPv6`

### DIFF
--- a/Net/include/Poco/Net/HTTPClientSession.h
+++ b/Net/include/Poco/Net/HTTPClientSession.h
@@ -164,9 +164,10 @@ public:
 	const SocketAddress& getSourceAddress4();
 		/// Returns the last IPv4 source address set with setSourceAddress
 
+#if defined(POCO_HAVE_IPv6)
 	const SocketAddress& getSourceAddress6();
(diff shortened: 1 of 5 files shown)


gelldur:
Failing CI as I can tell is not fault of this change.

gelldur:
@aleks-f this PR was merged to master but on current master those changes do not exist. 
https://github.com/pocoproject/poco/blob/master/Net/src/HTTPClientSession.cpp
I can't even find in [history related changes](https://github.com/pocoproject/poco/commits/master/Net/src/HTTPClientSession.cpp).

Did someone force push?  
Does `-DPOCO_NET_NO_IPv6` is still relevant ?

aleks-f:
@gelldur I am not quite sure anymore how or why I merged this into master - nothing should go there except the releases. There's been quite a bit of discrepancies between release branches  and devel since the 1.11 series and we're still recovering from that - that's the best explanation I have.

Long story short, I am sorry for dropping this one; please send a pull to devel branch and we'll release it in 1.13.1

gelldur:
Thank you, that is all I need. I will resubmit.

PR #4050, opened 2023-06-08
rename arc -> poco_arc

### LINKED ISSUE #3349: Consider renaming ActiveRecord binary from arc
On multiple Linux distributions, there is already binary named `arc` for the ARC file archiver/compressor: http://arc.sourceforge.net/.  It's old and probably not too widely used anymore, but it has been around for a long time in the /usr/bin/arc slot.

Perhaps the poco utility could instead be `poco_arc` or something?

stkw0:
Fix #3349

### DIFF
--- a/ActiveRecord/Compiler/CMakeLists.txt
+++ b/ActiveRecord/Compiler/CMakeLists.txt
@@ -11,7 +11,7 @@ endif()
 add_executable(ActiveRecordCompiler ${SRCS})
 set_target_properties(ActiveRecordCompiler
 	PROPERTIES
-	OUTPUT_NAME arc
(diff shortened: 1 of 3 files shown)


aleks-f:
@stkw0 there is also `Makefile`:

https://github.com/pocoproject/poco/blob/2a6434a86d9fcf3f41e8d442832051cf3ff6263b/ActiveRecord/Compiler/Makefile#L11

and VS projects.

stkw0:
For VS is enough to change Compiler.progen file or it's requires also to `sed -i 's/arc/poco_arc/g' Compiler_vs*` ? Never used VS

PR #4085, opened 2023-07-18
Usage modern C++ features on JSON modules

cngzhnp:
Starting from this module, I want to use modern C++ features such as:

- Use `nullptr` instead of 0.
- Use `auto` instead of iterator type itself.
- Use `default` keyword instead of copy constructor, destructor implementation
- Use `override` keyword if necessary.
- Use `using` keyword instead of `typedef`.
- Use proper functions like `empty()` call instead of `object.size() == 0`

### DIFF
--- a/Foundation/src/DataURIStream.cpp
+++ b/Foundation/src/DataURIStream.cpp
@@ -1,4 +1,4 @@
-﻿//
+//
 // DataURIStream.cpp
 //
 // Library: Foundation
--- a/JSON/src/Array.cpp
+++ b/JSON/src/Array.cpp
@@ -42,14 +42,7 @@ Array::Array(int options):
 }
 
 
-Array::Array(const Array& other) :
-	_values(other._values),
-	_pArray(other._pArray),
-	_modified(other._modified),
-	_escapeUnicode(other._escapeUnicode),
-	_lowercaseHex(other._lowercaseHex)
-{
-}
+Array::Array(const Array& other) = default;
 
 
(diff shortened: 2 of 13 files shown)


teksturi:
This is little bit old PR already but I still answer this. It does not make much sense to change style for just small portion of code base. It would be much better to do just one change at time. Example this

> * Use proper functions like `empty()` call instead of `object.size() == 0`

could be done to whole repo. It would be much more easy to accept those kind of changes.

matejk:
#4354

cngzhnp:
> This is little bit old PR already but I still answer this. It does not make much sense to change style for just small portion of code base. It would be much better to do just one change at time. Example this
> 
> > * Use proper functions like `empty()` call instead of `object.size() == 0`
> 
> could be done to whole repo. It would be much more easy to accept those kind of changes.

Hello @teksturi, sorry for late answer. 

However, my intention was not just a change into small portion of codebase like this, also I am not fan of inconsistent things. I just want to split up code review based on submodules. Creating a PR with 100 files changed which was not easy to review and apply changes. So, doing it step by step would be much better idea at least for me at that time.

andrewauclair:
@matejk did you mean to delete devel? Doesn't seem right.

matejk:
@cngzhnp, would you be willing to rebase this PR to main?

matejk:
PR #4613 was created from these changes. See details there.

Closing this PR.

PR #4086, opened 2023-07-18
Remove std::aligned_storage as it is deprecated in C++23.

Romain-Geissler-1A:
Instead, replace it with std::max_align_t in the Any implementation, as really we would like to be able to store any object with any alignment in the small object optimization case. Typically the size and alignment of std::max_align_t is 8 or 16 on most platforms. Added a static assert to ensure that this change doesn't result in wasting more unused memory in case the size of the storage buffer is smaller than this maximum alignment (which is right now 64, so shall be ok on all platforms).

### DIFF
--- a/Foundation/include/Poco/Any.h
+++ b/Foundation/include/Poco/Any.h
@@ -21,6 +21,7 @@
 #include <algorithm>
 #include <typeinfo>
 #include <cstring>
+#include <cstddef>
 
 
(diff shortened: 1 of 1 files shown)


obiltschnig:
lgtm

Romain-Geissler-1A:
And a typo here as well... The comparison in the static assert should have been in the other direction...

PR #4090, opened 2023-07-18
Spare one byte in the small string optimization of Any. Also fix swap…

Romain-Geissler-1A:
… for big objects (pHolder was never swapped).

Also, I think this may be faster, since you don't need to constantely compare 64 bytes array for `isEmpty` or to memset it in the constructor/destructor/erase.

The idea is that `pHolder` shall always contain the pointer the real content, but in case of object small enough it pointing to the `holder` member directly, rather than a newly allocated buffer. This way we don't need to "store" the information about the fact that the storage is local or not, we get it just by comparing pointers.

Wait for the CI to run for this, as I didn't test in practice. I just found strange this "+ 1" byte when I read the code earlier today.

### DIFF
--- a/Foundation/include/Poco/Any.h
+++ b/Foundation/include/Poco/Any.h
@@ -57,9 +57,7 @@ union Placeholder
 	///
 	/// If Holder<Type> fits into POCO_SMALL_OBJECT_SIZE bytes of storage,
 	/// it will be placement-new-allocated into the local buffer
-	/// (i.e. there will be no heap-allocation). The local buffer size is one byte
-	/// larger - [POCO_SMALL_OBJECT_SIZE + 1], additional byte value indicating
-	/// where the object was allocated (0 => heap, 1 => local).
+	/// (i.e. there will be no heap-allocation).
 	///
 	/// Important: for SOO builds, only same-type (or trivial both-empty no-op)
 	/// swap operation is allowed.
@@ -77,10 +75,8 @@ public:
 
 #ifndef POCO_NO_SOO
 
-	Placeholder(): pHolder(0)
-	{
-		std::memset(holder, 0, sizeof(Placeholder));
-	}
+	Placeholder(): pHolder(nullptr)
+	{}
 
 	~Placeholder()
 	{
@@ -90,7 +86,10 @@ public:
 	void swap(Placeholder& other) noexcept
 	{
 		if (!isEmpty() || !other.isEmpty())
+		{
 			std::swap(holder, other.holder);
+			std::swap(pHolder, other.pHolder);
+		}
 	}
 
 	void erase()
@@ -100,13 +99,12 @@ public:
 
 	bool isEmpty() const
 	{
-		static char buf[SizeV+1] = {};
-		return 0 == std::memcmp(holder, buf, SizeV+1);
+		return !pHolder;
 	}
 
 	bool isLocal() const
 	{
-		return holder[SizeV] != 0;
+		return !pHolder || pHolder == reinterpret_cast<PlaceholderT*>(&holder);
 	}
 
 	template<typename T, typename V,
@@ -114,9 +112,9 @@ public:
 	PlaceholderT* assign(const V& value)
 	{
 		erase();
-		new (reinterpret_cast<PlaceholderT*>(holder)) T(value);
-		setLocal(true);
-		return reinterpret_cast<PlaceholderT*>(holder);
+		pHolder = reinterpret_cast<PlaceholderT*>(&holder);
+		new (pHolder) T(value);
+		return pHolder;
 	}
 
 	template<typename T, typename V,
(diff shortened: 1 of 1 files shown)


andrewauclair:
```cpp
pHolder = reinterpret_cast<PlaceholderT*>(holder);
new (pHolder) T(value);
```
This is almost certainly UB. I can't think of a way it would be valid.

Romain-Geissler-1A:
Maybe, but it as UB as the previous implementation was already. One might want to remove the reinterpret cast by making `holder` a union, made of a pointer and the char buffer.

Note that on my side I don't really plan on continuing working on this pull request. I had to look at this implementation for some reason in the past, I found it a bit strange and proposed this, but I will happily close this pull request if no one on Poco side intends to take it over (as apparently it doesn't work as tests are broken).

aleks-f:
I will look at it, but didn't have time yet. SOO is disabled in the default build and still considered experimental.

andrewauclair:
I'm specifically referring to this line: `pHolder = reinterpret_cast<PlaceholderT*>(holder);`.

> 11.5.1 At most one of the non-static data members of an object of union type can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time[.](https://eel.is/c++draft/class.union.general#2.sentence-2)

Setting one union member as a pointer to another is UB. I tried this for a little while on compiler explorer and always got SIGSEGV.

I also think the current implementation is UB. Not sure there's much to be done about that.

aleks-f:
@andrewauclair I don't think that the standard chapter you are quoting applies in this case - the union holds either value or pointer. I don't recall any sanitizer ever complaining, and those would tell for sure if something is not right.

can you please share the exact code with the current implementation that is segfaulting?

andrewauclair:
> can you please share the exact code with the current implementation that is segfaulting?

Sorry, it wasn't Poco code, it was just some code I was trying based on the changes in this pull request. I'll paste it below. Maybe I'm just completely misunderstanding something here.

```cpp
#include <iostream>

template <typename PlaceholderT, unsigned int SizeV = 32>
union Placeholder {
    PlaceholderT* assign(const PlaceholderT& value)
    {
        pHolder = reinterpret_cast<PlaceholderT*>(holder);
        new (pHolder) PlaceholderT(value);
        return pHolder;
    }

    mutable unsigned char holder[SizeV];
    std::max_align_t           aligner;
    PlaceholderT* pHolder;
};

int main()
{
    Placeholder<double> v;

    v.assign(4.4);

    std::cout << *v.pHolder; // SIGSEGV
}
```

aleks-f:
@andrewauclair I'm not sure what the purpose of the `reinterpret_cast` would be in this example. 

Why not do just:
```c++
// ...
PlaceholderT* assign(const PlaceholderT& value)
{
    new (holder) PlaceholderT(value);
    return reinterpret_cast<PlaceholderT*>(holder);
}
//...

int main()
{
    Placeholder<double> v;
    std::cout << *v.assign(4.4);
}
```

But, more importantly, that is not a representative example of what `Poco::Any` does - `PlaceholderT` is not the held type, it is a placeholder for value of type `T` provided to the `assign` function template:
https://github.com/pocoproject/poco/blob/b61d63da94fe0f716ce3fea49113c9001ba3a67a/Foundation/include/Poco/Any.h#L112

And to retrieve the stored value, one must use `AnyCast<ValueType>` with `ValueType` being the exact type stored:
https://github.com/pocoproject/poco/blob/b61d63da94fe0f716ce3fea49113c9001ba3a67a/Foundation/include/Poco/Any.h#L404

In any case, I got intrigued by the idea that the extra byte can be avoided and I do intend to examine the idea, although I doubt it is feasible - it's been a while, but I couldn't come up with a way to achieve it when I wrote the SOO support.

andrewauclair:
@aleks-f Thank you for the explanation. I think I see where my thought process was going wrong now.

aleks-f:
> The idea is that `pHolder` shall always contain the pointer the real content, but in case of object small enough it pointing to the `holder` member directly, rather than a newly allocated buffer. This way we don't need to "store" the information about the fact that the storage is local or not, we get it just by comparing pointers.

@Romain-Geissler-1A I looked at this proposal. The reason why it doesn't work is because of this:
```c++
bool isLocal() const
{
  return !pHolder || pHolder == reinterpret_cast<PlaceholderT*>(holder);
}
```

in particular, this part:

`pHolder == reinterpret_cast<PlaceholderT*>(holder);`

is always true - `holder` and `pHolder` are two names for the same memory location and their content is always identical to the `sizeof(PlaceholderT*)` extent, so the extra byte is needed to know what is held.

Unless you have something to add here, I'll close this PR

Romain-Geissler-1A:
Mmmmh, re-reading the diff, I don't agree with the above statement.

Let me quickly push an update, and if it's still red, I will close the pull request (as I actually don't use this class, nor Poco in my projects, I had to investigate some Poco code for other projects being developed in my company).

Romain-Geissler-1A:
It doesn't work, I give up. Sorry for the noise.

PR #4097, opened 2023-07-25
Add ability to pass additional compiler settings via powershell build script

chrisbednarski:
I needed to build poco with a custom struct alignment

Added ability to pass additional compiler options to the powershell script

Tested the build process both with and without passing extra compiler options.

Example PS command:
```
$env:CL = "/DWINDOWS_IGNORE_PACKING_MISMATCH"
.\buildwin.ps1 -poco_base . -vs 170 -action build -linkmode all -config both -platform Win32 -omit "Data/MySQL,Data/PostgreSQL" -openssl_base C:\OpenSSL -verbosity normal -compiler_options /Zp1
```

### DIFF
--- a/Data/Data_vs170.vcxproj
+++ b/Data/Data_vs170.vcxproj
@@ -340,6 +340,7 @@
       <TreatWChar_tAsBuiltInType>true</TreatWChar_tAsBuiltInType>
       <ForceConformanceInForLoopScope>true</ForceConformanceInForLoopScope>
       <RuntimeTypeInfo>true</RuntimeTypeInfo>
+      <AdditionalOptions>$(AdditionalCompilerOptions)</AdditionalOptions>
       <PrecompiledHeader/>
       <WarningLevel>Level3</WarningLevel>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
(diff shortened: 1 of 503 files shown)


aleks-f (changes requested):
- [ ] please clean up these redundant commits by syncing with the current devel branch
- [ ] these additional options should be added to earlier versions VS projects (or at least for vs160, and document that the parameter is no-op for earlier versions)

chrisbednarski:
rebased with devel branch

chrisbednarski:
* [x]  cleaned up redundant commits by syncing with the current devel branch
* [x]  additional options added to earlier versions VS projects (vs140-vs160)

aleks-f:
@chrisbednarski there's a [VS2022 projects issue on ARM](https://github.com/pocoproject/poco/issues/4207), so your effort will have to be coordinated with that. Also, in 2022, we should set C++17 as default. Please see if @hmartinez82 will fix that, or you can take care of that issue yourself as a part of this pull

aleks-f:
@chrisbednarski also, bear in mind that these may get overwritten in the future if someone regenerates them using [ProGen](https://github.com/pocoproject/poco/tree/devel/ProGen), so this change should be added there as well

chrisbednarski:
rebased with devel branch

Data/Data_vs170.vcxproj
```diff
@@ -340,6 +340,7 @@
       <TreatWChar_tAsBuiltInType>true</TreatWChar_tAsBuiltInType>
       <ForceConformanceInForLoopScope>true</ForceConformanceInForLoopScope>
       <RuntimeTypeInfo>true</RuntimeTypeInfo>
+      <AdditionalOptions>$(AdditionalCompilerOptions)</AdditionalOptions>
```
aleks-f, line 343: @chrisbednarski all of these `AdditionalOptions` for vs160 and 170 are redundant;
https://github.com/pocoproject/poco/blob/4a9285c997234bc216657ec8f1ce1b23b66fd223/Data/Data_vs170.vcxproj#L348

vs140 and 150 will be [removed](https://github.com/pocoproject/poco/issues/4274) for 1.13.0 release

aleks-f (changes requested):

PR #4103, opened 2023-08-01
Fix openssl session resumption, FTPS certificate validation vs hostname

chrisbednarski:
- `_socketIndex` was uninitialised in `SSLManager`, which meant `SSL_get_ex_data` in `SecureSocketImpl::onSessionCreated` always returned NULL
  - session resumption was not working because of this
  - I considered adding `CRYPTO_free_ex_index` to the `SSLManager` destructor, but poco does not pass any callback pointers to `SSL_get_ex_new_index`, so there is no need to clear them
- pass host name when attaching the secure socket stream
  - without this, the FTPS client tries to validate the certificate against the IP address of the server
- add `setQuietShutdown` and `ignoreUnexpectedEof` functions to the openssl context
  - bidirectional shutdown on Microsoft FTPS **upload** connections is timing out. The quiet shutdown option marks the connection as closed and does not try to receive the close_notify alert (which was timing out)
  - bidirectional shutdown on filezilla FTPS connections works well (after the session resumption was fixed)


chrisbednarski:
I'm looking at the failing tests

chrisbednarski:
rebased

matejk (commented):
I approve this merge request.

@aleks-f: any objections to merge to devel?

Issue #2291, opened 2018-04-13
Visitor Pattern for Dynamic::Var


obiltschnig:
Implement the Visitor pattern for `Dynamic::Var` and `Dynamic::VarHolder`.

Main issue here is how to deal with unknown types. A simple solution would be to just deal with a limited set of types (basically all for which we have a VarHolder, and simply throw a Poco::NotImplementedException in `VarHolder::accept()`. Will probably be sufficient for my initial use case. 

It would be interesting to implement `accept()` as a template, with the Visitor class as template argument. This would make it possible to provide customized Visitor classes, containing `visit()` methods for all used types. However, for obvious reasons C++ does not support virtual member function templates.

A different idea would be a "dynamic" Visitor class with the ability to register "visit" functions for a specific type as std::function objects or lambdas. Something to consider for 2.0.

PR #4144, opened 2023-09-09
add visitor pattern implementation for Poco::Dynamic::Var

bas524:
Complimentary for #2291
It's a simple implementation of runtime visitor
Main idea is using unique handler for every type which can be saved in the Dynamic::Var

As handler you can use lambda, std::function and function pointer with signature:
```cpp
std::function<void(const T&)>;
void (*)(const T &);
```

example
```cpp
// ...
Visitor visitor;
visitor.addHandler<std::string>([](const std::string& s) { std::cout << s; });
// ...
Poco::Dynamic::Var v = getVar();
if (visitor.visit(v))
{
    // successfully processed
}
```

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarVisitor.h
+++ b/Foundation/include/Poco/Dynamic/VarVisitor.h
@@ -1,0 +1,99 @@
+#ifndef Foundation_VarVisitor_INCLUDED
+#define Foundation_VarVisitor_INCLUDED
+
+//
+// VarVisitor.h
+//
+// Library: Foundation
+// Package: Dynamic
+// Module:  VarVisitor
+//
+// Definition of the VarVisitor class.
+//
+// Copyright (c) 2023, Applied Informatics Software Engineering GmbH.
+// and Contributors.
+//
+// SPDX-License-Identifier:	BSL-1.0
+//
+
+
+#include "Poco/Dynamic/Var.h"
+#include <unordered_map>
+#include <functional>
+
+namespace Poco {
+namespace Details {
+
+struct TypeInfoHash
+{
+	inline std::size_t operator()(std::type_info const& t) const { return t.hash_code(); }
+};
+
+struct EqualRef
+{
+	template <typename T>
+	bool operator()(std::reference_wrapper<T> a, std::reference_wrapper<T> b) const
+	{
+		return a.get() == b.get();
+	}
+};
+
+using TypeInfoRef = std::reference_wrapper<std::type_info const>;
+
+using HandlerCaller = std::function<void(const Poco::Dynamic::Var&)>;
+
+template <typename T>
+using HandlerPointer = void (*)(const T &);
+
+template <typename T>
+using Handler = std::function<void(const T&)>;
+
+}	// Details
+
+namespace Dynamic {
+
+class Foundation_API Visitor
+	/// VarVisitor class.
+{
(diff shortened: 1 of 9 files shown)


Foundation/include/Poco/Dynamic/VarVisitor.h
```diff
@@ -0,0 +1,99 @@
...
+	
+public:
+	template <typename T>
+	bool addHandler(const Details::Handler<T> &f)
+	/// Add handler for specific type T which holds in Var
+	/// This method is more safe, because it saves copy of handler : lambda or std::function
+	/// Returns true if handler was added
+	{
+		auto result = _handlers.emplace(std::ref(typeid(T)),
+										Details::HandlerCaller([handler = f](const Poco::Dynamic::Var& x)
```
aleks-f, line 71: too much indent (3 is enough)

Foundation/include/Poco/Dynamic/VarVisitor.h
```diff
@@ -0,0 +1,99 @@
...
+	}
+	
+	template <typename T>
+	bool addHandler(Details::HandlerPointer<T> f)
+	/// Add handler for specific type T which holds in Var
+	/// This method is less safe, because it saves only copy of function pointer
+	/// Returns true if handler was added
+	{
+		auto result = _handlers.emplace(std::ref(typeid(T)),
+										Details::HandlerCaller([handlerPointer = f](const Poco::Dynamic::Var& x)
```
aleks-f, line 85: too much indent (3 is enough)

Foundation/src/VarVisitor.cpp
```diff
@@ -0,0 +1,32 @@
...
+//
+
+#include "Poco/Dynamic/VarVisitor.h"
+
+namespace Poco {
+namespace Dynamic {
+
+bool Visitor::visit(const Var &x) const
+{
+	bool wasHendled = false;
```
aleks-f, line 22: spelling, wasHandled, not hendled

Foundation/testsuite/src/VarTest.cpp
```diff
@@ -3103,6 +3110,169 @@ void VarTest::testSharedPtr()
 	assertTrue(p.referenceCount() == 1);
 }
 
+struct ProcessDummy
+{
+	Var &v;
+	ProcessDummy(Var &var) : v(var) {}
+	void operator()(const Dummy &d) {
```
aleks-f, line 3117: style

aleks-f (changes requested):
@bas524 we need that CI passing. rebase your branch with `devel` to get the most recent changes and I will restart any failing CI (sometimes they just timeout) that does not show a serious bug

Foundation/testsuite/src/VarTest.cpp
bas524: done

Foundation/src/VarVisitor.cpp
bas524: done

Foundation/include/Poco/Dynamic/VarVisitor.h
bas524: done

Foundation/include/Poco/Dynamic/VarVisitor.h
bas524: done

PR #4156, opened 2023-09-21
Allow passing raw fd's into ServerSocket

russelltg:
I'm writing an application that receives a fd from systemd that's bound on port 80, and I would like Poco to be able to host the server on the fd that I receive. 

This code enables this usecase. I've tested it and it works well


russelltg:
I'm marking this as draft as I realize that it's a breaking change due to the very similar constructor with port.

Maybe instead of a constructor this case should be supported with constructing a default instance then calling a .useFd function or something?

russelltg:
Ok, I think this is a better approach to avoid issues with implicit integer conversion and to make this API use a bit more explicit at callsites.

Thanks for reviewing :)

aleks-f (changes requested):
add a test case [here](https://github.com/russelltg/poco/blob/fd_into_serversocket/Net/testsuite/src/SocketTest.cpp), please

russelltg:
I've added a test. I'm not on windows at the moment so haven't tested that the windows portion builds/runs, but it works on Linux.

matejk:
Code compiles on macOS (clang 15.0.0), unit tests pass.

Net/include/Poco/Net/SocketImpl.h
```diff
@@ -142,6 +142,13 @@ class Net_API SocketImpl: public Poco::RefCountedObject
 		/// If the library has not been built with IPv6 support,
 		/// a Poco::NotImplementedException will be thrown.
 
+	void useFileDescriptor(poco_socket_t fd);
+		/// Use a external file descriptor for the socket. Required to be careful
+		/// about what kind of file descriptor you're passing to make sure it's compatable
```
matejk, line 147: Typo: compatable -> compatible?

matejk (commented):
I approve the changes with only one minor comment from my side.

PR #4190, opened 2023-10-16
CMake: Use CMAKE_INSTALL_* variables from GNUInstallDirs

matejk:
Poco includes GNUInstallDirs (#3105), however the provided variables are not used in all places.

Module cmake files that are installed to /usr/lib/x86_64-linux-gnu/cmake/Poco on Ubuntu or Debian, can refer to non-existent libraries.

This PR replaces usage of hardcoded paths with variables from GNUInstallDirs.


tuduongquyet:
Voted

PR #4223, opened 2023-10-25
Virtualize ServerApplication::handlePidFile()

pavledragisic:


obiltschnig:
Why doesn't `replaceOption()` take an `Option` parameter, like `addOption()`? Seems inconsistent.

aleks-f:
> Why doesn't `replaceOption()` take an `Option` parameter, like `addOption()`? Seems inconsistent.

@pavledragisic change it, but it also requires changes on our side (devs, dispatcher and s7server) so either align it on our side or create an issue there

obiltschnig:
Then at least also add an `addOption()` overload, please, so that the API stays consistent.

aleks-f:
> Then at least also add an `addOption()` overload, please, so that the API stays consistent.

alright, @pavledragisic do it like that

aleks-f:
@obiltschnig after a bit more thinking, how about this instead:
```c++
template <typename T>
void OptionSet::replaceHandler(const std::string& opt, T* pObj, Callback<T> func)
```

That is really what we need - to tell the `Application` that we will handle this option and replacing the whole Option is the proverbial cannon to kill a mosquito :)

aleks-f:
> If that's all you need, then how about just making `handlePidFile()`, `handleDaemon()`, etc. virtual?

virtualizing will work; at this time, I would limit it to handlePidFile only and I would also like to move it out of #ifdef FAMILY_UNIX, so that it is available on all platforms

aleks-f:
> > If that's all you need, then how about just making `handlePidFile()`, `handleDaemon()`, etc. virtual?
> 
> virtualizing will work; at this time, I would limit it to handlePidFile only and I would also like to move it out of #ifdef FAMILY_UNIX, so that it is available on all platforms

@pavledragisic do it like this ^
https://github.com/pocoproject/poco/blob/e47b92d641d3efffe9a81cf8f4a940f7267a015b/Util/include/Poco/Util/ServerApplication.h#L174
discard current changes in this branch, move this function out of #ifdef, and make it virtual. We then have to override it in our processes that require strict PID creation timing

PR #4255, opened 2023-11-08
4248 add progen powershell script

cunj123:


progen.ps1
```diff
@@ -0,0 +1,109 @@
...
+	[switch] $help
+)
+
+function Process-Input
+{
+	if ($help -eq $true)
+	{
+		Write-Host 'Usage:'
+		Write-Host '------'
+		Write-Host 'progen.ps1 [-poco_base <dir>]'
```
aleks-f, line 31: there should also be `-samples` and `-tests` (default off)

aleks-f (changes requested):

Issue #4307, opened 2023-11-28
`NotificationCenter` data race


aleks-f:
**Describe the bug**
All observers container access is protected by mutex, except access by destructor destroying the container.

**To Reproduce**
Thread sanitizer

**Expected behavior**
Clean destruction

**Logs**
```
WARNING: ThreadSanitizer: data race (pid=70947)
  Write of size 8 at 0x0001070011c8 by main thread:
// ...
    #11 Poco::NotificationCenter::~NotificationCenter() NotificationCenter.cpp:31 (libPocoFoundationd.100.dylib:arm64+0x104418)
    #12 Poco::TaskManager::~TaskManager() TaskManager.cpp:52 (libPocoFoundationd.100.dylib:arm64+0x1cb4a0)
// ...
Previous read of size 8 at 0x0001070011c8 by thread T11 (mutexes: write M0):
// ...
#8 Poco::NotificationCenter::postNotification(Poco::AutoPtr<Poco::Notification>) NotificationCenter.cpp:72 (libPocoFoundationd.100.dylib:arm64+0x1050bc)
// ..
```

**Please add relevant environment information:**
osx clang

PR #4312, opened 2023-11-28
4307/8/9/10 data races

aleks-f:


Foundation/src/NotificationCenter.cpp
```diff
@@ -29,6 +29,14 @@ NotificationCenter::NotificationCenter()
 
 NotificationCenter::~NotificationCenter()
 {
+	try
+	{
+		Mutex::ScopedLock lock(_mutex);
+	}
```
matejk, line 35: Since `_mutex` protects `_observers` was it meant that list of observers is cleared while holding the mutex in scoped lock?

Otherwise I do not see the point of having the lock in place.

Foundation/include/Poco/Task.h
```diff
@@ -98,6 +98,8 @@ class Foundation_API Task: public Runnable, public RefCountedObject
 		/// If task has been cancelled prior to this call, it only sets
 		/// the state to TASK_FINISHED and notifies the owner.
 
+	bool hasOwner() const;
+		/// Returns true iff the task has an owner.
```
matejk, line 102: Typo: iff

matejk (commented):
Reviewed PR. There is one point that is not clear to me why it is done the way it is.

Foundation/src/NotificationCenter.cpp
aleks-f: yes, this was reported by clang thread sanitizer #4307 
This fix is not pretty and not even bulletproof but I guess it narrows the race opportunity somewhat; an alternative could be making `_observers` a `SharedPtr`, and then temporarily holding a reference to it in all accessors

Foundation/include/Poco/Task.h
aleks-f: actually, it is [intentional](https://www.merriam-webster.com/dictionary/iff)

Foundation/src/NotificationCenter.cpp
matejk: What about having it like this?

```
NotificationCenter::~NotificationCenter()
{
	try
	{
		Mutex::ScopedLock lock(_mutex);
		for (auto& o: _observers)
			o->disable();

		_observers.clear();
	}
	catch(...)
	{
		poco_unexpected();
	}
}

```

aleks-f: yes, that's fine. you can change it and merge this PR

Foundation/include/Poco/Task.h
matejk: One can learn something new every day. 😀

PR #4330, opened 2023-12-09
Stricter DateTimeParser for known formats (fixed #569)

### LINKED ISSUE #569: SyntaxException for DateTimeParser::parse not working
From http://stackoverflow.com/questions/25285740/syntex-exception-for-datetimeparserparse-of-poco-not-working :

In the documentation to Poco::DateTimeParser:.parse(const
std::string& fmt, const std::string& str, int& timeZoneDifferential)

Parses a date and time in the given format from the given string.
Throws a SyntaxException if the string cannot be successfully parsed.

As I understand this sentence the given datestring "str" must be in
the format described in paramter "fmt", otherwise a SyntaxException is thrown - or do I miss something ?

But even if i give wrong format like "sjdcg-12987-kjchsjkc" or an empty format. its parsing the string and giving some garbage value.

How can i solve this issue:

the code is like this-

string datetime="jnghjgnbcfjb";

```
try
{
  DateTimeParser::parse(DateTimeFormat::ISO8601_FRAC_FORMAT,datetime,dt,tzd);
}
catch(const Poco::Exception& exc)
{
  cout << exc.displayText() << endl;
}
```

matejk:
Cherry picked fix from devel-experimental with a few additional adaptations for current state of devel and few improvements.


Foundation/testsuite/src/DateTimeParserTest.cpp
```diff
@@ -758,6 +840,18 @@ void DateTimeParserTest::testParseDayOfWeek()
...
 
+void DateTimeParserTest::testBad(const std::string& fmt, const std::string& dateStr, int tzd)
+{
+	try
+	{
+		DateTime dt;
+		DateTimeParser::parse(fmt, dateStr, dt, tzd);
+		fail ("must fail: "s + fmt + ", " + dateStr);
+	}
+	catch(const Poco::Exception) { }
```
aleks-f, line 851: catch by reference to silence CodeQL

aleks-f (changes requested):

PR #4333, opened 2023-12-11
enh(NumberFormatter): Introduce backward compatible options for formatHex functions

### LINKED ISSUE #3667: NumberFormatter: add Options enum for controlling prefix and lowercase
Following up the discussion in #3657, let's fix this before 1.12 release.

matejk:
Fixes #3667 and replaces #3657.


aleks-f:
@matejk @bas524 apparently, the sendFile pull did not have the VS projects regenerated, so we have CI failures not. That's why I asked for a new pull to trigger build and testrun. anyway, all affected VS projects should be regenerated with [progen](https://github.com/pocoproject/poco/wiki/Generating-VS-Projects-using-ProGen)

```
     Creating library ..\lib64\PocoNet.lib and object ..\lib64\PocoNet.exp
SocketImpl.obj : error LNK2019: unresolved external symbol TransmitFile referenced in function "public: __int64 __cdecl Poco::Net::SocketImpl::sendFile(class Poco::FileInputStream &,unsigned __int64)" (?sendFile@SocketImpl@Net@Poco@@QEAA_JAEAVFileInputStream@3@_K@Z) [D:\a\poco\poco\Net\Net_vs170.vcxproj]
..\bin64\PocoNet64.dll : fatal error LNK1120: 1 unresolved externals [D:\a\poco\poco\Net\Net_vs170.vcxproj]
Error: Final attempt failed. Child_process exited with error code 1
```

matejk:
> @matejk @bas524 apparently, the sendFile pull did not have the VS projects regenerated, so we have CI failures not. That's why I asked for a new pull to trigger build and testrun. anyway, all affected VS projects should be regenerated with [progen](https://github.com/pocoproject/poco/wiki/Generating-VS-Projects-using-ProGen)
> 
> ```
>      Creating library ..\lib64\PocoNet.lib and object ..\lib64\PocoNet.exp
> SocketImpl.obj : error LNK2019: unresolved external symbol TransmitFile referenced in function "public: __int64 __cdecl Poco::Net::SocketImpl::sendFile(class Poco::FileInputStream &,unsigned __int64)" (?sendFile@SocketImpl@Net@Poco@@QEAA_JAEAVFileInputStream@3@_K@Z) [D:\a\poco\poco\Net\Net_vs170.vcxproj]
> ..\bin64\PocoNet64.dll : fatal error LNK1120: 1 unresolved externals [D:\a\poco\poco\Net\Net_vs170.vcxproj]
> Error: Final attempt failed. Child_process exited with error code 1
> ```

I'll take care first thing in the morning since I messed up.

matejk:
@aleks-f, are there any comments from your side on this PR or can I merge to devel?

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -43,6 +43,15 @@ class Foundation_API NumberFormatter
 		FMT_ON_OFF
 	};
 
+	enum class Options
+	{
+		OPT_NONE = 0,
+		OPT_PREFIX = 1 << 0,
+			/// Prepend prefix 0x
+		OPT_LOWERCASE = 1 << 1
+			/// Use uppercase letters for hexadecimal values
```
teksturi, line 52: OPT_LOWERCASE and comment says "Use UPPERCASE" ???

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -43,6 +43,15 @@ class Foundation_API NumberFormatter
 		FMT_ON_OFF
 	};
 
+	enum class Options
```
teksturi, line 46: Also add comment here to explain what this means.

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -573,17 +737,33 @@ inline std::string NumberFormatter::format0(unsigned int value, int width)
...
+	return formatHex(value, prefix ? Options::OPT_PREFIX : Options::OPT_NONE);
+}
+
+
+inline std::string NumberFormatter::formatHex(unsigned value, Options options)
 {
 	std::string result;
-	intToStr(value, 0x10, result, prefix);
+	const bool prefix { (options & Options::OPT_PREFIX) != Options::OPT_NONE};
+	const bool lowercase { (options & Options::OPT_LOWERCASE) != Options::OPT_NONE};
```
teksturi, line 749: Just asking about styling. Do Podo now use {} initializing? Still if can be used which I like is it like this that bracket is seperated from variable name? "lowercase{" vs "lowecase {" also I think at least last } should also include space "Options::OPT_NONE}" vs "Options::OPT_NONE }"

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -820,17 +1055,33 @@ inline std::string NumberFormatter::format0(UInt64 value, int width)
...
+}
+
+
+inline std::string NumberFormatter::formatHex(UInt64 value, int width, Options options)
 {
 	std::string result;
-	intToStr(value, 0x10, result, prefix, width, '0');
+	const bool prefix { (options & Options::OPT_PREFIX) != Options::OPT_NONE};
+	const bool lowercase { (options & Options::OPT_LOWERCASE) != Options::OPT_NONE};
+	intToStr(value, 0x10, result, prefix, width, '0', 0, lowercase);
```
teksturi, line 1084: I'm totally ok that these are in header with inline as these really get optimized away basically. How ever I wonder should we move deprecated to cpp file. We do not care those so much as they are really deprecated. This heder get's quite big because of those.

In the end totally your call.

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -190,17 +272,32 @@ class Foundation_API NumberFormatter
...
+
+	[[deprecated("use formatHex with options instead")]]
+	static std::string formatHex(unsigned long long value, int width, bool prefix);
 		/// Formats a 64-bit integer value in hexadecimal notation,
 		/// right justified and zero-padded in a field having at least
 		/// the specified width. If prefix is true, "0x" prefix is
 		/// prepended to the resulting string.
 
+	static std::string formatHex(unsigned long long value, int width, Options options = Options::OPT_NONE);
+		/// Formats a 64-bit integer value in hexadecimal notation,
```
teksturi, line 294: In some places we it is documented just that use integer value but in some places we have complete type. Which one it should be?

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -59,20 +68,36 @@ class Foundation_API NumberFormatter
 		/// right justified and zero-padded in a field
 		/// having at least the specified width.
 
-	static std::string formatHex(int value, bool prefix = false);
+	[[deprecated("use formatHex with options instead")]]
+	static std::string formatHex(int value, bool prefix);
```
teksturi, line 72: Can we move all deprecated declarations to end of class definition? This way it is nicer to see this class as first are the ones which we should use.

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -59,20 +68,36 @@ class Foundation_API NumberFormatter
...
 
-	static std::string formatHex(int value, int width, bool prefix = false);
+	static std::string formatHex(int value, Options options = Options::OPT_NONE);
+		/// Formats an int value in hexadecimal notation.
+		/// Options (see NumberFormatter::Options) define the format of the
+		/// generated string.
+		/// The value is treated as unsigned.
+
+	[[deprecated("use formatHex with options instead")]]
+	static std::string formatHex(int value, int width, bool prefix);
```
teksturi, line 85: So why do you remove default prefix false? That change will break things if someone is using these functions.

teksturi: Aa I see it now. That was probably issue in first place. Because those are ambiguous with wrong usage.

Foundation/include/Poco/NumberFormatter.h
aleks-f: what we should have for cases like this is style checking with clang-tidy or some similar tool. lots of time gets wasted on style violations that could be resolved before code is pushed

Foundation/include/Poco/NumberFormatter.h
aleks-f: yes, these should probably go to cpp file. a rule of thumb is single liners and templates in headers, everything else in cpp

Foundation/include/Poco/NumberFormatter.h
aleks-f: I agree - deprecated functions' reason for existence is precisely not to break code and give people time to adjust

aleks-f (changes requested):

Foundation/include/Poco/NumberFormatter.h
teksturi: Definitly! That is basically most time saving thing what can be done. In companies but also in open source. Style checking is done with clang-format. Needs little bit hacking to get it work with everything but it is worth a time.

Foundation/include/Poco/NumberFormatter.h
matejk: I removed the default prefix because there is ambiguity and compile fails.

```
[[deprecated("use formatHex with options instead")]]
static std::string formatHex(int value, int width, bool prefix = false);
static std::string formatHex(int value, int width, Options options = Options::OPT_NONE);

```
Default was no prefix which is also the default when using options. The behaviour for the user if it did not use the `prefix` does not change. Deprecation warning is issues only for users that explicitly used bool prefix argument.

```
/Users/matejk/git/github/poco/Foundation/src/Logger.cpp:252:18: error: call to 'formatHex' is ambiguous
                message.append(NumberFormatter::formatHex(addr, 4));
                               ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/matejk/git/github/poco/Foundation/include/Poco/NumberFormatter.h:704:37: note: candidate function
inline std::string NumberFormatter::formatHex(int value, int width, bool prefix)
                                    ^
/Users/matejk/git/github/poco/Foundation/include/Poco/NumberFormatter.h:710:37: note: candidate function
inline std::string NumberFormatter::formatHex(int value, int width, Options options)
                                    ^
```

Foundation/include/Poco/NumberFormatter.h
matejk: I reused the text that was there before but I agree: it can be improved.

Foundation/include/Poco/NumberFormatter.h
matejk: I intentionally left them where they were before for the users to easily see the new functions that they should use instead. I can move them to the end of the class, no problem.

Foundation/include/Poco/NumberFormatter.h
matejk: I did not see a rule against brace initialization in coding style (or I missed it). Will change to parentheses.

Foundation/include/Poco/NumberFormatter.h
matejk: Resolved in new commit.

Foundation/include/Poco/NumberFormatter.h
matejk: Resolved in new commit.

Foundation/include/Poco/NumberFormatter.h
matejk: Resolved in new commit.

Foundation/include/Poco/NumberFormatter.h
matejk: Moved deprecated functions to cpp file.

Foundation/include/Poco/NumberFormatter.h
matejk: Resolved in new commit.

Foundation/include/Poco/NumberFormatter.h
matejk: Resolved in new commit.

matejk:
@teksturi and @aleks-f, thank you very much for your reviews.

I pushed corrected code. Can you please take a look if it is OK to merge?

Foundation/include/Poco/NumberFormatter.h
teksturi: Aa true. Nice

Foundation/include/Poco/NumberFormatter.h
```diff
@@ -43,6 +43,17 @@ class Foundation_API NumberFormatter
...
 	};
 
+	enum class Options
+		/// Option to control the format of the generated string.
+	{
+		OPT_DEFAULT = 0,
+			/// formatHex defaults: No 0x prefix, uppercase hexadecimal values
+		OPT_PREFIX = 1 << 0,
+			/// formatHex: Prepend prefix 0x
+		OPT_LOWERCASE = 1 << 1
```
teksturi, line 53: I still think if it is good to have OPT_ prefix. At least I do not use those with enum classes. Those where totally different when just enum where used.

```
Options::OPT_DEFAULT;
Options::DEFAULT;

Options::OPT_PREFIX;
Options::PREFIX;

Options::OPT_LOWERCASE;
Options::LOWERCASE;
```

And this is nitpicking and I leave it up to you to choose which you prefer. My vote go still without OPT_. Everything looks very good. Nice job.

teksturi (approved):

Foundation/include/Poco/NumberFormatter.h
matejk: Renamed. Now it is the time to do it right. It will not be changed soon again.

PR #4334, opened 2023-12-11
Improve c++17 support

teksturi:
Poco now use C++17. We can take some old code away because of that. We
also raise requiments for C++17 so everything works as excpected with
it.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -42,7 +42,7 @@
             "forcedInclude": [],
             "compilerPath": "/usr/bin/clang",
             "cStandard": "c11",
-            "cppStandard": "c++14",
+            "cppStandard": "c++17",
             "compileCommands": "",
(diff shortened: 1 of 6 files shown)


Foundation/src/double-to-string.cc
```diff
@@ -56,15 +56,15 @@ bool DoubleToStringConverter::HandleSpecialValues(
     StringBuilder* result_builder) const {
```
aleks-f, line 56: we do not fix embedded dependencies directly. Instead, you should report it upstream

aleks-f (changes requested):

Foundation/src/double-to-string.cc
teksturi: Removed that commit

matejk (approved):
I agree with the changes.

aleks-f (approved):

Issue #4328, opened 2023-12-07
Environment::nodeId Should Throw SystemException When Node ID is 0


aleks-f:
### Discussed in https://github.com/pocoproject/poco/discussions/4326

<div type='discussions-op-text'>

<sup>Originally posted by **andrewauclair** December  7, 2023</sup>
We've had a few issues with this recently. If the Environment_UNIX.cpp EnvironmentImpl::nodeIdImpl fails to find a node ID it simply leaves the node ID as all 0's. A similar thing appears to happen on Windows. We then assume that this is a valid ID and use it. Some parts of our system end up getting 0's and other parts get a valid node ID and fail to communicate.

I think that this would work better if a SystemException was thrown when nodeIdImpl fails to set a value.</div>

aleks-f:
> I think that this would work better if a SystemException was thrown when nodeIdImpl fails to set a value.

That's actually what the documentation says, but it should throw if it can not produce `nodeId` for an reason.

@andrewauclair please create a PR, and @matejk and myself will review

obiltschnig:
`EnvironmentImpl::nodeIdImpl()` should be changed to also look for WiFi adapters if no Ethernet adapter was found.

PR #4336, opened 2023-12-11
Throw exceptions when node ID retrieval is unsuccessful and support Wireless Adapters on Windows

andrewauclair:
Changes for issue #4328. Throwing exceptions for nodeIdImpl on Linux and adding support for Wireless Adapters on Windows.


Foundation/src/Environment_WIN32U.cpp
```diff
@@ -209,31 +209,65 @@ void EnvironmentImpl::nodeIdImpl(NodeId& id)
...
-	}
-	if (GetAdaptersInfo(pAdapterInfo, &len) == NO_ERROR)
+
+	try
 	{
-		pAdapter = pAdapterInfo;
+		if (rc == ERROR_BUFFER_OVERFLOW)
+		{
+			delete[] reinterpret_cast<char*>(pAdapterInfo);
+			pAdapterInfo = reinterpret_cast<IP_ADAPTER_INFO*>(new char[len]);
```
teksturi, line 218: If new throws now we get double free. These things are always hard to check. Why won't we just use smart pointer for pAdapterInfo so we do not need to think how these kind of things go.

Foundation/src/Environment_WINCE.cpp
```diff
@@ -153,45 +153,71 @@ std::string EnvironmentImpl::nodeNameImpl()
...
-		delete[] reinterpret_cast<char*>(pAdapterInfo);
-		throw SystemException("cannot get network adapter list");
-	}
+
 	try
 	{
+		if (rc == ERROR_BUFFER_OVERFLOW)
+		{
+			delete[] reinterpret_cast<char*>(pAdapterInfo);
+			pAdapterInfo = reinterpret_cast<IP_ADAPTER_INFO*>(new char[len]);
```
teksturi, line 171: Here we have same problem as in WIN32U.cpp

Foundation/src/Environment_WIN32U.cpp
andrewauclair: > If new throws we get double free

I'm not seeing a way that would happen. The catch is only for Poco `SystemException`s and if another exception is thrown, the final `delete[]` won't execute.

teksturi: Yeah mistake from my part that I did not checked expection type close enogh.

Foundation/src/Environment_WINCE.cpp
```diff
@@ -153,45 +153,71 @@ std::string EnvironmentImpl::nodeNameImpl()
...
+		}
+
+		// ethernet and wifi adapters not found, fail the search
+		if (!found)
+		{
+			throw SystemException("No ethernet or wifi adapter found");
+		}
 	}
-	catch (Exception&)
+	catch (SystemException&)
```
matejk, line 218: Catch by const reference.

Foundation/src/Environment_WIN32U.cpp
```diff
@@ -209,31 +209,65 @@ void EnvironmentImpl::nodeIdImpl(NodeId& id)
...
+			throw SystemException("cannot get network adapter list");
 		}
+
+		// ethernet and wifi adapters not found, fail the search
+		if (!found)
+		{
+			throw SystemException("No ethernet or wifi adapter found");
+		}
+	}
+	catch (SystemException&)
```
matejk, line 265: Catch by const reference

Foundation/src/Environment_WIN32U.cpp
```diff
@@ -209,31 +209,65 @@ void EnvironmentImpl::nodeIdImpl(NodeId& id)
...
+		}
+		else
+		{
+			throw SystemException("cannot get network adapter list");
 		}
+
+		// ethernet and wifi adapters not found, fail the search
+		if (!found)
+		{
+			throw SystemException("No ethernet or wifi adapter found");
```
matejk, line 262: Some exception strings start with capital word, some don't. IMO they shall be unified. Majority of strings used with SystemException are lower case.

Foundation/src/Environment_WIN32U.cpp
```diff
@@ -209,31 +209,65 @@ void EnvironmentImpl::nodeIdImpl(NodeId& id)
...
-		pAdapterInfo = reinterpret_cast<IP_ADAPTER_INFO*>(new char[len]);
-	}
-	else if (rc != ERROR_SUCCESS)
-	{
-		delete[] reinterpret_cast<char*>(pAdapterInfo);
-		throw SystemException("cannot get network adapter list");
-	}
-	if (GetAdaptersInfo(pAdapterInfo, &len) == NO_ERROR)
+
+	try
```
teksturi, line 213: Still consider doing something like: (not tested)

```c++
	ULONG len = sizeof(IP_ADAPTER_INFO);

	std::unique_ptr<char[]> buffer = std::make_unique<char[]>(len);
	pAdapterInfo = reinterpret_cast<IP_ADAPTER_INFO*>(buffer.get());

	// Make an initial call to GetAdaptersInfo to get
	// the necessary size into len
	DWORD rc = GetAdaptersInfo(pAdapterInfo, &len);
	if (rc == ERROR_BUFFER_OVERFLOW)
	{
		buffer = std::make_unique<char[]>(len);
		pAdapterInfo = reinterpret_cast<IP_ADAPTER_INFO*>(buffer.get());
	}
```

That way you do not need try catch at all.

andrewauclair: I was trying to avoid new things for this PR. I only see `std::make_unique` in a couple of the newer Poco modules. The bigger issue here is the static alias violations jumping between char* and IP_ADAPTER_INFO*. I think I have a fix for that part.

Foundation/src/Environment_WIN32U.cpp
andrewauclair: Changed this one and the ones in Environment_UNIX.cpp to lower case.

Foundation/src/Environment_WIN32U.cpp
```diff
@@ -202,38 +202,65 @@ void EnvironmentImpl::nodeIdImpl(NodeId& id)
...
+
 		while (pAdapter && !found)
 		{
 			if (pAdapter->Type == MIB_IF_TYPE_ETHERNET && pAdapter->AddressLength == sizeof(id))
 			{
-				found = true;
 				std::memcpy(&id, pAdapter->Address, pAdapter->AddressLength);
+
+				// found an ethernet adapter, we can return now
+				return;
```
teksturi, line 234: You now return straight away as you removed found. That is ok as now this follow RAII.
However you still have bool found. That can also removed as state is never changed.

So code will simplyfie to this

```c++
	if (GetAdaptersInfo(pAdapterInfo.get(), &len) != NO_ERROR)
	{
		throw SystemException("cannot get network adapter list");
	}

	IP_ADAPTER_INFO* pAdapter = pAdapterInfo.get();
	while (pAdapter)
	{
		if (pAdapter->Type == MIB_IF_TYPE_ETHERNET && pAdapter->AddressLength == sizeof(id))
		{
			std::memcpy(&id, pAdapter->Address, pAdapter->AddressLength);

			// found an ethernet adapter, we can return now
			return;
		}
		pAdapter = pAdapter->Next;
	}

	// if an ethernet adapter was not found, search for a wifi adapter
	pAdapter = pAdapterInfo.get();
	while (pAdapter)
	{
		if (pAdapter->Type == IF_TYPE_IEEE80211 && pAdapter->AddressLength == sizeof(id))
		{
			std::memcpy(&id, pAdapter->Address, pAdapter->AddressLength);

			// found a wifi adapter, we can return now
			return;
		}
		pAdapter = pAdapter->Next;
	}

	throw SystemException("no ethernet or wifi adapter found");

```

andrewauclair: Thought I had already done that. I guess not. It is removed now.

teksturi (approved):
To my eayes everything looks good.

PR #4339, opened 2023-12-12
Fix RemoteSyslogChannel setProperty value check

teksturi:
In RemoteSyslogChannel::setProperty we compare just 4 first letters. So if value would have example be syslevel then this function would work wrong. Nothing major but nice to fix.


teksturi:
Kinda better fix would be to make also iStartsWith and iEndsWith. This kind of situation would not happen with those. If wanted I can do those also and use those where needed.

Issue #2967, opened 2020-03-29
build Poco Net failed MinGW [TIMESTAMP_REQUEST enum vs macro]


sk-landry:
Good morning, 

I'm building POCO from Github's master branch. (commit = 3fc3e5f5b8462f7666952b43381383a79b8b5d92) 
I use MinGW 7.3 32bits bundled in QT. I am on Windows 10.
The build of POCO NET failed because of the enum TIMESTAMP_REQUEST define in ICMPv4PackekImpl.h

> enum MessageType
> {
> 	TIMESTAMP_REQUEST ,
> };

The error indicates that there is a macro TIMESTAMP_REQUEST define somewhere, I don't know where but it is causing a conflict

My solution was to rename the enum from TIMESTAMP_REQUEST  to simply TIMESTAMP.

For information, I've previously built POCO 1.9.0-release without any issues with the same environment.

iSLC:
It conflicts with a macro named `TIMESTAMP_REQUEST` inside the `wincrypt.h` header.

```c
#define TIMESTAMP_REQUEST ((LPCSTR) 78)
```

matejk:
When trying to compile with MinGW 11 there are more compile issues with current state of devel.

Will fix and create pull request.

PR #4343, opened 2023-12-14
mingw 11: fixes to compile some libraries

matejk:
When verifying #2967 with mingw 11 that is provided with Qt installation I encountered more compile issues.

This pull requests solves some of them in Foundation, Net, Data and CMake files.

The following components compile:

- Encodings, XML, JSON, Util, Net, MongoDB, Redis, Prometheus, Data, Data/SQLite, Zip, PageCompiler, File2Page

Others don't and it would make lots of effort to make them compile and work.

The tests don't compile and fail with different errors.

### DIFF
--- a/Net/include/Poco/Net/SocketDefs.h
+++ b/Net/include/Poco/Net/SocketDefs.h
@@ -26,7 +26,9 @@
 
 #if defined(POCO_OS_FAMILY_WINDOWS)
 	#include "Poco/UnWindows.h"
-	#define FD_SETSIZE 1024 // increase as needed
+	#ifndef FD_SETSIZE
+		#define FD_SETSIZE 1024 // increase as needed
+	#endif
 	#include <winsock2.h>
 	#include <ws2tcpip.h>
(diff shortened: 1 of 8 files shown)


Net/include/Poco/Net/SocketDefs.h
```diff
@@ -26,7 +26,9 @@
 
 #if defined(POCO_OS_FAMILY_WINDOWS)
 	#include "Poco/UnWindows.h"
-	#define FD_SETSIZE 1024 // increase as needed
+	#ifndef FD_SETSIZE
+		#define FD_SETSIZE 1024 // increase as needed
```
teksturi, line 30: I quickly read about this and it might be that with MinGW default is 64. Would this be problem?

matejk: Yes it is defined to 64. I don't know about the implications.

teksturi: Ok. Maybe we just go with Windows default here.

Issue #4274, opened 2023-11-21
Remove VS 140, 150 Projects


aleks-f:
As stated - vs140 and 150 will not be pre-generated and distributed. It will still be possible to generate them using Progen, but POCO will assume c++17 from 1.13.0 on, so your mileage may vary.

PR #4342, opened 2023-12-13
RFC: Remove Windows CE support

teksturi:
As Windows CE does not support C++17 nor Visual Studio 2017 anymore support for it should be removed. There is still Windows CE documentation which tells situation.

---

RFC for now as I'm not totally sure if you guys have plans to continue support it with somehow. It is totally ok for reject. Note for reviewers. It is easier to review commits as they have logically groped.

PR #4323, opened 2023-12-05
Merge DNSSD into main Poco

soroshsabz:
ITNOA

Merge DNSSD into main Poco project

fixes https://github.com/pocoproject/poco/issues/2208

PR #4344, opened 2023-12-15
Remove VS 140, 150 Projects

teksturi:
Remove old Visual Studio project files for 140 and 150. This is done because
Poco will move on to C++17 and older Visual Studio does not support it.
It is still possible to generate these files.

Fixes: #4274 

---

If something else should also be deleted like root/build_vs140.cmd just say
and I will delete those as well. Also review in commit level. Will be easier
as they are logically grouped and do only one thing per commit.


ProGen/rebuildall32.bat
```diff
@@ -9,11 +9,8 @@ mkdir ProGen\log
 
 call ProGen\progen.bat Win32 >ProGen\log\progen32.log 2>&1
 rd /q /s lib bin     & call buildwin  90 rebuild all both Win32 samples tests devenv  > ProGen\log\90.x32.log 2>&1
```
matejk, line 11: Should the "90"(Visual Studio 2008) project files be removed also? Same applies to the .gitignore.

teksturi: I do not know. See these for more info.

https://github.com/pocoproject/poco/pull/4342#issuecomment-1854714427
https://github.com/pocoproject/poco/pull/4323#pullrequestreview-1766477057

matejk: I see. @aleks-f would know.

aleks-f: The only purpose for the VS90 projects is use  as a template by ProGen - it launches the VS conversion tool and gives it VS90 project file as a template to generate a later VS project versions (with some "massaging" in between). Nothing is built with VS90 projects anymore.

PR #4345, opened 2023-12-15
Build.macos.openssl@1.1

bas524:
This PR fixes problem whith poco tests on macos when we use makefiles and openssl version 1.1

Problem root couse is "we hope that we build with openssl@1.1 because we installed it, but default openssl in brew is openssl@3"

Problem visalisation is failed test 
```shell
1: CppUnit::TestCaller<CryptoTest>.testEncryptDecryptGCM
    "Poco::IOException:
I/O error: error:1C800066:Provider routines::cipher operation failed"
    in "<unknown>", line -1
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,7 @@ jobs:
           ./configure --everything --no-prefix --omit=PDF
           --odbc-include=/usr/local/opt/unixodbc/include --odbc-lib=/usr/local/opt/unixodbc/lib
           --mysql-include=/usr/local/opt/mysql-client/include --mysql-lib=/usr/local/opt/mysql-client/lib &&
+          --include-path="/usr/local/opt/openssl@1.1/include" --library-path="/usr/local/opt/openssl@1.1/lib" &&
           make all -s -j4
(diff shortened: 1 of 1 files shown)


matejk:
Changes in the PR are appreciated to correct wronly used paths.

Although it seems that they do not address the real problem with the CGM test, because it also happens in macOS cmake jobs that use openssl 1.1 and 3 properly.

bas524:
> Изменения в PR приветствуются для исправления путей, используемых только wronly.
> 
> Хотя кажется, что они не решают реальную проблему с тестом CGM, потому что это также происходит в заданиях macOS cmake, которые правильно используют openssl 1.1 и 3.

Let's resolve problems ste by step ))
Now I don't see failed builds with cmake, if they are, of course I'll try to fix them

matejk:
> Let's resolve problems ste by step )) Now I don't see failed builds with cmake, if they are, of course I'll try to fix them

I am currently working on the same issue and temporarily disabled the tests. There is a discussion in github that might help resolving the problem permanently.

https://github.com/openssl/openssl/issues/21119

PR #4007, opened 2023-04-13
add sendfile method for streamsocket

bas524:
This PR is motivated by #3277
I want add sendfile method to the stream socket
This functionality is cross platform, i.e. on posix systems it uses sendfile, on windows - TransmitFile

@obiltschnig , @aleks-f , Please review this changes and than I'll push changes in vcproj files and cmake files. Because there are many changes which adds mswsock.lib.

P.S. And further I'll add sendfile functionality to the httpserver

PR #4351, opened 2023-12-16
Using native sendfile from OS for HttpServer

bas524:
This PR is motivated by https://github.com/pocoproject/poco/pull/3368 and continuation of https://github.com/pocoproject/poco/pull/4007

If OS has native implementation of sendfile function then it will be used (only for cmake projects).
Other projects (make and vcxporj) you can uncomment define ```POCO_HAVE_SENDFILE``` in ```Poco/Config.h```

Note for emscripten sendfile will throw NotImplemented exception, because https://github.com/emscripten-core/emscripten/pull/16234

Simple proof that this optimization is useful

Test without optimization
```shell
testFile:

OK (1 tests)

./Net-testrunner testFile  0.04s user 0.01s system 15% cpu 0.297 total
```

Test with optimization
```shell
testFile:

OK (1 tests)

./Net-testrunner testFile  0.03s user 0.01s system 5% cpu 0.842 total
```
15% vs 5%


Net/src/HTTPServerResponseImpl.cpp
```diff
@@ -126,9 +126,22 @@ void HTTPServerResponseImpl::sendFile(const std::string& path, const std::string
 	{
 		_pStream = new HTTPHeaderOutputStream(_session);
 		write(*_pStream);
+#ifdef POCO_USE_SENDFILE_FOR_HTTPSERVER
 		if (_pRequest && _pRequest->getMethod() != HTTPRequest::HTTP_HEAD)
 		{
+			delete _pStream; // delete the stream to flush the HTTP headers to the socket, required by HTTP 1.0 and above
```
teksturi, line 132: Nicer would be first just call flush(). Not very good habit to flush with destructor.

Net/src/HTTPServerResponseImpl.cpp
```diff
@@ -126,9 +126,22 @@ void HTTPServerResponseImpl::sendFile(const std::string& path, const std::string
...
+			_pStream = nullptr;
+
+			Poco::Int64 sent = 0;
+			Poco::Int64 offset = 0;
+			while (sent < length)
+			{
+				offset = sent;
+				sent = _session.socket().sendFile(istr, offset);
```
teksturi, line 140: Why did we not implemented default for sendFile() so it can always sendFile even if it is not supported by platform. It would just use normal method then. Then we do not even need this POCO_USE_SENDFILE_FOR_HTTPSERVER. Not very nice in multiplatform library that sendFile just throws on some platforms.

CMakeLists.txt
```diff
@@ -241,6 +241,9 @@ option(POCO_ENABLE_STD_MUTEX "Set to OFF|NO using mutex from standard library (d
 if (POCO_ENABLE_STD_MUTEX)
 	add_definitions(-DPOCO_ENABLE_STD_MUTEX)
 endif ()
+
+option(POCO_USE_SENDFILE_FOR_HTTPSERVER "Set to OFF|ON using native sendFile function from OS for http server method HTTPServerResponse::sendFile" OFF)
```
teksturi, line 245: Let's try to avoid this kind of options if possible. We definitely can. We have couple option how this is not needed at all.

- We test at compile time if sendFile() is available
- We implement default behavior for sendFile() that it will always send file even when platform does not have sendfile operation.

I definetly would like option two. Does anyone know reason why we should not do that?

Net/src/HTTPServerResponseImpl.cpp
```diff
@@ -128,7 +128,20 @@ void HTTPServerResponseImpl::sendFile(const std::string& path, const std::string
 		write(*_pStream);
 		if (_pRequest && _pRequest->getMethod() != HTTPRequest::HTTP_HEAD)
 		{
+#ifdef POCO_USE_SENDFILE_FOR_HTTPSERVER
+			delete _pStream; // delete the stream to flush the HTTP headers to the socket, required by HTTP 1.0 and above
+			_pStream = nullptr;
+
+			Poco::Int64 sent = 0;
+			Poco::Int64 offset = 0;
+			while (sent < length)
```
teksturi, line 137: This can stay here for ever because _sendfile() is so buggie. It does not do prober error handling at Linux at least. We definitely need to look closer how to fix that before start using that. I also do not understand why it returns Int64. We example use 32 bit systems so it does not make much sense to force 64 bit return value.

obiltschnig: If we have any doubts about that we'll leave it out.

bas524: @teksturi , thank you for review . I'll fix logical mistake with ```sent``` and ```offset```

bas524: And about Int64...
Poco hasn't platform dependent integral types like size_t and ssize_t and I made a decision that Int64 is a compromise

Net/src/HTTPServerResponseImpl.cpp
bas524: Thank you. I replace destructor with ```flush()```

Net/src/HTTPServerResponseImpl.cpp
bas524: I refactored PR and now ```sendFile``` avaliable only on that platforms which have native sendfile function

CMakeLists.txt
bas524: I replaced option with cmake feature ```check_symbol_exists```

Net/src/HTTPServerResponseImpl.cpp
bas524: I hope that now code more correct. Please review it

Net/src/SocketImpl.cpp
```diff
@@ -1372,12 +1372,12 @@ void SocketImpl::error(int code, const std::string& arg)
 		throw IOException(NumberFormatter::format(code), arg, code);
 	}
 }
-
+#ifdef POCO_HAVE_SENDFILE
```
teksturi, line 1375: Still why won't we just implement sendFile so that it can be used with every platforms? It would just use sendfile / sendfile64 / TransmitFile when those are available. Then users can just always use our sendFile without thinking platforms. This is probably what they want anyway if they want to make multiplatform application. It is Poco's job to hide these kind of things.

We example also do not have seperete sendFile and transmitFile functions.

Net/src/SocketImpl.cpp
```diff
@@ -1433,21 +1434,26 @@ Poco::Int64 _sendfile(poco_socket_t sd, FileIOS::NativeHandle fd, Poco::UInt64 o
...
 
-Poco::Int64 SocketImpl::sendFile(FileInputStream &fileInputStream, Poco::UInt64 offset)
+IntPtr SocketImpl::sendFile(FileInputStream &fileInputStream, UIntPtr offset)
 {
 	FileIOS::NativeHandle fd = fileInputStream.nativeHandle();
-	Poco::UInt64 fileSize = fileInputStream.size();
+	UIntPtr fileSize = fileInputStream.size();
 	std::streamoff sentSize = fileSize - offset;
-	Poco::Int64 sent = 0;
+	IntPtr sent = 0;
```
teksturi, line 1442: Still if _sendFile sends multiple times this will be just last sent. So user will not actually get total size which is sended just last send size.

Net/src/SocketImpl.cpp
teksturi, line 1375: Still why do we not get rid of whole 

`throw Poco::NotImplementedException("sendfile not implemented for this platform");`

Then we just use read and normal socket to do send if that is not supported. Then user really does not have to think about these things. They just use our sendFile and it will work everytime. This is probably what they want as they use Poco which is multiplatform library.

Net/src/SocketImpl.cpp
```diff
@@ -1433,21 +1434,26 @@ Poco::Int64 _sendfile(poco_socket_t sd, FileIOS::NativeHandle fd, Poco::UInt64 o
...
 	sighandler_t sigPrev = signal(SIGPIPE, SIG_IGN);
 	while (sent == 0)
 	{
 		errno = 0;
 		sent = _sendfile(_sockfd, fd, offset, sentSize);
+		if (sent < 0)
+		{
+			error(errno, std::string("[sendfile error]") + Error::getMessage(errno));
```
teksturi, line 1450: Does not seem right. See other places where error() is used It should not use Error::getMessage(errno) as it looks like error will actually check hose. But anyway. This is probably even not the right place to do error handling. Every function should do throwing at they own. So _sendFile should be place to throw

Net/src/SocketImpl.cpp
bas524: No, sendFile sends only once, it repeat only when nothing was sent, it can happen when errno is equal to EAGAIN or EWOULDBLOCK

Net/src/SocketImpl.cpp
bas524: This behavior was approved with @aleks-f https://github.com/pocoproject/poco/pull/4007#discussion_r1419524268

Net/src/SocketImpl.cpp
bas524: _sendFile is a free function it can't throw exception by call error, code will be too complex
and _sendFile uses for *nix only, on windows doesn't

Net/src/SocketImpl.cpp
teksturi: Hmm. Ok. Seems strange to me as now users also need to do ifdef rules to use Poco. But if that was chosen already we go with that.

matejk:
@bas524 , can you please rebase on latest code on main branch, please?

bas524:
> @bas524 , can you rebase on latest code on main branch, please?

Hi, ok. I'll do it this week

Foundation/include/Poco/FileStream.h
```diff
@@ -79,7 +79,7 @@ class Foundation_API FileIOS: public virtual std::ios
 	NativeHandle nativeHandle() const;
 		/// Returns native file descriptor handle
 
-	Poco::UInt64 size() const;
+	UIntPtr size() const;
```
matejk, line 82: I don't think that changing the return type is appropriate.

Foundation/include/Poco/FileStream_POSIX.h
```diff
@@ -61,7 +61,7 @@ class Foundation_API FileStreamBuf: public BufferedBidirectionalStreamBuf
 	NativeHandle nativeHandle() const;
 		/// Returns native file descriptor handle
 	
-	Poco::UInt64 size() const;
+	UIntPtr size() const;
```
matejk, line 64: Why changing the return type?

Net/include/Poco/Net/SocketImpl.h
```diff
@@ -477,12 +477,17 @@ class Net_API SocketImpl: public Poco::RefCountedObject
 
 	bool initialized() const;
 		/// Returns true iff the underlying socket is initialized.
-
-	Poco::Int64 sendFile(FileInputStream &FileInputStream, Poco::UInt64 offset = 0);
+#ifdef POCO_HAVE_SENDFILE
+	IntPtr sendFile(FileInputStream &FileInputStream, UIntPtr offset = 0);
```
matejk, line 481: Same here: why change the type?

bas524: The problem is that sendfile function has architecture dependency in their params and return values
If we always use int64 than potentially we can have a bug in x32 systems.
If we always use int32 than we restrict users

Foundation/include/Poco/FileStream_POSIX.h
bas524: Result shoud be platform depended because we use it for sendfile.

Foundation/include/Poco/FileStream.h
bas524: I described it below.
The main idea - that filesize is platform depended

Net/include/Poco/Net/SocketImpl.h
matejk: What bug can we have on 32-bit systems?

Poco::Int64 as used before the PR can hold smaller 32-bit values. If necessary they can be cast to proper values when passing them to sendfile and not do the change across the library.

bas524: Imagine
x32 system
filesize = UINT_MAX + N
function sendFile has UInt64 param for filesize
system function sendfile has uint32 param for filesize.
on call sendfile we have a bug
....
I did changes only in sensitive places for reducing the probability of error

aleks-f: Strictly speaking, the idea is noble, but practically I am not in favor of this massive change in the interface types only to satisfy peculiarities of a system call on 32-bit systems. 

Since there is `POCO_HAVE_SENDFILE`, there can also be `SendFileSizeT` for casting when calling that API. In practice, this is mostly for Linux and that call is [standardized](https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib-sendfile.html). Other known platforms peculiarities can be defined now, or as they come. 

It's 2024, we should not be bending backwards for 32-bit systems.

aleks-f (changes requested):

Net/include/Poco/Net/SocketImpl.h
bas524: I have reverted types

Foundation/include/Poco/FileStream_POSIX.h
bas524: I have reverted

Foundation/include/Poco/FileStream.h
bas524: I have reverted types

matejk (approved):

Issue #4348, opened 2023-12-15
0 (or NULL) for null pointer shall be replaced with nullptr


matejk:
As described in the title.

Warning option `-Wzero-as-null-pointer-constant` can be used to locate places in the code.

teksturi:
I recommend using clang-tidy for these kind of things. We should automate these kind of things. Else it means more work to keep track. If these are not automated it is harder and harder new dev to come in to the project and follow all rules.

Here nice blog post https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/ 

We probably just need to start keeping list of external files or move them to own directory. I actually vote for own folder idea for those. Eather common external/third party source dir or our every lib get's own. This thing will have many benefits.

 - Easy to keep track of external debencies.
 - Easy way ignore those so they are not changed so easily
 - Easiear to check update for all of those
 - We can exlude them from code format automations.
 - For new dev it is very documentaries 
 
  See example how chromium does it https://github.com/chromium/chromium/tree/main/third_party

aleks-f:
> I recommend using clang-tidy for these kind of things. We should automate these kind of things. Else it means more work to keep track. If these are not automated it is harder and harder new dev to come in to the project and follow all rules.

i am in favor of this. It will, however, require the buildsystems (CMake is the official buildsystem, but makefiles and pre-generated VS projects should stay, too) modifications. Best to create a separate issue for this.

PR #4352, opened 2023-12-16
Clang tidy showcase

teksturi:
This PR is just showcase what kind of things clang-tidy could do. So I am not suggesting we take all of these checks in use. Currently not buildable as it makes mistakes with some cases. It did also found many bugs. I have to send fixes to those first before doing real clang-tidy things.

We just need to get these thing automated so they do not happened again. Note that these are only checks which clang-tidy can fix automatically. There is also bunch of good checks which can give just warning.

This was done for #4348


matejk:
@teksturi, this is excellent!

How would clang-tidy be automated?

Shall it be added as the last build target in CMake and make the build fail?

Or is it safe for automatic fixes to be applied also?

obiltschnig:
We should enable clang-tidy checks as part of normal CI.
In the first step, just print the warnings but not fail. After all things are fixed, make the build step fail if clang-tidy reports issues. Probably need to clarify what warnings or errors would cause the build to fail also (maybe some are acceptable/unavoidable).
I don't want automatic fixes for now.

PR #4353, opened 2023-12-16
Fix some issues found with clang-tidy

teksturi:
Clang-tidy did find some problems from code. I choose to fix these separately as it makes it easier to automate clang-tidy at some point.

---

Every commit has  own fix type and they have messages. Look them separately so reviewing is easier.


matejk (approved):
I approve the changes.

IMO this PR is safe to be included to 1.13. @aleks-f ?

PR #4357, opened 2023-12-18
fix(NamedEvent): Release semaphore ID in dtor when created with semget (fixes #2971)

### LINKED ISSUE #2971: Poco::NamedEvent does not release System V semaphore on Linux
Repro steps: create array of e.g. 256 Poco::NamedEvent instances. You'll get an error e.g.
System exception : cannot create named mutex ... semget() failed, errno=28

The Linux system allows up to 128 semaphore Ids and Poco::NamedEvent does not release semaphore id of the destructed Poco::NamedEvent instance.

I think in the NamedEvent_UNIX.cpp NamedEventImpl::~NamedEventImpl() there should be a call:

#if (POCO_OS == POCO_OS_LINUX) 
    // remove semaphores
   semctl (semid, 0, IPC_RMID);
...

matejk:
When a semaphore id is created with semget, it is released in dtor with semctl(..IPC_RMID).

### DIFF
--- a/Foundation/include/Poco/NamedEvent_UNIX.h
+++ b/Foundation/include/Poco/NamedEvent_UNIX.h
@@ -43,6 +43,7 @@ private:
 	sem_t* _sem;
 #else
 	int _semid;  // semaphore id
+	bool _createdId; // semaphore id was created with this instance
 #endif
 };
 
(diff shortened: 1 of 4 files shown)


matejk:
Releasing semaphore IDs on Linux seems to be solved, however it seems that named event is not correctly implemented on macOS also.

The following test throws exception with errno 28 (ENOSPC):

```
		int i = 0;
		for (; i < 40000; i++)
		{
			name = std::string("TestEvent ") + std::to_string(i);
			NamedEvent* ne = new NamedEvent(name);
			delete ne;
		}
```

`sem_unlink` shall probably be called somewhere, however I am not familiar with semaphores in details.

matejk:
@aleks-f , @obiltschnig , please take a look at this PR. It seems to work properly, however I am not 100% percent familiar with NamedEvent.

Issue #4341, opened 2023-12-13
1.12.4 version Json set enum value failed


xzzh999:
**Describe the bug**

```
Poco::JSON::Object::Ptr obj = new Poco::JSON::Object();

// set some enum value
obj->set("taskType", SOME_ENUM_VALUE);

Poco::Dynamic::Var var(obj);
std::string result = var.convert<std::string>();
```

got exception:

`poco got exception: Can not convert to std::string`

if changed to this:
```
obj->set("taskType", int(SOME_ENUM_VALUE));
```

it works fine.

**To Reproduce**
always failed

**Expected behavior**
json convert ok

**Logs**
`poco got exception: Can not convert to std::string`

**Please add relevant environment information:**
 - Win 10 x64 19044.3693
 - 1.12.4

**Additional context**
none

PR #4363, opened 2023-12-21
Allow set enum as json value

bas524:
fix issue #4341
add testEnum

```cpp
        enum SAMPLE_ENUM
	{
		SE_VALUE = 42
	};
	
	Poco::JSON::Object::Ptr obj = new Poco::JSON::Object();
	obj->set("simple_enum", SE_VALUE);
	Poco::Dynamic::Var var(obj);
	std::string expected = "{\"simple_enum\":42}";
	std::string result = var.convert<std::string>();
	assertEquals(expected, result);
		
	SAMPLE_ENUM se = obj->get("simple_enum").extract<SAMPLE_ENUM>();
	assertTrue(se == SE_VALUE);
```
Canges are simple, I've added branch for every default method VarHolderImpl<T>

```cpp
     void convert(Int8& val) const
	{
		if constexpr (std::is_enum<T>::value)
		{
			convertToSigned(std::underlying_type_t<T>(_val), val);
		}
		else
		{
			VarHolder::convert(val);
		}
	}   
```

### DIFF
--- a/Foundation/include/Poco/Dynamic/Var.h
+++ b/Foundation/include/Poco/Dynamic/Var.h
@@ -88,11 +88,18 @@ public:
 	Var();
 		/// Creates an empty Var.
 
-	template <typename T>
+	template <typename T, std::enable_if_t<!std::is_enum<T>::value, bool> = true>
 	Var(const T& val)
 		/// Creates the Var from the given value.
 	{
 		construct(val);
+	}
+	
+	template <typename T, std::enable_if_t<std::is_enum<T>::value, bool> = true>
+	Var(const T& val)
+		/// Creates the Var from the given enum value trough cast to int
+	{
+		construct(static_cast<int>(val));
 	}
 
 	Var(const char* pVal);
--- a/Foundation/include/Poco/Dynamic/Var.h
+++ b/Foundation/include/Poco/Dynamic/Var.h
@@ -88,18 +88,11 @@ public:
 	Var();
 		/// Creates an empty Var.
 
-	template <typename T, std::enable_if_t<!std::is_enum<T>::value, bool> = true>
+	template <typename T>
 	Var(const T& val)
 		/// Creates the Var from the given value.
 	{
 		construct(val);
-	}
-	
-	template <typename T, std::enable_if_t<std::is_enum<T>::value, bool> = true>
-	Var(const T& val)
-		/// Creates the Var from the given enum value trough cast to int
-	{
-		construct(static_cast<int>(val));
 	}
 
 	Var(const char* pVal);
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -773,7 +773,8 @@ public:
 		{
 			convertToSigned(std::underlying_type_t<T>(_val), val);
 		}
-		else {
+		else
+		{
 			VarHolder::convert(val);
 		}
 	}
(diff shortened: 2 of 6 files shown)


Foundation/include/Poco/Dynamic/Var.h
```diff
@@ -88,12 +88,19 @@ class Foundation_API Var
...
 		/// Creates the Var from the given value.
 	{
 		construct(val);
 	}
+	
+	template <typename T, std::enable_if_t<std::is_enum<T>::value, bool> = true>
+	Var(const T& val)
+		/// Creates the Var from the given enum value trough cast to int
+	{
+		construct(static_cast<int>(val));
```
matejk, line 102: I think that it would make sense to use std::undelying_type when casting. An enum can be declared with specific integer type.

https://en.cppreference.com/w/cpp/types/underlying_type

JSON/testsuite/src/JSONTest.cpp
```diff
@@ -2369,7 +2369,23 @@ void JSONTest::testRemove()
 	assertTrue(nl[1] == "baz");
 
 }
-
+void JSONTest::testEnum()
+{
+	enum SAMPLE_ENUM
```
matejk, line 2374: Does this work for enum class also?

JSON/testsuite/src/JSONTest.cpp
```diff
@@ -2369,7 +2369,23 @@ void JSONTest::testRemove()
...
+	{
+		SOME_ENUM_VALUE = 42
+	};
+	
+	Poco::JSON::Object::Ptr obj = new Poco::JSON::Object();
+	
+	obj->set("taskType", SOME_ENUM_VALUE);
+	
+	Poco::Dynamic::Var var(obj);
+	std::string expected = "{\"taskType\":42}";
```
matejk, line 2384: What about converting the value back to enum with `operator T()` or `extract()`?

matejk (changes requested):
The idea is very good. It needs some more work to be complete.

JSON/testsuite/src/JSONTest.cpp
bas524: For now - yes

Foundation/include/Poco/Dynamic/VarHolder.h
```diff
@@ -731,6 +767,264 @@ class VarHolderImpl: public VarHolder
 		return typeid(T);
 	}
 
+	void convert(Int8& val) const
+	{
+		if constexpr (std::is_enum<T>::value)
+		{
+			convertToSigned(std::underlying_type_t<T>(_val), val);
+		}
+		else {
```
aleks-f, line 776: style (this one, and many others) - curly brace goes in the new line

aleks-f (changes requested):
please add [Var tests](https://github.com/pocoproject/poco/blob/devel/Foundation/testsuite/src/VarTest.cpp), `extract()` must work

Foundation/include/Poco/Dynamic/VarHolder.h
bas524: Ups, sorry.
Fixed

bas524:
> please add [Var tests](https://github.com/pocoproject/poco/blob/devel/Foundation/testsuite/src/VarTest.cpp), `extract()` must work

Ok, I've done

aleks-f:
merged into poco issue branch to trigger CI

aleks-f:
#4434

Issue #4335, opened 2023-12-11
Implement patches that Debian/Ubuntu applies when preparing deb packages


matejk:
Debian Linux and Ubuntu Linux maintainers apply a set of patches to original Poco sources.

These patches shall be reviewed and possibly applied to the main code if they address some generic issues.

`poco-**.debian.tar.xz` from Debian packages web site contains the patches: https://packages.debian.org/sid/libpoco-dev

PR #4380, opened 2024-01-03
Incorporated Debian patches

matejk:
#4335

Applied relevant patched from https://packages.debian.org/sid/libpoco-dev that Debian maintainers use to prepare Poco 1.11.0 packages. Some of the issues addressed in the patched are already resolved in 1.13, some are skipped (like disabling certain tests).

### DIFF
--- a/Foundation/include/Poco/Platform.h
+++ b/Foundation/include/Poco/Platform.h
@@ -240,6 +240,11 @@
 #endif
 
 
+#ifdef __SOFTFP__
+#define POCO_NO_FPENVIRONMENT
+#endif
+
+
 #if defined(__clang__)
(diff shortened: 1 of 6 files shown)


aleks-f:
@matejk I'll merge this and then cherry pick to devel, but typically the workflow should be the other way around (PR to devel, then cherry pick into release branch, if appropriate).

In the end, the result is the same if both are done, but on the last release I had to manually reconcile quite a bit of changes that were in release branches but not in devel. It is quite frustrating having to do that (and even more for users to have bug fixed in an earlier release re-emerge later) and we should really make sure it does not happen again

aleks-f (commented):
LGTM

PR #4390, opened 2024-01-08
enh(DateTimeParser): option to cleanup input string before parsing (#569)

matejk:
Added a second set of `parse` functions with a new argument `cleanup` to explicitly request the input string to be cleaned.

Existing `parse` and `tryParse` perform cleanup. These could be marked as deprecated to issue compiler warnings to transition to stricter checking by default in one of the next releases.

@killerbot242, @aleks-f : what is your opinion on this?

### DIFF
--- a/Foundation/src/DateTimeParser.cpp
+++ b/Foundation/src/DateTimeParser.cpp
@@ -17,6 +17,7 @@
 #include "Poco/DateTime.h"
 #include "Poco/Exception.h"
 #include "Poco/Ascii.h"
+#include "Poco/String.h"
 
 
 namespace Poco {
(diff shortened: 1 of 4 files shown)


aleks-f (changes requested):
I say, just trim surrounding spaces indiscriminately and let's leave it at that, no additional functions and no future deprecation.

killerbot242:
I agree with Aleks, let's just trim, document this, so a committed feature, and nothing more.

Foundation/src/DateTimeParser.cpp
```diff
@@ -42,10 +43,19 @@ namespace Poco {
...
 
-void DateTimeParser::parse(const std::string& fmt, const std::string& str, DateTime& dateTime, int& timeZoneDifferential)
+inline void cleanupInputString(std::string& str)
+{
+	Poco::trimInPlace(str);
+}
+
+void DateTimeParser::parse(const std::string& fmt, const std::string& dtStr, DateTime& dateTime, int& timeZoneDifferential, bool cleanup)
 {
+	std::string str(dtStr);
```
aleks-f, line 53: to avoid string copy here, this may be implemented using `string_view` into `find_first_not_of` and `find_last_not_of` of the original string

matejk: It is not easy to avoid copying without major changes in the parser to use string_view instead (or in addition to string).

Issue #4019, opened 2023-04-26
MSYS2/mingw cmake linking problem


srebrnyp:
Host Environment:
*Host: x64-windows
*Compiler: GNU 12.2.0

When trying to link any of executable component from poco (ex: arc,tec) the same linking issue occurs for each component:
```
[514/575] cmd.exe /C "cd . && C:\msys64\mingw64\bin\x86_64-w64-mingw32-g++.exe -std=gnu++20 -m64 -O3 -DNDEBUG  Encodings/Compiler/CMakeFiles/EncodingsCompiler.dir/src/TextEncodingCompiler.cpp.obj -o bin\tec.exe -Wl,--out-implib,lib\libtec.dll.a -Wl,--major-image-version,0,--minor-image-version,0  lib/libPocoNet.dll.a  lib/libPocoUtil.dll.a  -liphlpapi  -lws2_32  lib/libPocoXML.dll.a  C:/msys64/home/psrebrny/one/onespin/2023.2/install/x64-mingw/lib/libexpat.a  -lm  lib/libPocoJSON.dll.a  lib/libPocoFoundation.dll.a  C:/msys64/home/psrebrny/one/onespin/2023.2/install/x64-mingw/lib/libpcre2-8.a  C:/msys64/home/psrebrny/one/onespin/2023.2/install/x64-mingw/lib/libzlib.dll.a  -liphlpapi  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
FAILED: bin/tec.exe 
[...]

PR #4391, opened 2024-01-10
mingw compile and link improvements (#4019)

matejk:
This PR resolves link errors reported in #4019.

IMO building Poco with MinGW with CMake is in a bad state in general.

Foundation-testrunner, for example doesn't find `Poco::SHA2Engine` and `Data `does not compile at all.

Shall any additional effort be invested in MinGW+CMake?


aleks-f (approved):

aleks-f:
> Shall any additional effort be invested in MinGW+CMake?

Anyone needing MinGW build is free to fix it (and take ownership of the maintenance).

PR #4394, opened 2024-01-13
gcc/clang (-fvisibility=hidden): corrections to compile and work properly

### LINKED ISSUE #3331: Apple Silicon ARM64 :  Invalid access: Can not convert empty value.
I have compiled poco for arm64 for Apple Silicon Native support. But for arm devices JSON parsing fails even for simplest json contracts. Please see below code snippet.

```
std::string json = "{\"clientConfig\" : \"Franky\"}";
        Poco::JSON::Parser jsonParser;
        Poco::Dynamic::Var jsonObject = jsonParser.parse(json);
        Poco::JSON::Object::Ptr jsonPtr = jsonObject.extract<Poco::JSON::Object::Ptr>();
        std::string testStr = jsonPtr->getValue<std::string>("clientConfig"); -> Here I get that exception.
```
If I check` jsonPtr->get("clientConfig").isEmpty()` , it returns true as well.

I tried to compile with -fvisibility=hidden as well but same result.
If I compiled for x86_64 and run on Silicon Macs(through rosetta), it work fine. On iOS it works fine too.
If I run in debug mode on Apple Silicon Mac, it works fine, issue is only in release mode.

Any ideas what's causing this?

### LINKED ISSUE #4393: iOS ARM64 : Invalid access: Can not convert empty value.
**Describe the bug**
This bug is similar to this - https://github.com/pocoproject/poco/issues/3331 but on iOS. Exact same config is failing. 

Exact same code snippet and same except as provided in the description of the above bug.
Only in Release mode
-fvisibility=hidden compiler flags
ENABLE_TESTABILITY set to NO

**To Reproduce**
NA

**Expected behavior**
Json parsing should succeed.

**Logs**
NA

**Screenshots**
NA

 - iOS-16 
 - POCO Version - 1.11.7

matejk:
When using option `-fvisibility=hidden` (or `set(CMAKE_CXX_VISIBILITY_PRESET hidden)`) some of the symbols are not exported from libraries and some templates used in `Poco::Dynamic` are instantiated multiple times in different compilation units and therefore type comparison in `VarHolder` fails.

This PR resolves these issues and other problems related to hidden visibility.

Visibility is also changed to hidden when compiling with CMake.

Fixes #3331, fixes #4393.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,8 +46,6 @@ endif()
 include(CXX1x)
 check_for_cxx17_compiler(CXX17_COMPILER)
 
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-
 # If a C++17 compiler is available, then set the appropriate flags
 if(CXX17_COMPILER)
 	set(CMAKE_CXX_STANDARD 17)
@@ -559,6 +557,7 @@ message(STATUS "[cmake] Build type:             ${CMAKE_BUILD_TYPE}")
 string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE)
 message(STATUS "[cmake] Build with cxx flags:   ${CMAKE_CXX_FLAGS_${BUILD_TYPE}} ${CMAKE_CXX_FLAGS}")
 message(STATUS "[cmake] Build with c flags:     ${CMAKE_C_FLAGS_${BUILD_TYPE}} ${CMAKE_C_FLAGS}")
+message(STATUS "[cmake] C++ symbol visibility:  ${CMAKE_CXX_VISIBILITY_PRESET}")
 
 foreach(component ${Poco_COMPONENTS})
 	message(STATUS "Building: ${component}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,8 @@ endif()
 include(CXX1x)
 check_for_cxx17_compiler(CXX17_COMPILER)
 
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+
 # If a C++17 compiler is available, then set the appropriate flags
 if(CXX17_COMPILER)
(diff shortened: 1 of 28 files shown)


CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ endif()
 include(CXX1x)
 check_for_cxx17_compiler(CXX17_COMPILER)
 
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
aleks-f, line 49: I would leave this for 1.14; also, if we make it default for `cmake`, then it should also become default for `make`
just create a `cmake` and a `make` CI job with hidden visibility and we'll remove those jobs once we make hidden default

aleks-f (changes requested):

CMakeLists.txt
matejk: Added make and cmake jobs for macOS with visibility set to hidden.

JSON/include/Poco/JSON/Array.h
```diff
@@ -26,8 +26,23 @@
...
+
+}
+
+
+#if defined(POCO_OS_FAMILY_WINDOWS)
+// Explicitly instantiated shared pointer in JSON library
+extern template class Poco::SharedPtr<Poco::JSON::Array>;
+#else
+// Explicitly instantiated shared pointer in JSON library
+extern template class JSON_API Poco::SharedPtr<Poco::JSON::Array>;
```
mbhaskar98, line 42: here JSON_API visibilty macro is with non - POCO_OS_FAMILY_WINDOWS. 

But in cpp file its with - POCO_OS_FAMILY_WINDOWS. Is this expected? 

same with Object.h and Object.cpp file.

matejk: Yes, it is expected. Explicitly instantiating templates is exported differently on Windows.

PR #4397, opened 2024-01-15
Silence unnecessary NoCertificateException to  fix #4370

### LINKED ISSUE #4370: Can we silence NoCertificateException in NetSSL_Win Net::Context::certificate when mustFindCertificate is false?
Only place I see Net::Context::certificate() called is in SecureSocketImpl::loadCertificate which has a mustFindCertificate parameter. Net::Context::certificate() doesn't know about that parameter so far, so it always throws an exception in the following code which is then ignored again in loadCertificate:
``
	if (_certNameOrPath.empty())
		throw NoCertificateException("Certificate requested, but no certificate name or path provided");
``
The problem is that despite being ignored later it prints out those exceptions on the console and I'm getting quite a few of those.

If we would pass on mustFindCertificate to Net::Context::certificate() it could ignore that check and directly do an empty throw. Or maybe you have some way to silence an exception. Or return some invalid X509Certificate. Anything which silences this would be nice.

mzeilfelder:
Avoid creating log messages for NoCertificateException when that exception should be ignored.
Passing on the mustFindCertificate variable one level deeper so Poco::Net::X509Certificate::certificate  already knows it shouldn't create a message for this case.

No idea if this is the best way to handle this, which is why I only created an issue for this originally, but it's one way to fix it.

Alternative might for example be to always create an NoCertificateException, but only pass a message-string when mustFindCertificate is true (if this is possible with your exceptions, I didn't dig that deep). In which  case the documentation change I made for certificate wouldn't be necessary either.

### DIFF
--- a/NetSSL_Win/src/Context.cpp
+++ b/NetSSL_Win/src/Context.cpp
@@ -140,13 +140,18 @@ void Context::addTrustedCert(const Poco::Net::X509Certificate& cert)
 }
 
 
-Poco::Net::X509Certificate Context::certificate()
+Poco::Net::X509Certificate Context::certificate(bool mustFindCertificate)
 {
 	if (_pCert)
 		return Poco::Net::X509Certificate(_pCert, true);
 
 	if (_certNameOrPath.empty())
-		throw NoCertificateException("Certificate requested, but no certificate name or path provided");
+	{
+		if (mustFindCertificate)
+			throw NoCertificateException("Certificate requested, but no certificate name or path provided");
+		else
+			throw;
+	}
 
 	if (_options & OPT_LOAD_CERT_FROM_FILE)
 	{
--- a/NetSSL_Win/include/Poco/Net/Context.h
+++ b/NetSSL_Win/include/Poco/Net/Context.h
@@ -201,11 +201,13 @@ public:
 		///
 		///   context.requireMinimumProtocol(PROTO_TLSV1_2);
 
-	Poco::Net::X509Certificate certificate();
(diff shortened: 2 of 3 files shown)


NetSSL_Win/src/Context.cpp
```diff
@@ -140,13 +140,18 @@ void Context::addTrustedCert(const Poco::Net::X509Certificate& cert)
...
 
 
-Poco::Net::X509Certificate Context::certificate()
+Poco::Net::X509Certificate Context::certificate(bool mustFindCertificate)
 {
 	if (_pCert)
 		return Poco::Net::X509Certificate(_pCert, true);
 
 	if (_certNameOrPath.empty())
-		throw NoCertificateException("Certificate requested, but no certificate name or path provided");
```
matejk, line 149: I inspected the code and did not find where this exception is logged.

mzeilfelder: Oh sorry, I'm an idiot and wasted your time. Visual Studio logs all exceptions, I just didn't notice that before for whatever reason and thought it has to do with the POCO_IMPLEMENT_EXCEPTION without digging further :-( So my throw in this patch still logs an exception (just without that string). Which obviously improves nothing.  Only way to get no message in VS would be to avoid throwing exceptions completely. Which I guess would need more rework like handling an invalid X509Certificate or something like that. And I'm really not up to providing a patch for that. Don't even know if it would fit in this library or if exceptions are simply used as error codes sometimes.

So can be closed I guess. And thanks for taking a look. At least I learned something.

Issue #661, opened 2015-01-07
Automatic Lib Init (NetworkInitializer) is not working on MinGW


asari-fd:
I'm writing a network client using Poco and MinGW.
It seems that I have to call Poco::Net::initializeNetwork() before calling any network-related functions, otherwise I get an exception like "DNS error: EAI: 10093", which means WSANOTINITIALISED (according to http://support.microsoft.com/kb/819124/) -- is it intended?

I think I'd be happy if I can always omit initializeNetwork() calls. Thanks.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.1 LTS
Release:    14.04
Codename:   trusty
```

```
$ dpkg -l | grep mingw32
ii  mingw32                           4.2.1.dfsg-2ubuntu1                      amd64        Minimalist GNU win32 (cross) compiler
ii  mingw32-binutils                  2.20-0.2ubuntu1                          amd64        Minimalist GNU win32 (cross) binutils
ii  mingw32-runtime                   3.15.2-0ubuntu1                          all          Minimalist GNU win32 (cross) runtime
```

```
$ i586-mingw32msvc-gcc --version
i586-mingw32msvc-gcc (GCC) 4.2.1-sjlj (mingw32-2)
Copyright (C) 2007 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
$ cat VERSION 
1.6.0 (2014-12-22)
```

asari-fd:
I understand. Thanks for your comment.

Speakus:
The issue is inside Poco/Net/Net.h file. I update this to fix issue: http://pastebin.com/3fWrnw0q

aleks-f:
[fixed](https://github.com/pocoproject/poco/commit/e469e4fc62163db69385a2cfadcb00dfb4ed1196#diff-d423a64f4fe3cb7dd86a8785ea0518ef)

OldWolf2:
With MinGW-w64 7.2.0 and poco-1.8.1-release, this problem recurs but only if POCO_STATIC=1 and we static link to Poco. (It worked correctly for a DLL-based POCO build). I work around this by explicitly calling Poco::Net::initializeNetwork() in my code.

aleks-f:
See if this helps and let us know, we may add it:
```diff
[...]

PR #4402, opened 2024-01-17
Implemented automated network library initialization for Windows MinGW targets

jackorobot:
Fixes issue #661.
Tested working using MinGW gcc 13.2.0 on Windows 10.

### DIFF
--- a/Net/include/Poco/Net/Net.h
+++ b/Net/include/Poco/Net/Net.h
@@ -101,11 +101,13 @@ std::string htmlize(const std::string& str);
 // Automate network initialization (only relevant on Windows).
 //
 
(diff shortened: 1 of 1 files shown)


Net/include/Poco/Net/Net.h
```diff
@@ -101,11 +101,13 @@ std::string htmlize(const std::string& str);
 // Automate network initialization (only relevant on Windows).
 //
 
-#if defined(POCO_OS_FAMILY_WINDOWS) && !defined(POCO_NO_AUTOMATIC_LIB_INIT) && !defined(__GNUC__)
+#if defined(POCO_OS_FAMILY_WINDOWS) && !defined(POCO_NO_AUTOMATIC_LIB_INIT)
```
matejk, line 104: `POCO_COMPILER_MINGW` can be used for code that is only MINGW specific.

jackorobot: That is a great suggestion. I have applied and pushed that change

PR #4405, opened 2024-01-19
fix(Environment): define _USE_32BIT_TIME_T on 32-bit Windows

matejk:
When compiling `Environment_WIN32U.cpp` on 32-bit Windows it is required to set the compiler definition `-D_USE_32BIT_TIME_T` to be able to properly use system call `GetAdaptersInfo`.

Unfortunately it must be set on the compiler command line. It does not work if it is set inside the cpp file. Foundation CMake file was therefore modified to set the define for a single file.

Is it necessary to have this for VS project files also? Is Poco still compiled with VS projects on 32-bit Windows?

PR also includes an improvement to pre-allocate larger buffer for network adapters to avoid scanning for available adapters for mist common situations.

### DIFF
--- a/Foundation/CMakeLists.txt
+++ b/Foundation/CMakeLists.txt
@@ -134,6 +134,13 @@ if(NOT BUILD_SHARED_LIBS)
 endif()
 
 if(WIN32)
+	if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+		# See comments in Environment_WIN32U.cpp.
+		#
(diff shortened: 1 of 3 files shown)


matejk:
The define is set only when compiling a single file, not complete project. Alternative is to use different function as written in the related issue.

aleks-f:
I'm not sure if progen is currently capable of doing this. I'd say, never mind the VS solutions; they are not officially supported, and if someone needs it, they can do it and send a pull. Everyone else has cmake-generated projects.

matejk:
This PR does not contain optimal solution. GetAdaptersAddresses shall be used.

Issue #4411, opened 2024-01-23
NULL pointer: strategy when setting rotation never in FileChannel


thebearon:
Setting `rotation` property as `never` via `setProperty(...)` results in the following:
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
NULL pointer: strategy [in file "/.../Foundation/src/FileChannel.cpp", line 283]
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

PR #4412, opened 2024-01-23
fix: NULL pointer strategy when setting rotation never #4411

thebearon:
Regression from 66e93f98cc3530c9ea5a748cf2248d2592b61853.

### DIFF
--- a/Foundation/testsuite/src/FileChannelTest.cpp
+++ b/Foundation/testsuite/src/FileChannelTest.cpp
@@ -59,6 +59,33 @@ FileChannelTest::~FileChannelTest()
 }
 
 
+void FileChannelTest::testRotateNever()
+{
+	std::string name = filename();
+	try
+	{
+		AutoPtr<FileChannel> pChannel = new FileChannel(name);
+		pChannel->setProperty(FileChannel::PROP_ROTATION, "never");
+		pChannel->open();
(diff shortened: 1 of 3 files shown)


Foundation/testsuite/src/FileChannelTest.cpp
```diff
@@ -59,6 +59,33 @@ FileChannelTest::~FileChannelTest()
...
+		}
+		File f(name);
+		assertTrue (f.exists());
+		f = name + ".0";
+		assertTrue (!f.exists());
+	}
+	catch (...)
+	{
+		remove(name);
+		throw;
```
matejk, line 83: You shall fail the test here instead of re-throwing exception.

thebearon: Thanks for the review! The rationale for copying the pattern from the other similar tests was that it would primarily test that the log isn't rotated (or at least a simple case of that), and as a bonus it also fails on the bug that's been fixed. Changing the throw would deviate from the pattern, and make it less like the other tests.
Shall I go through with the change regardless?

matejk: I see. I did not look at other test cases in the same file. I agree that it is better to stick to the pattern.

PR #4417, opened 2024-01-25
fix(logs): synchronise log file rotation and compression.

matejk:
This PR solves log file naming when compression is enabled by locking log file name rotation and compression with a mutex.

The file being compressed is renamed by prefixing to be skipped by purge. The prefix is removed at the end of compression procedure.

Purge strategy also interferes with this process and might have to be included in this locking process.

### DIFF
--- a/Foundation/include/Poco/ArchiveStrategy.h
+++ b/Foundation/include/Poco/ArchiveStrategy.h
@@ -23,6 +23,8 @@
 #include "Poco/File.h"
 #include "Poco/DateTimeFormatter.h"
 #include "Poco/NumberFormatter.h"
+#include "Poco/Mutex.h"
+#include "Poco/Condition.h"
 #include <atomic>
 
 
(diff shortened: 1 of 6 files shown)


aleks-f:
Lots of CI failures, something is not right

matejk:
Unit test that I wrote is still timing sensitive. Log rotation, purge and compression has some design flaws. I wanted to check with you if the approach in draft PR is good enough as temporary improvement.

matejk:
The PR was updated: file being compressed is prefixed with `"__"`, which is removed when compressions is done.

It looks like the combination of mutex and renaming the file solves the issue. I was not able to reproduce anymore.

matejk:
PR is ready to be reviewed.

The reason for failed task on Windows is missing compile flags for C++17 in VS files and that was resolved on branch "4368-oracle-odbc-tests".

aleks-f:
moved this to 1.14 (interface changes and stil some CI errors)

Issue #4392, opened 2024-01-11
Environment_WIN32U nodeIdImpl access violation in 1.13.0


mecsco:
**Describe the bug**
After upgrading from 1.12.4 to 1.13.0, in Environment_WIN32U.cpp the EnvironmentImpl::nodeIdImpl function throws an access violation when making the second call to GetAdaptersInfo if the first call has returned 111 (ERROR_BUFFER_OVERFLOW)

This appears to have been introduced in https://github.com/pocoproject/poco/issues/4328 (https://github.com/pocoproject/poco/pull/4336, https://github.com/pocoproject/poco/commit/cfc9ce380ee9970f2d90cf9061541e239d06d261).

**Expected behavior**
Same behaviour as 1.12.4

**Please add relevant environment information:**
 - Windows 10 64 bit
 - POCO 1.13.0

andrewauclair:
Can you send the length returned by the first `GetAdaptersInfo`? It should be a multiple of the struct size from what I understood while making this change.

mecsco:
`len` is 3840, and `sizeof(IP_ADAPTER_INFO)` is 648

andrewauclair:
I believe this is some sort of packing issue. On the default packing the `IP_ADAPTER_INFO` struct is 704 bytes for me, `GetAdaptersInfo` returns 2816 (Win10 x64), which is a proper 4 multiple.

winnt.h contains the following assert, which verifies that the default packing option is used. Unfortunately it appears that this assert doesn't happen for other headers, such as `iphlpapi.h` in this case.

```cpp
static_assert(__alignof(LARGE_INTEGER) == 8, "Windows headers require the default packing option. Changing this can lead to memory corruption."
    " This diagnostic can be disabled by building with WINDOWS_IGNORE_PACKING_MISMATCH defined.");
```

648 appears to be the correct size for x86. I can't find a way to get that size for x64. Even forcing `#pragma pack(push, 1)` before including the header for `IP_ADAPTER_INFO` results in a minimum of 676.

mecsco:
Apologies, I've clarified the environment in my original comment - the project configuration is indeed x86

andrewauclair:
According to [this](https://learn.microsoft.com/en-us/windows/win32/api/iptypes/ns-iptypes-ip_adapter_info?redirectedfrom=MSDN) page on `IP_ADAPTER_INFO`, there's an extra step to properly use it on 32 bit platforms.

[...]

PR #4419, opened 2024-01-27
Implement GetAdaptersAddresses API

andrewauclair:
Upgrade from `GetAdaptersInfo` to `GetAdaptersAddresses` API. The code has been swapped back to a buffer of bytes because the data structure built by `GetAdaptersAddresses` is a linked list and the returned size is not a multiple of the `IP_ADAPTERS_ADDRESSES` struct.

I've also set the `GAA_FLAG_SKIP_DNS_SERVER` flag in the `GetAdaptersAddresses` calls. We're only interested in the physical addresses of the interfaces, not their DNS information. I have to confirm next week at work, but I believe we've experienced large delays due to this in the past with the current `GetAdaptersInfo` implementation.

This pull request fixes issue #4392 and replaces PR #4405.

### DIFF
--- a/Foundation/src/Environment_WIN32U.cpp
+++ b/Foundation/src/Environment_WIN32U.cpp
@@ -16,15 +16,13 @@
 #include "Poco/Exception.h"
 #include "Poco/UnicodeConverter.h"
 #include "Poco/Buffer.h"
+
 #include <sstream>
 #include <cstring>
 #include <memory>
-#include "Poco/UnWindows.h"
-#include <winsock2.h>
-#include <wincrypt.h>
-#include <ws2ipdef.h>
-#include <iphlpapi.h>
 
+#include <winsock2.h>
+#include <iphlpapi.h>
 
 #if defined(_MSC_VER)
 #pragma warning(disable:4996) // deprecation warnings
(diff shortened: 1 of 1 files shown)


Foundation/src/Environment_WIN32U.cpp
```diff
@@ -16,16 +16,14 @@
 #include "Poco/Exception.h"
 #include "Poco/UnicodeConverter.h"
 #include "Poco/Buffer.h"
+
 #include <sstream>
 #include <cstring>
 #include <memory>
-#include "Poco/UnWindows.h"
```
aleks-f, line 22: @andrewauclair why remove this? the main purpose of that header is to reduce windows bloat through definition of `WIN32_LEAN_AND_MEAN`

aleks-f (changes requested):

Foundation/src/Environment_WIN32U.cpp
andrewauclair: I ended up starting the PR #4405 code from @matejk. I can put that back if you want.

Foundation/src/Environment_WIN32U.cpp
```diff
@@ -34,200 +33,206 @@
...
-
-
-void EnvironmentImpl::setImpl(const std::string& name, const std::string& value)
-{
-	std::wstring uname;
-	std::wstring uvalue;
-	UnicodeConverter::toUTF16(name, uname);
-	UnicodeConverter::toUTF16(value, uvalue);
-	if (SetEnvironmentVariableW(uname.c_str(), uvalue.c_str()) == 0)
+	std::string EnvironmentImpl::getImpl(const std::string& name)
```
matejk, line 36: There is no need to indent all of the functions in this file. Please revert the indentation as it was before and as it is also used in other files.

andrewauclair: Huh, weird. I bet Visual Studio did this to me. Definitely didn't show it as a diff locally in VS, which is really annoying.

andrewauclair: Looks fixed now.

aleks-f (approved):

matejk (approved):
I approve the implementation.

PR #3444, opened 2021-10-25
Allow using `Poco::FileStream` to wrap arbitrary file handles/descriptors as C++ streams

dgrunwald:
In our application we open files in Python code, pass the file handle/descriptor to C++ code, and then use `FileStream` to access the file from C++. It's also possible for Python to pass a duplicate of the stdin handle to C++, so we need to handle the `ERROR_BROKEN_PIPE` special case.

PR #4424, opened 2024-01-29
Allow using Poco::FileStream to wrap arbitrary file handles/descriptors as C++ streams 

matejk:
Contains PR #3444 rebased on latest devel.

* Allow opening a file descriptor/HANDLE as C++ stream.

* FileStream: treat read from closed pipe as EOF.

* chore(Filestream): conde style (naming)


aleks-f (changes requested):
a testcase should be added. 
this should be released in 1.14 - it changes the interface, so it is more appropriate to introduce it in major release

Issue #4367, opened 2023-12-21
`SQLite` `FTS5` (full text search)


aleks-f:
https://www.sqlite.org/fts5.html

SQLITE full text search.

PR #4430, opened 2024-01-31
Update SQLite (3.43 > 3.45) and enable FTS5 (#4367)

OgreTransporter:
Update SQLite (3.43 > 3.45) and enable FTS5 (#4367)


aleks-f (changes requested):
- [x] redirect to devel branch
- [x] FTS should not be enabled by default
- [x] VS projects should not be edited directly; instead, use [progen](https://github.com/pocoproject/poco/wiki/Generating-VS-Projects-using-ProGen) file to regenerate projects (that said, (1) VS projects are not officially supported and are here because some of us like having them around; I'm not sure in how good shape are 160 projects given that we require c++17 now, and (2) there's [this change](https://github.com/pocoproject/poco/pull/4410/commits/545c770e21aa82422a3bdf9dde6a1e1ab19047f9), which should be in devel later today)

aleks-f (approved):

aleks-f:
#4433 (this PR is too messy and did not trigger all CI)

PR #4434, opened 2024-02-01
Allow set enum as json value (#4363)

aleks-f:
* fix issue #4341 add testEnum
if you set enum as json value then it will be cast to int

* add behavior for enum into VarHolderImpl add tests for enum classes with different underline integral types

* replace static_cast with convertTo* methods, they can check types and limits

* fix missing types for convert methods

* fix code style add testEnumType to the VarTest.cpp, extract() works


Foundation/include/Poco/Dynamic/VarHolder.h
```diff
@@ -787,6 +823,287 @@ class VarHolderImpl: public VarHolder
...
+		{
+			convertToSigned(std::underlying_type_t<T>(_val), val);
+		}
+		else
+		{
+			VarHolder::convert(val);
+		}
+	}
+
+	void convert(Int64& val) const
```
matejk, line 862: I think that having partial specialisation for signed (and another for unsigned) integer types would simplify the code. The functions are identical except for the type.

Foundation/include/Poco/Dynamic/VarHolder.h
```diff
@@ -787,6 +823,287 @@ class VarHolderImpl: public VarHolder
...
+		{
+			val = (std::underlying_type_t<T>(_val) != 0);
+		}
+		else
+		{
+			VarHolder::convert(val);
+		}
+	}
+
+	void convert(float& val) const
```
matejk, line 962: Is conversion of enum to float meaningful?

Foundation/include/Poco/Dynamic/VarHolder.h
```diff
@@ -787,6 +823,287 @@ class VarHolderImpl: public VarHolder
...
+			std::string str = NumberFormatter::format(std::underlying_type_t<T>(_val));
+			Poco::UnicodeConverter::convert(str, val);
+		}
+		else
+		{
+			VarHolder::convert(val);
+		}
+	}
+
+	bool isArray() const
```
matejk, line 1023: I think that partial specialisation could also simplify all of the is* member functions.

Foundation/include/Poco/Dynamic/VarHolder.h
aleks-f: the problem is these are virtual functions so they become overloads, not overrides

Foundation/include/Poco/Dynamic/VarHolder.h
aleks-f: probably not, but I don't think it hurts

Foundation/include/Poco/Dynamic/VarHolder.h
aleks-f: I'm not sure how would this work, consolidating all functions into one `is()` and having an enum designator? that would break a lot of code

matejk: After taking another look, my first suggestion does not make much sense.

Foundation/include/Poco/Dynamic/VarHolder.h
matejk: True, missed that.

matejk (approved):

Issue #4414, opened 2024-01-24
Improve NotificationCenter speed and usability


aleks-f:
**Is your feature request related to a problem? Please describe.**
Currently, the `NotificationCenter` calls `(N)Observer` callback in the same thread where the subscribed-to event occurs. This slows down the event collection by executing the handling functionality "inline" in the event collection loop. 

Additionally, notification dispatch in `(N)Observer` depends on `dynamic_cast` to determine whether to actually post the `Notification` to the `Observer`. The downside of this approach (aside from the `dynamic_cast` performance hit) is that it is not possible to distinguish observers by any other criteria except the compile-time notification type. 

**Describe the solution you'd like**
Decoupling event detection from handling would allow to collect event-generating data without being slowed down by the processing workload.
Adding an additional (runtime) matching criteria would relax the current constraints and only actually interested observers could be determined prior to dynamic casting and calling the handler.

- `NotificationCenter`:
  - `enqueueNotification()`
- `NObserver` (all optional, `*` is only to describe what it conceptually is, could also be a `unique_ptr`):
  - `dequeueNotification()`
  - `NotificationQueue*`
  - `Poco::Thread*`
  - `Dispatcher*` (`Runnable` child)
  - `(*match)(const Poco::Any&)` callback

**Describe alternatives you've considered**
Alternatively, this could be implemented as a separate pair of classes, inheriting from the existing ones (eg. `ActiveNotificationCenter` and `ActiveObserver`)

**Additional context**
All this should be added as a default-off option, ie. without affecting the existing functionality.
I'm inclined to deprecate the `Poco::Observer` in the future, eventually renaming `NObserver` to `Observer` with perhaps notifications being `std::unique_ptr` which are moved into the handlers.

matejk:
`NObserver` and `Observer` shall definitely be merged into one implementation.

Is `NotificationCenter::enqueueNotification()` planned to be an enhanced `postNotification()` which does criteria matching?

Will a new thread be introduced that performs notification dispatching to registered observers to avoid the "inline" handling function?

obiltschnig:
Rather than one class making do too much, I'd prefer a new class that does things differently, but in one specific way only.

aleks-f:
> Is `NotificationCenter::enqueueNotification()` planned to be an enhanced `postNotification()` which does criteria matching?

No, `NotificationCenter` should only enqueue notifications into a `NotificationQueue`. 

[...]

PR #4444, opened 2024-02-06
AsyncObserver

aleks-f:
#4414 

Summary:

- existing functionality is only broken in the (now deprecated) [AbstractObserver::accepts()](https://github.com/pocoproject/poco/pull/4444/files#diff-c01387ad7bebc0f44dc73dd78c3b81758e087ac8e5c5083d9e359ca9d4c3b313R41-R44)
  This was a "quick and dirty" external path to the goal of filtering notifications by name (in addition to type-filtering)
- optional [_matcher](https://github.com/pocoproject/poco/blob/b6e310fe2ddd13abcfc82c4567c7591c54f7ea3d/Foundation/include/Poco/NObserver.h#L150) is added to `NObserver`
  (defaults to null and does not break existing functionality)
- [NotificationCenter::observersTonotify()](url) filters out the non-interested observers
  - this is the only place where `dynamic_cast` performance hit is taken
  - unnecessary casting is avoided when matcher is present and does not match
  - since observers are filtered, only the ones with matching type are notified, so `notify()` can cast [statically](https://github.com/pocoproject/poco/blob/b6e310fe2ddd13abcfc82c4567c7591c54f7ea3d/Foundation/include/Poco/NObserver.h#L93)
- `AsyncObserver` is child of `NObserver` and runs its own [thread](https://github.com/pocoproject/poco/blob/b6e310fe2ddd13abcfc82c4567c7591c54f7ea3d/Foundation/include/Poco/AsyncObserver.h#L135-L158) to decouple Notification posting from handling
- added`AsyncNotificationCenter`, a child class of the `NotificationCenter`; currently, there is a problem with default static version of this observer on windows:
  https://github.com/pocoproject/poco/blob/71560b0f1c3b68990a554d4aba59759c87590f2e/Foundation/testsuite/src/NotificationCenterTest.cpp#L252-L255

### DIFF
--- a/Foundation/Foundation_vs160.vcxproj
+++ b/Foundation/Foundation_vs160.vcxproj
@@ -1493,6 +1493,7 @@
     <ClInclude Include="include\Poco\AccessExpireStrategy.h" />
     <ClInclude Include="include\Poco\ActiveDispatcher.h" />
     <ClInclude Include="include\Poco\ActiveMethod.h" />
+    <ClInclude Include="include\Poco\AsyncObserver.h" />
     <ClInclude Include="include\Poco\ActiveResult.h" />
     <ClInclude Include="include\Poco\ActiveRunnable.h" />
     <ClInclude Include="include\Poco\ActiveStarter.h" />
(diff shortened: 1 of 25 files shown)


matejk (approved):
Looks OK to me.

Issue #4453, opened 2024-02-13
SQLParser not included in the release archive


aleks-f:
#4452 
#4455

PR #4456, opened 2024-02-13
SQLParser not included in the release archive #4453

aleks-f:


Data/include/Poco/Data/SQLParser.h
matejk, line 1: This file can be removed.

PocoDoc/cfg/mkdoc-poco.xml
```diff
@@ -7,7 +7,7 @@
 				${PocoBuild}/*/include/Poco/*/*/*.h
 				${PocoBuild}/*/include/Poco/*/*.h
 				${PocoBuild}/*/*/include/Poco/*/*/*.h
-				${PocoBuild}/Data/src/sql-parser/*.h
+				${PocoBuild}/Data/SQLParser/*.h
```
matejk, line 10: Is it necessary to have SQL parser includes for user documentation generation?

Data/ODBC/Makefile
```diff
@@ -14,7 +14,7 @@ objects = Binder ConnectionHandle Connector EnvironmentHandle \
 
 target_includes = $(POCO_BASE)/Data/testsuite/include
 ifndef POCO_DATA_NO_SQL_PARSER
-	target_includes += $(POCO_BASE)/Data/src
+	target_includes += $(POCO_BASE)/Data/SQLParser $(POCO_BASE)/Data/SQLParser/src
```
matejk, line 17: No need to add SQL parser includes here.

matejk (changes requested):
Some simplifications/cleanup can be done. See other comments.

Data/include/Poco/Data/SQLParser.h
aleks-f: It is [used](https://github.com/pocoproject/poco/blob/a3ced88384be4e070a26c397b39281600ded3337/Data/testsuite/src/SQLParserTest.cpp#L18) by testsuite.

aleks-f: actually, it could be removed, probably best to do s

PocoDoc/cfg/mkdoc-poco.xml
aleks-f: Probably best to remove it too

matejk (approved):

PR #4463, opened 2024-02-16
Discourage using configure/make to build Poco

matejk:
CMake is officially supported build system for building Poco.

To discourage using configure and make, a warning is printed and the user has to explicitly force usage of configure script.


aleks-f (commented):
I don't know about this. I don't want this obstacle for my own reasons but, more importantly, likely most packagers use make. 

I suggest to just make sure the documentation is kept accurate and clear about cmake being the official build system
https://github.com/pocoproject/poco/blob/88be66972a07968f027b68860add1d43aa5f135c/doc/00200-GettingStarted.page#L246

matejk:
Linux packagers seem to use cmake (at least Fedora, Debian and Ubuntu).

What about having just a message?

aleks-f:
> Linux packagers seem to use cmake (at least Fedora, Debian and Ubuntu).
> 
> What about having just a message?

Message only is fine, I think the documentation is the most important part, so we have no strict obligation to keep it absolutely immaculate - just RTFM anyone complaining about it.

PR #4472, opened 2024-02-26
Ability to Specify CMake NO_SONAME

### LINKED ISSUE #4466: Add Support for CMake NO_SONAME Linker Flag
We include the Poco .so files together with our software with just .so (no .so.94 file).

A new flag `POCO_USE_NO_SONAME` that defaults to false and is passed to the `set_target_properties` as: `NO_NAME POCO_USE_NO_SONAME`.

**Describe alternatives you've considered**
None that I can think of. The only other way to do this is for me to manually change the all the Poco CMakeLists.txt files each time we update. There is no way to override this flag in the cmake command.

**Additional context**
This is the change I've made to all the Poco targets we use. The actual change to Poco would be configurable through the `POCO_USE_NO_SONAME` flag.
```
set_target_properties(Foundation
	PROPERTIES
	VERSION ${SHARED_LIBRARY_VERSION} SOVERSION ${SHARED_LIBRARY_VERSION}
	OUTPUT_NAME PocoFoundation
	DEFINE_SYMBOL Foundation_EXPORTS
	NO_SONAME TRUE # We don't want to generate an .so and .so.94, just an .so because we install Poco directly alongside our software
)
```

I'll be submitting a PR with these changes when I get a chance to write them properly.

andrewauclair:
This changes adds the ability to set the cmake NO_SONAME target property on Poco targets. Fixes #4466.


aleks-f (commented):
should there be a minimal CI run to test that this actually does what it is supposed to do?

andrewauclair:
Should at least some CI checks to make sure the default behavior still generates named so files (numbered in this case, .so.101). I don't think it's critical to check that it works properly with `POCO_NO_SONAME=ON`. Unless there's some way to build a single project (i.e. Foundation) and verify on a .so file is created.

aleks-f:
> Unless there's some way to build a single project (i.e. Foundation) and verify on a .so file is created.

There is a way - disable everything except CppUnit and Foundation, execute a command verifying that relevant files (don't) exist, and run foundation tests to make sure everything executes as expected.

The type and structure of the CI job would be similar to this one:

https://github.com/pocoproject/poco/blob/c624b2787849a92e5bcb4eb9ec67e6e363bbadad/.github/workflows/ci.yml#L149-L164

andrewauclair:
After trying this for a while and running into RPATH issues, I'm thinking this might have been a dumb idea to do in the first place. Just because it's fewer files to install doesn't mean it's better. I should switch back to using sonames.

Issue #4482, opened 2024-03-05
ProcessRunner does not detect launch errors


aleks-f:
**Describe the bug**
Launch thread silently ignores any errors or exceptions.

**To Reproduce**
Launch a non-existing executable.

EDIt: there seems to be no reliable portable way to detect what exactly happens with the launched process (not even when the executable does not exist). Seems best to add the executable file existence check which includes `cwd` and environment path, and not even try to launch a non-existent executable.

matejk:
Fixed in #4483

PR #4483, opened 2024-03-05
enh(ProcessRunner): does not detect launch errors #4482

aleks-f:
TODO: detect properly if executable does not exist


matejk:
@aleks-f, I think that the solution implemented in the pushed works properly on *nix platforms. I need to finalise some details for Windows.

matejk:
@aleks-f, can this be merged to main?

Issue #4431, opened 2024-01-31
Consolidate LogFile implementation to use FileOutputStream


matejk:
`LogFile` has different implementations for Windows using Win32 API and STD for other platforms where `Poco::FileOutputStream` is used.  `Poco::FileOutputStream` has Windows and Posix implementations.

`LogFile` calls `FlushFileBuffers` on Windows which forcefully flushes buffers to disk. Posix equivalent would be `fsync`.

* New function `fsync` (or `flushToDisk`) shall be added to `FileStream_POSIX` and `FileStream_WIN32`.
* Implementation of `LogFile` shall be unified to use `FileOutputStream`

Related: #2443, #4429.

matejk:
Logically, there should be only LogFile, because we have POSIX and WIN32 FileStream; performance may be the reason for the current state.

Some benchmarking would help to make the right decision, ie. whether it makes sense to introduce `LogFile_POSIX`, instead of having just one `LogFile`.

As for flush/sync, those should be two separate properties and they should have strategies with reasonable defaults.

PR #4484, opened 2024-03-07
LogFile uses FileOutputStream instead of custom platform implementations (#4431)

matejk:
#4431.

`FileOutputStream` implements platform specific streams that are used by `LogFile`. There is no need to have additional platform specific implementations.

The only missing functionality in `FileStream` was flushing to disk.

Flushing on Posix vs Win32 was semantically different (iostream flush vs file system flush and was resolved in #4429). Flush option now means file system flush.

Since file system flush has major impact on write performance, the default flush setting is now false.


aleks-f (approved):

obiltschnig:
Note that there's also a difference in line endings on Windows. `FileOutputStream` always uses `\n` only. The Windows-specific `LogFile` also properly translates `\n` to `\r\n`. So this will break things on Windows.

matejk:
Thanks, @obiltschnig. i missed that detail.

I replaced hardcoded `'\n'` with `POCO_DEFAULT_NEWLINE_CHARS`.

Issue #3997, opened 2023-04-04
operator<<() for SocketAddress doesn't work with ADL


vehlwn:
Poco 1.12.4.

Operator<<() declared [here](https://github.com/pocoproject/poco/blob/1211613642269b7d53bea58b02de7fcd25ece3b9/Net/include/Poco/Net/SocketAddress.h#L304) is not findable using ADL because it's declared in global namespace and not in a namespace of SocketAddress. Due to that it's not possible to output SocketAddress into BOOST_LOG_TRIVIAL because it expects to find it using ADL. See the beginning of issue [here](https://github.com/boostorg/log/issues/207).

Is this operator<<() even part of public API? I couldn't find any references to it neither on the doc page for [SocketAddress](https://docs.pocoproject.org/current/Poco.Net.SocketAddress.html) nor Poco::Net [namespace](https://docs.pocoproject.org/current/Poco.Net.html).

PR #4491, opened 2024-03-12
Moving operator<< and operator>> overloads into the Poco::Net namespace.

andrewauclair:
Addresses the issues in #3997.


andrewauclair:
Looks like I missed something. I'll investigate why gcc doesn't like this.

andrewauclair:
The issue was the `operator<<` for `MacAddress`. I forgot that `MacAddress` is just an alias of `std::vector<unsigned char>`. Now that I see this, I think it's a bit wrong for Poco to define a `operator<<` for `std::vector<unsigned char>`.

PR #4494, opened 2024-03-13
PocoDoc: search support

vrabac6:
The PocoDoc did not support integration with FTS5 SQLite database for the purpose of inputting clean content into the database. To enhance its functionality, the following improvements have been implemented:

1. Expanded the `writeDescription` method by adding an output stream containing clean content.
2. Introduced a new method named `writeSearchIndex` , responsible for retrieving textual documentation along with links. This method is called within existing methods `writeClass` ,  `writePackage` , and `writeNameSpace`


aleks-f (changes requested):
- [ ] Windows CI is failing; this needs update of vs90 project and [regeneration](https://github.com/pocoproject/poco/wiki/Generating-VS-Projects-using-ProGen) of 160 and 170 projects
  ```  DocWriter.cpp
  D:\a\poco\poco\PocoDoc\src\DocWriter.cpp(29,10): error C1083: Cannot open include 
  file: 'Poco/Data/Session.h': No such file or directory 
  [D:\a\poco\poco\PocoDoc\PocoDoc_vs170.vcxproj]
  Error: Final attempt failed. Child_process exited with error code 1```
- [ ] to prevent sanitizer CI failures, either pick [these changes](https://github.com/pocoproject/poco/commit/03c2fa3eea16d2716f596bdb0393a5036b819980) or simply merge `devel` branch
- [ ] search support should be a runtime option
  - there should be a command line parameter added to `PocoDoc` to generate searchable SQLite DB
  - obviously, a precondition is SQLite built with FTS support
  - there should be a way to check at runtime if `Data/SQLite` supports `FTS`; only if it does can searchable content be produced (otherwise runtime error if searchable cmd line param was give)

PocoDoc/src/PocoDoc.cpp
```diff
@@ -351,7 +346,23 @@ class PocoDocApp: public Application
 		File file(path);
 		file.createDirectories();
 
-		DocWriter writer(_gst, path.toString(), config().getBool("PocoDoc.prettifyCode", false), _writeEclipseTOC);
+		bool searchIndex = false;
+
+		if (config().getBool("PocoDoc.searchIndex", false))
```
aleks-f, line 351: this is fine, but there should also be possible to [specify it on the command line](https://github.com/vrabac6/poco/blob/21ed183dc1697784bd9d99cab4c18ee8a3c73be9/PocoDoc/src/PocoDoc.cpp#L142):

aleks-f (changes requested):

aleks-f (changes requested):
- [x] this PR is from `devel` branch, which was retired 6 months ago; `pocoproject/main` branch needs to be merged here before this PR is merged
- [ ] FTS5 is available only on Android API Level > 24; see with @zaleksandar if this makes it worth keeping the FTS3/4 support alongside FTS5

Overall consolidation of the FTS capabilities is needed (this should probably be a separate issue):
- [ ] there are discrepancies/inaccuracies in the build systems:
  - [ ] no need for `=<path>` [here](https://github.com/vrabac6/poco/blob/75e02e4c8c4acec0f4cc715c88e8f2da9ebe7403/configure#L85)
  - [ ] [this](https://github.com/vrabac6/poco/blob/75e02e4c8c4acec0f4cc715c88e8f2da9ebe7403/configure#L258) looks like mixing incompatible defines (probably the second one has no effect)
  - [ ] apparently, VS projects have [FTS3 define hardcoded](https://github.com/vrabac6/poco/blob/75e02e4c8c4acec0f4cc715c88e8f2da9ebe7403/Data/SQLite/SQLite_vs160.vcxproj#L231); on Windows FTS5 should be hardcoded in the projects (regenerated from [SQLite.progen](https://github.com/vrabac6/poco/blob/75e02e4c8c4acec0f4cc715c88e8f2da9ebe7403/Data/SQLite/SQLite.progen#L11), not edited directly)
  - [ ] SQLite test has [hardcoded FTS3](https://github.com/vrabac6/poco/blob/75e02e4c8c4acec0f4cc715c88e8f2da9ebe7403/Data/SQLite/testsuite/src/SQLiteTest.cpp#L3461)
  - [ ] CMake build system should be adjusted to be in accordance with make build, coordinate with @matejk

vrabac6:
> * [ ]  FTS5 is available only on Android API Level > 24; see with @zaleksandar if this makes it worth keeping the FTS3/4 support alongside FTS5
> 

According to the latest [Android documentation](https://developer.android.com/google/play/requirements/target-sdk), all new apps must target API level 34, and even older apps now target a minimum of API level 31. This makes it effectively impossible for apps to run on API levels lower than 24. Therefore, it should be safe and efficient to always enable FTS5 without needing to maintain support for FTS3/4.

PocoDoc/Makefile
```diff
@@ -10,6 +10,10 @@ objects = DocWriter PocoDoc
 
 target         = PocoDoc
 target_version = 1
-target_libs    = PocoCppParser PocoFoundation PocoXML PocoUtil
+target_libs    = PocoCppParser PocoDataSQLite PocoData PocoFoundation PocoXML PocoUtil
```
matejk, line 13: Newly added data libraries shall be added to CMake also. PocoDoc doers not compile with CMake currently.

PocoDoc/Makefile
```diff
@@ -10,6 +10,10 @@ objects = DocWriter PocoDoc
...
+
+ifdef POCO_ENABLE_SQLITE_FTS5
+CXXFLAGS += -DPOCO_ENABLE_SQLITE_FTS5
```
matejk, line 16: Same here to enable full text search.

matejk (changes requested):

matejk:
[PocoDoc.cmake.patch](https://github.com/user-attachments/files/17615798/PocoDoc.cmake.patch)

Patch to correct CMake files for text search.

PocoDoc/Makefile
vrabac6: Fixed in ef13d0fab675b72f6db2bc185d1ecab1a2494b31

PocoDoc/Makefile
vrabac6: Fixed in ef13d0fab675b72f6db2bc185d1ecab1a2494b31

matejk (approved):

PR #4512, opened 2024-03-26
4435 secure sock thread

aleks-f:


NetSSL_OpenSSL/src/SecureSocketImpl.cpp
```diff
@@ -300,6 +306,9 @@ void SecureSocketImpl::shutdown()
 			int rc = ::SSL_shutdown(_pSSL);
 #endif
 			if (rc < 0) handleError(rc);
+
+			l.unlock();
```
matejk, line 310: Why is mutex unlocked explicitly before end of scope only in this function and not also others?

NetSSL_OpenSSL/include/Poco/Net/SecureSocketImpl.h
```diff
@@ -284,16 +284,21 @@ class NetSSL_API SecureSocketImpl
...
 
-	SSL* _pSSL;
+	std::atomic<SSL*> _pSSL;
 	Poco::AutoPtr<SocketImpl> _pSocket;
 	Context::Ptr _pContext;
 	bool _needHandshake;
 	std::string _peerHostName;
 	Session::Ptr _pSession;
 	bool _bidirectShutdown = true;
+	mutable MutexT _mutex;
```
matejk, line 301: Is `_mutex` intended to protect only _pSSL or also other related members (`_pContext`, `_pSession`)?

matejk (approved):
Changes seem OK to me. I added few comments where some things were not clear to me.

NetSSL_OpenSSL/include/Poco/Net/SecureSocketImpl.h
aleks-f: mutex protects all openssl calls; `_pSSL` pointer was made atomic, smart pointers have atomic refcounting

PR #4522, opened 2024-04-05
Post 1.13.3 sync master to devel

matejk:
Merge all of the remaining changes from master to devel after release 1.13.3 is merged to master.


DLLVersion.rc
```diff
@@ -4,8 +4,8 @@
 
 #include "winres.h"
 
-#define POCO_VERSION 1,13,1,0
-#define POCO_VERSION_STR "1.13.1"
+#define POCO_VERSION 1,13,2,0
```
aleks-f, line 7: shouldn't this be `1,13,3,0`?

aleks-f (changes requested):

DLLVersion.rc
matejk: Version 1.13.3 was released with this wrong version in the rc file.

obiltschnig: I re-released it this morning, hoping that none of the packagers has yet used the faulty packages ;-)

PR #4530, opened 2024-04-13
4529 sql logger shutdown

aleks-f:


Data/include/Poco/Data/SQLChannel.h
```diff
@@ -233,7 +243,7 @@ class Data_API SQLChannel: public Poco::Channel, Poco::Runnable
 	void reconnect();
 		/// Closes and opens the DB connection.
 
-	bool processOne(int minBatch = 0);
+	bool processBatch(int minBatch = 0);
 		/// Processes one message.
 		/// If the number of acummulated messages is greater
```
matejk, line 248: Does the comment still match the function processBatch()

Data/src/SQLChannel.cpp
```diff
@@ -301,7 +303,7 @@ void SQLChannel::run()
 		Thread::sleep(sleepTime);
```
matejk, line 303: Would it be possible to use conditions instead of sleep?

Data/src/SQLChannel.cpp
```diff
@@ -291,32 +300,251 @@ void SQLChannel::run()
...
+			pFileChannel->log(msg);
+			n += batch;
+			_logged += batch;
+			_source.erase(_source.begin(), _source.begin()+batch);
+			_pid.erase(_pid.begin(), _pid.begin() + batch);
+			_thread.erase(_thread.begin(), _thread.begin() + batch);
+			_tid.erase(_tid.begin(), _tid.begin() + batch);
+			_priority.erase(_priority.begin(), _priority.begin() + batch);
+			_text.erase(_text.begin(), _text.begin() + batch);
+			_dateTime.erase(_dateTime.begin(), _dateTime.begin() + batch);
```
matejk, line 428: Can code duplication be reduced with a lambda?

matejk (approved):
I think that PR is OK.

Few comments for minor improvements added.

Data/src/SQLChannel.cpp
aleks-f: I used event with interruptible wait (to make sure timed flush occurs), good suggestion

Issue #4537, opened 2024-04-22
Net: Socket::available causes connection reset on Windows UDP sockets


andrewauclair:
**Describe the bug**
Created a DatagramSocket with local address 127.0.0.1:0 and remote address 127.0.0.1:2345. After calling `sendBytes` on this UDP socket, `poll` returns true, then `available` returns -1 and the last error is "connection reset by peer". I suspect this is from the `recvfrom` call within `available`. If I change this to use the IP of a remote machine as the remote address, poll properly times out and returns false after `sendBytes`.

**To Reproduce**
```cpp
#include <iostream>
void DatagramSocketTest::testLocalIPIssue()
{
	DatagramSocket socket(SocketAddress("127.0.0.1", 0), true, true);

	socket.connect(SocketAddress("127.0.0.1", 2345));

	unsigned char values[5]{};
	socket.sendBytes(values, 5);

	Poco::Timespan timeout(5000);

	if (socket.poll(timeout, Socket::SELECT_READ))
	{
		std::cout << socket.available(); // -1 and last error is "connection reset by peer"
	}
}
```

**Expected behavior**
Poll returns false, just like it would when local and remote IPs don't match.

**Please add relevant environment information:**
 - Windows 10
 - POCO 1.12.5

andrewauclair:
Should we be using `Poco::Net::UDPServer` and `Poco::Net::UDPClient` instead of a `Poco::Net::DatagramSocket` directly to avoid issues like this?

andrewauclair:
I ended up closing and reopening the socket if there was an error after calling `available()`. It still doesn't seem that great to me that `available()` could cause an error and return -1.

andrewauclair:
Some additional testing resulted in `receiveBytes` throwing a `std::bad_array_new_length` because `available` returned -1.

andrewauclair:
Still an issue.

PR #4539, opened 2024-04-25
Throw Error in Socket::available if Socket Error Occurs 

andrewauclair:
Changes to help address the issues I was seeing in #4537. This change makes `SocketImpl::available` throw an error exception if the result of `recvfrom` is negative and there was a socket error for UDP sockets. This will correctly inform the calling code that the socket is invalid and needs to be reset. This also fixes the issue with `receiveBytes` throwing `std::bad_array_new_length` because it will instead throw a Poco exception first when calling `available`.

### DIFF
--- a/Net/src/SocketImpl.cpp
+++ b/Net/src/SocketImpl.cpp
@@ -653,6 +653,8 @@ int SocketImpl::available()
 	{
 		std::vector<char> buf(result);
 		result = recvfrom(sockfd(), &buf[0], result, MSG_PEEK, nullptr, nullptr);
+
+		if (result < 0) error();
 	}
 #endif
 	return result;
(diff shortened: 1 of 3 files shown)


aleks-f (changes requested):
I generally don't like just querying for something throwing an exception. 

I would prefer `available` remaining as is, and instead document that negative value means invalid socket and fix all the affected code to take that into account (this may be an exception where it makes sense).

andrewauclair:
I can change it back and throw exceptions in places that poco calls available. If that's what you want. I do feel that the exception forces the issue better. But I agree with your statement about querying functions.

matejk:
Superseded by #5127, which implements the approach suggested in the review (fix callers instead of having `available()` throw).

PR #4542, opened 2024-04-27
Decode word

bas524:
RFC 2047 decodeWord
uses separate string which contains simbols between chunks and if this string contains only space ot \n or \t or \v than trim it


matejk:
I verified the changes with the online tool https://dogmamix.com/MimeHeadersDecoder/. Changes are correct.

Issue #4544, opened 2024-04-30
The load balancing issue in Poco::ActiveThreadPool


siren186:
**Describe the bug**
Each thread in `Poco::ActiveThreadPool` has its own `Poco::NotificationQueue`.
If I have 2 types of Runnable like this:
```cpp
class LongTimeTask : public Poco::Runnable {
    void run() {
        sleep(2 * 60 * 1000);
    }
};

class ShortTimeTask : public Poco::Runnable {
    void run() {
        sleep(1);
    }
};

int capacity = 2;
Poco::ActiveThreadPool pool(capacity);

for (int i = 0; i < 10; i++) {
    pool.start(new LongTimeTask); // [BUG!!!] all long time task will enqueue to thread#1
    pool.start(new ShortTimeTask); // [BUG!!!] all short time task will enqueue to thread#2
}
```

bas524:
Hi!
I can't agree with a bug.
I verified behaivour on the latest main branch with this code
```cpp
	Poco::AtomicCounter lttCount;
	class LongTimeTask : public Poco::Runnable {
		Poco::AtomicCounter &_counter;
	public:
		LongTimeTask(Poco::AtomicCounter &counter) : _counter(counter) {}
		void run() override {
			_counter++;
			puts("ltt task");
			Poco::Thread::sleep(2 * 1000);
		}
	};
	Poco::AtomicCounter sttCount;
	class ShortTimeTask : public Poco::Runnable {
		Poco::AtomicCounter &_counter;
	public:
		ShortTimeTask(Poco::AtomicCounter &counter) : _counter(counter) {}
		void run() override {
			_counter++;
			puts("stt task");
			Poco::Thread::sleep(1);
		}
	};
	
	int capacity = 2;
	Poco::ActiveThreadPool pool(capacity);
	
	for (int i = 0; i < 10; i++) {
		LongTimeTask ltt(lttCount);
		pool.start(ltt);
		ShortTimeTask stt(sttCount);
		pool.start(stt);
	}
	pool.joinAll();
	assertEqual(10, lttCount.value());
	assertEqual(10, sttCount.value());
```
Output was
```shell
testActiveThreadLoadBalancing: ltt task
stt task
stt task
stt task
stt task
stt task
stt task
stt task
stt task
stt task
stt task
ltt task
ltt task
ltt task
ltt task
ltt task
ltt task
ltt task
ltt task
ltt task
```

siren186:
[...]

PR #4548, opened 2024-05-07
Motivated by "The load balancing issue in Poco::ActiveThreadPool" #4544

bas524:
Optimization allows redistribute tasks to the idle threads
When ActiveThread dequeue next Runnable, it checks is optimization is true and if yes it try to find idle thread and resend task
By default optimization is false

### DIFF
--- a/Foundation/include/Poco/ActiveThreadPool.h
+++ b/Foundation/include/Poco/ActiveThreadPool.h
@@ -41,16 +41,19 @@ class Foundation_API ActiveThreadPool
 	/// The thread pool always keeps fixed number of threads running.
 	/// Use case for this pool is running many (more than os-max-thread-count) short live tasks
 	/// Round-robin model allow efficiently utilize cpu cores
+	/// Using redistributeTasks option allows optimize reusage of idle threads
 {
 public:
 	ActiveThreadPool(int capacity = static_cast<int>(Environment::processorCount()) + 1,
-		int stackSize = POCO_THREAD_STACK_SIZE);
+		int stackSize = POCO_THREAD_STACK_SIZE,
+		bool redistributeTasks = false);
 		/// Creates a thread pool with fixed capacity threads.
 		/// Threads are created with given stack size.
 
 	ActiveThreadPool(std::string  name,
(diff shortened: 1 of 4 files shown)


matejk:
Thanks for the contribution.

IMO partial refactoring of ActiveThreadPool (and related classes) to have a single task queue in the pool where the threads take the work from would automatically create optimal load-balancing and simplify the implementation at the same time.

bas524:
> Thanks for the contribution.
> 
> IMO partial refactoring of ActiveThreadPool (and related classes) to have a single task queue in the pool where the threads take the work from would automatically create optimal load-balancing and simplify the implementation at the same time.

Hi!
I thougt about it. If notificationqueue was lock-free than you proposition shoud be correct. In our case realisation with single queue would contains many locks and bad performance. 
May be flat combinig implementation is better choise for us...

andrewauclair:
This change appears to fail with the Thread Sanitizer. The actual log has much more in it <https://github.com/pocoproject/poco/actions/runs/8982526627/job/24670333088?pr=4548>.

```
testActiveThreadLoadBalancing: ==================
WARNING: ThreadSanitizer: data race (pid=33153)
  Read of size 8 at 0x7ffe63aa0b60 by thread T16:
    #0 Poco::ActiveThread::run() <null> (libPocoFoundation.so.103+0xf9197)
    #1 <null> <null> (libPocoFoundation.so.103+0x1cc80e)
    #2 Poco::ThreadImpl::runnableEntry(void*) <null> (libPocoFoundation.so.103+0x1cdebd)
 ```

matejk:
@bas524  Does the PR #4624 with different implementation that was merged to main satisfy the needs?

bas524:
> @bas524 Does the PR #4624 with different implementation that was merged to main satisfy the needs?

Hi, I'll check it next week

bas524:
> @bas524 Does the PR #4624 with different implementation that was merged to main satisfy the needs?

@matejk 
I answered into https://github.com/pocoproject/poco/pull/4624#issuecomment-2408475465

Issue #4358, opened 2023-12-18
Reorganise source code of external libraries (pre-requisite for #4354)


matejk:
From #4354 by @teksturi :

Separate external source files to own folder

It would be quite hard to exclude every external source without moving them to different folder. This same thing will happened also with other static analyzer tools, formatting tools, compiler options. So if we move them separated folders we can always exclude them more easily.

We can consider couple options here (example for zlib). Note that "external" can also be replaced with some other word if wanted.

1. Foundation/external/
2. Foundation/external/zlib/
3. Foundation/src/external/
4. Foundation/src/external/zlib/
5. Foundation/src/zlib
6. external/
7. external/zlib

My personal vote goes to 2, 4 or 7. Probably 7 because that way it so easy to seperete everything to one place. We can have example update script which checks all new version for those.

Still need to think what about external files which are also in include/ there we would probably need external folder. Example of this kind of file is "Foundation/include/Poco/ordered_map.h"

Maybe something like this for external files.

![image](https://github.com/pocoproject/poco/assets/3656678/cfd173d5-fc59-465f-9ec7-09027a3a6983)

I still need to prototype a little bit. We probably still do not want to build these as libs as before. Also we probably want to have nice include paths

#include "Poco/external/pdjson/pdjson.h"
We could also do some magic that we can always do that even when POCO_UNBUNDLED is used. But that is not very high priority but would be nice.

---

Structure reorganization is needed if we want to have clang-tidy automated in good way. Else we will clatter with exclude rules. That would be same work as moving files to different location. Moving them will also help in other stuff like taking clang-format in use.

Easiest reorganization would be this

```
Foundation/src/External/
Foundation/include/Poco/External/
```

This way almost nothing change but we can exclude those pretty easily.

aleks-f:
I'd name the directory `ext`

PR #4579, opened 2024-06-07
RFC: Hide zlib and expat libs from the user of Poco libraries

teksturi:
Seperete zlib and expat so they are only expose internally. There should not be reason for user to use these.

This makes https://github.com/pocoproject/poco/issues/4358 easier to implement as all external libraries are not exposed to end user.


teksturi:
Please say if we example still want to expose ParserEngine. I will try to make necessary changes for that. I do not remember exactly if that would be easy as this has been quite long in my POC box.

Foundation/include/Poco/DeflatingStream.h
```diff
@@ -104,7 +101,7 @@ class Foundation_API DeflatingIOS: public virtual std::ios
 	/// order of the stream buffer and base classes.
 {
 public:
-	DeflatingIOS(std::ostream& ostr, DeflatingStreamBuf::StreamType type = DeflatingStreamBuf::STREAM_ZLIB, int level = Z_DEFAULT_COMPRESSION);
+	DeflatingIOS(std::ostream& ostr, DeflatingStreamBuf::StreamType type = DeflatingStreamBuf::STREAM_ZLIB, int level = -1);
```
andrewauclair, line 104: Going from `Z_DEFAULT_COMPRESSION` to just `-1` is a loss of info for users. Can we add a `POCO_DEFAULT_COMPRESSION` const for this?

Foundation/include/Poco/InflatingStream.h
```diff
@@ -22,11 +22,8 @@
 #include "Poco/BufferedStreamBuf.h"
 #include <istream>
 #include <ostream>
-#if defined(POCO_UNBUNDLED)
-#include <zlib.h>
-#else
-#include "Poco/zlib.h"
-#endif
+
+struct z_stream_s;
```
andrewauclair, line 26: Don't think this really needs to be here. Just do `struct z_stream_s* _pZstr;` in the class (unless I'm not seeing another usage here).

Foundation/src/DeflatingStream.cpp
```diff
@@ -25,29 +31,17 @@ DeflatingStreamBuf::DeflatingStreamBuf(std::istream& istr, StreamType type, int
...
-	_zstr.data_type = 0;
-	_zstr.adler     = 0;
-	_zstr.reserved  = 0;
-
-	_buffer = new char[DEFLATE_BUFFER_SIZE];
-
-	int rc = deflateInit2(&_zstr, level, Z_DEFLATED, 15 + (type == STREAM_GZIP ? 16 : 0), 8, Z_DEFAULT_STRATEGY);
+	std::unique_ptr<char[]> buffer(new char[DEFLATE_BUFFER_SIZE]);
+
+	std::unique_ptr<z_stream> pZstr = std::make_unique<z_stream>(z_stream{});
```
matejk, line 36: The argument `z_stream{}` to `make_unique` seems wrong. Parameters for the ctor shall be passed there, probably the ones used before to initialize `_zstr`.

Foundation/testsuite/src/ZLibTest.cpp
```diff
@@ -17,6 +17,12 @@
 #include "Poco/StreamCopier.h"
 #include "Poco/Buffer.h"
 #include <sstream>
+#if defined(POCO_UNBUNDLED)
+#include <zlib.h>
+#else
+// Quirk before we move zlib to external libs.
+#include "../../src/zlib.h"
```
matejk, line 24: The path to zlih.h shall be added CMake's include path instead.

Foundation/src/DeflatingStream.cpp
andrewauclair: Assuming `Z_NULL` isn't something special and is just a `nullptr` (I would have to check), then this does the same thing. `z_stream{}` value-initializes a `z_stream` and `std::make_unique<z_stream>(z_stream{})` calls the move constructor.

matejk: Thanks, I was not sure if default init of integers sets the values to 0.

andrewauclair: Default init does not, but value init does. `int i;` is uninitialized, but `int i{};` is 0.

matejk:
@teksturi , any updates on this?

Foundation/include/Poco/DeflatingStream.h
matejk: There is an issue with other (non-default) compression levels whenever the user decided to go with non-default.

Shall Poco maintain a copy of compression values that are identical with the ones in zlib?

@andrewauclair , @teksturi

matejk:
Replaced with #4724.

Issue #4592, opened 2024-06-22
Significant performance degradation of Poco::DateTimeParser


tyler92:
Consider the following code:

```cpp
#include <Poco/DateTimeFormat.h>
#include <Poco/DateTimeFormatter.h>
#include <Poco/DateTimeParser.h>
#include <stdexcept>

std::string toString(const Poco::DateTime& dt)
{
    return Poco::DateTimeFormatter::format(dt, Poco::DateTimeFormat::ISO8601_FRAC_FORMAT);
}

Poco::DateTime parsePoco(const std::string& text)
{
    int tzd = 0;
    const auto format = Poco::DateTimeFormat::ISO8601_FRAC_FORMAT;
    auto parsed = Poco::DateTimeParser::parse(format, text, tzd);
    parsed.makeUTC(tzd);
    return parsed;
}

void ASSERT_EQ(const std::string& a, const std::string& b)
{
    if (a != b)
    {
        throw std::runtime_error("Test FAILED");
    }
}

int main()
{
    for (int i = 0; i < 100'000; ++i)
    {
        auto parsed = parsePoco("2024-06-21T09:43:11.340159Z");
        ASSERT_EQ(toString(parsed), "2024-06-21T09:43:11.340159Z");

        parsed = parsePoco("2024-06-21T09:43:11.340159+07:00");
        ASSERT_EQ(toString(parsed), "2024-06-21T02:43:11.340159Z");
    }

    return 0;
}
```

Execution time (Release configuration):

```
Poco 1.12.5:    96 ms
Poco 1.13.3: 17179 ms
```

With the format `Poco::DateTimeFormat::ISO8601_FRAC_FORMAT + " "` (with an additional space) execution time becomes better for 1.13.3.

I guess the reason is the following commit: https://github.com/pocoproject/poco/commit/4f1cf683079cc3b00be96259fc42f5c291ab4c77 (https://github.com/pocoproject/poco/pull/4330).

Up to 8 regex expressions are compiled on each call

https://github.com/pocoproject/poco/blob/4f1cf683079cc3b00be96259fc42f5c291ab4c77/Foundation/src/DateTimeFormat.cpp#L153-L156

if the provided format is known:

https://github.com/pocoproject/poco/blob/4f1cf683079cc3b00be96259fc42f5c291ab4c77/Foundation/src/DateTimeParser.cpp#L47

aleks-f:
[...]

PR #4593, opened 2024-06-25
DateTimeParser Validation and Performance Improvements

andrewauclair:
Improvements for #4592.

Improve the validation of `DateTimeParser` by validating the numbers parsed for `day`, `month`, `year`, `hour`, `minute`, `second`, `millis`, and `micros`. The value 0 will be used if no digits were found for `micros`.

These changes create new instances of `SyntaxException` thrown by `DateTimeParser::parse` and `DateTimeParser::parseTZD` when the previously listed fields do not contain the required number of digits or the string of digits is not a valid number.

Improve performance by caching the `RegularExpression`s created in `DateTimeFormat::isValid`.

### DIFF
--- a/Foundation/src/DateTimeFormat.cpp
+++ b/Foundation/src/DateTimeFormat.cpp
@@ -129,19 +129,6 @@ const std::string DateTimeFormat::MONTH_NAMES[] =
 };
 
 
-DateTimeFormat::RegexList DateTimeFormat::REGEX_LIST =
-{
-	&DateTimeFormat::ISO8601_REGEX,
-	&DateTimeFormat::RFC822_REGEX,
-	&DateTimeFormat::RFC1123_REGEX,
-	&DateTimeFormat::HTTP_REGEX,
-	&DateTimeFormat::RFC850_REGEX,
-	&DateTimeFormat::RFC1036_REGEX,
-	&DateTimeFormat::ASCTIME_REGEX,
-	&DateTimeFormat::SORTABLE_REGEX
-};
-
-
 bool DateTimeFormat::hasFormat(const std::string& fmt)
 {
 	return FORMAT_LIST.find(fmt) != FORMAT_LIST.end();
@@ -150,8 +137,7 @@ bool DateTimeFormat::hasFormat(const std::string& fmt)
 
 bool DateTimeFormat::isValid(const std::string& dateTime)
 {
-
-        static const RegularExpression regs[] = {
+        static const RegularExpression regexList[] = {
                 RegularExpression(DateTimeFormat::ISO8601_REGEX),
                 RegularExpression(DateTimeFormat::RFC822_REGEX),
                 RegularExpression(DateTimeFormat::RFC1123_REGEX),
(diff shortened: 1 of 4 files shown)


Foundation/src/DateTimeFormat.cpp
```diff
@@ -150,11 +150,25 @@ bool DateTimeFormat::hasFormat(const std::string& fmt)
...
+                RegularExpression(DateTimeFormat::RFC822_REGEX),
+                RegularExpression(DateTimeFormat::RFC1123_REGEX),
+                RegularExpression(DateTimeFormat::HTTP_REGEX),
+                RegularExpression(DateTimeFormat::RFC850_REGEX),
+                RegularExpression(DateTimeFormat::RFC1036_REGEX),
+                RegularExpression(DateTimeFormat::ASCTIME_REGEX),
+                RegularExpression(DateTimeFormat::SORTABLE_REGEX)
+        };
+		// make sure the regex list and the array of regexes are in sync
+		poco_assert((sizeof(regs) / sizeof(regs[0])) == REGEX_LIST.size());
```
matejk, line 165: REGEX_LIST is used only in this place. I propose to remove it as a class member and have it as a static member in this function.

Foundation/src/DateTimeParser.cpp
```diff
@@ -18,42 +18,140 @@
 #include "Poco/Exception.h"
 #include "Poco/Ascii.h"
 #include "Poco/String.h"
+#include <iostream>
 
+namespace {
+	using parse_iter = std::string::const_iterator;
 
-namespace Poco {
+	[[nodiscard]] parse_iter skip_non_digits(parse_iter it, parse_iter end)
```
matejk, line 26: Shall these helper functions be inlined?

Foundation/src/DateTimeFormat.cpp
andrewauclair: I left it where it is because it's part of the public API of this class, unfortunately.

Wasn't sure if this would be 1.13.x or 1.14 and didn't want to make a breaking change just in case.

Foundation/src/DateTimeParser.cpp
andrewauclair: No. `inline` is used in headers to avoid ODR issues. Free functions in cpp files should be either static or in an anonymous namespace to avoid IFNDR issues.

Foundation/src/DateTimeFormat.cpp
aleks-f: @andrewauclair 1.14, you can break it

Foundation/src/DateTimeParser.cpp
aleks-f: @andrewauclair please don't introduce new naming conventions, [we have our own](https://github.com/pocoproject/poco/blob/4f1cf683079cc3b00be96259fc42f5c291ab4c77/doc/CppCodingStyleGuide.doc)

aleks-f (changes requested):

matejk:
@andrewauclair Do you have any performance measurements with these changes compared to 1.13?

matejk (approved):

andrewauclair:
> @andrewauclair Do you have any performance measurements with these changes compared to 1.13?

Came up with the following when running in WSL with the code from issue #4592:

1.12.5: 0m0.063s
1.13.3: 0m11.991s (almost 200x slower?)
this pr: 0m0.185s (3x slower, likely the regex match)

Switching to a format string that doesn't match one of the expected ones, to avoid regex matches (`const auto format = Poco::DateTimeFormat::ISO8601_FRAC_FORMAT + " "`):

1.12.5: 0m0.063s
1.13.3: 0m0.065s
this pr: 0m0.085s

PR #4599, opened 2024-07-09
Update Darwin : change Archflags to POCO_TARGET_OSARCH

nickorr:
I was having issues compiling poco for arm64 on an x86 macOS machine, as a static build.  The end result was always x86.  Change is from : 

ARCHFLAGS          ?= -arch $(POCO_HOST_OSARCH)
to
ARCHFLAGS          ?= -arch $(POCO_TARGET_OSARCH)

Otherwise this always compiles to the HOST OS instead of target.  I think this forces you to set POCO_TARGET_OSARCH on the make command, but I'm not sure how to otherwise set a default.

Also OSFLAGS should be able to be something overwritten from the make command, but not sure how to do that.  I was hoping to : 

make install -s -j4 POCO_CONFIG=Darwin64-clang-libc++ MACOSX_DEPLOYMENT_TARGET=10.15 POCO_TARGET_OSARCH=arm64

So 10.15 instead of 10.11.  Also I think the isysroot part of OSFLAGS can be left out, as it should be part of the default path now.

### DIFF
--- a/build/config/Darwin-clang-libc++
+++ b/build/config/Darwin-clang-libc++
@@ -11,7 +11,7 @@
 #
 LINKMODE ?= SHARED
 
-ARCHFLAGS          ?= -arch $(POCO_HOST_OSARCH)
+ARCHFLAGS          ?= -arch $(POCO_TARGET_OSARCH)
 SANITIZEFLAGS      ?=
 OSFLAGS            ?= -mmacosx-version-min=10.11 -isysroot $(shell xcrun --show-sdk-path)
 


aleks-f (changes requested):
this fails on mac ci

nickorr:
Ok, yes so my change breaks your tests because without it being included as an option to the compile step, host ends up not being set at all, and I was adding that in on the configure command line.  But my use case is in doing cross compiling : trying to build arm on x86, and your tests don't cover that - I can't tell what hardware your tests are run on either, it doesn't seem to be showing up in the logs.

So feel free to close this as it currently does break as it removes the host arch - which it obviously needs.  FWIW the issue I had was that to get an arm compile on x86, I had to add : 

POCO_HOST_OSARCH=arm64 POCO_TARGET_OSARCH=x86_64

When I believe it should be the other way around ( unless I don't understand the poco compling process ).  In compiling other libraries that required a --host flag, I've always used --host=x86_64-apple-darwin - that's why I figured it was the wrong way around.

Cheers,
Nick

aleks-f:
> Ok, yes so my change breaks your tests because without it being included as an option to the compile step, host ends up not being set at all, and I was adding that in on the configure command line. But my use case is in doing cross compiling : trying to build arm on x86, and your tests don't cover that - I can't tell what hardware your tests are run on either, it doesn't seem to be showing up in the logs.
> 
> So feel free to close this as it currently does break as it removes the host arch - which it obviously needs. FWIW the issue I had was that to get an arm compile on x86, I had to add :
> 
> POCO_HOST_OSARCH=arm64 POCO_TARGET_OSARCH=x86_64
> 
> When I believe it should be the other way around ( unless I don't understand the poco compling process ). In compiling other libraries that required a --host flag, I've always used --host=x86_64-apple-darwin - that's why I figured it was the wrong way around.
> 
> Cheers, Nick

we have cross compile CI:

https://github.com/pocoproject/poco/blob/0bce71de240c2ea77bc955a488cc42813b2c50d8/.github/workflows/ci.yml#L47

nickorr:
Thanks @aleks-f but as far as I can see, your Mac CI doesn't do cross compiling to x86 and arm from a single Mac platform.  Or am I missing something?

aleks-f:
> Thanks @aleks-f but as far as I can see, your Mac CI doesn't do cross compiling to x86 and arm from a single Mac platform. Or am I missing something?

It doesn't. The make build system is not officially supported, but we use it for development and obviously CI. Over the years, there were various approaches to cross compiling. There is `CROSS_COMPILE` on Linux and AIX, and there are `ARCHFLAGS` for arm builds. In any case, I do not have the bandwidth to worry about it right now. If you can produce something consistent with the rest of the build system that does not break the CI, I'll merge it; otherwise, it is what it is

PR #4615, opened 2024-07-28
Fix samples/testsuite link unbundled pcre error.

xiongyong86:
1. fix the samples/testsuite link unbundled pcre error.

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -417,6 +417,42 @@ protected:
 		to = static_cast<T>(from);
 	}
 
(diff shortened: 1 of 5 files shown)


xiongyong86:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已收到，我会尽快给您回复，谢谢！

matejk:
@xiongyong86, on which operating system was it required to do these changes?

aleks-f:
@matejk there should be at least one linux unbundled CI job, otherwise we're in the dark regarding these things

matejk:
Yes, it should be, you are right. I did not realise that there is none.

PR #4616, opened 2024-07-29
Cppunit and data test enhancements

matejk:
Several code improvements of `CppUnit` and `Data/Test` as a result of code review while trying to reproduce unit tests failing with `std::bad_function_call`.


CppUnit/include/CppUnit/Test.h
```diff
@@ -49,37 +49,8 @@ class CppUnit_API Test
 	std::vector<std::string> _setup;
 };
 
-
 inline Test::~Test() = default;
 
-
```
aleks-f, line 55: are we sure these are not called from anywhere? we may end up with pure virtual calls if they are

matejk: You are right. It seemed to me that those functions were left there by mistake. I'll revert this.

PR #1859, opened 2017-08-26
Modify validation check in Poco::DateTime

komainu8:
Dear, developers.

I tried improved validation check in `Poco::DateTime` and `Poco::DateTime.assign` according to issues #1540's ideas.

I modified that used exception(InvalidArgumentException) in place of `poco_assert`.

Issue #1540, opened 2017-01-04
Poco::DateTime uses assertions for validation


kimgr:
We're unit-testing some code based on `Poco::DateTime`.

When we throw invalid dates at our code, this is printed to the console:

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
Assertion violation: month >= 1 && month <= 12 [in file "src/DateTime.cpp", line 67]
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

It turns out because `DateTime` uses `poco_assert` for validation, there doesn't seem to be a clean way to validate date times without double-checking them.

Would you consider patches to throw a documented exception type instead? Any ideas on which type is preferable? Just a Poco::Exception?

obiltschnig:
Yes. Throwing Poco::InvalidArgumentException would be appropriate.

kimgr:
Thanks! Sorry I never got around to this, something more urgent came up.

Meanwhile, I read about narrow vs wide contracts [1] and started to think that maybe the `DateTime` constructor is better off with a narrow contract with undefined behavior in the presence of obviously bad data.

That said, Poco strikes me as trying to avoid undefined behavior for the most part, so an exception seems more in style. Let me know what you prefer.

[1] http://stackoverflow.com/a/16734278/96963

PureAbstract:
For what little it's worth, I'd prefer that the constructor validates and throws on error.
Other alternatives might be have an 'IsValid()' member that could be called by those of us who care about avoiding undefined behaviour - another option might be for callers to pass an error_code by reference (as in std::filesystem). 

Whichever way it's done, in a library like Poco, there should be some way for the caller to verify that DateTime is doing what they expect it to do.

SmileGobo:
The bug is still in 1.13.3

PR #4640, opened 2024-08-17
Poco::DateTime uses assertions for validation #1540

aleks-f:
Is [negative](https://github.com/pocoproject/poco/blob/e4aaeacea8872e9361f3d67217c5a254cf6fc773/Foundation/testsuite/src/DateTimeTest.cpp#L96) year valid or [not](https://github.com/pocoproject/poco/blob/e4aaeacea8872e9361f3d67217c5a254cf6fc773/Foundation/include/Poco/DateTime.h#L55)?

#1859

### DIFF
--- a/Foundation/src/DateTime.cpp
+++ b/Foundation/src/DateTime.cpp
@@ -203,8 +203,8 @@ int DateTime::daysOfMonth(int year, int month)
 
 	if (month == 2 && isLeapYear(year))
 		return 29;
-	else
-		return daysOfMonthTable[month];
+	else if (month < 1 || month > 12)
+		return 0;
 	return daysOfMonthTable[month];
 }
 
--- a/Foundation/src/DateTime.cpp
+++ b/Foundation/src/DateTime.cpp
@@ -30,6 +30,7 @@ DateTime::DateTime()
 	_utcTime = now.utcTime();
 	computeGregorian(julianDay());
 	computeDaytime();
+	checkValid();
 }
 
 
@@ -43,14 +44,9 @@ DateTime::DateTime(const tm& tmStruct):
 	_millisecond(0),
 	_microsecond(0)
 {
-	poco_assert (_year >= 0 && _year <= 9999);
-	poco_assert (_month >= 1 && _month <= 12);
-	poco_assert (_day >= 1 && _day <= daysOfMonth(_year, _month));
-	poco_assert (_hour >= 0 && _hour <= 23);
-	poco_assert (_minute >= 0 && _minute <= 59);
-	poco_assert (_second >= 0 && _second <= 60);
-
-	_utcTime = toUtcTime(toJulianDay(_year, _month, _day)) + 10*(_hour*Timespan::HOURS + _minute*Timespan::MINUTES + _second*Timespan::SECONDS);
+	checkValid();
+	_utcTime = toUtcTime(toJulianDay(_year, _month, _day)) +
+		10*(_hour*Timespan::HOURS + _minute*Timespan::MINUTES + _second*Timespan::SECONDS);
 }
 
 
@@ -59,6 +55,7 @@ DateTime::DateTime(const Timestamp& timestamp):
 {
 	computeGregorian(julianDay());
 	computeDaytime();
+	checkValid();
 }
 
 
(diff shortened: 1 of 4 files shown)


obiltschnig:
I think we should support the full range of the Julian calendar starting Monday, January 1, 4713 BC (year -4712) as conversion from Julian day number may produce such dates.

Foundation/testsuite/src/DateTimeTest.cpp
```diff
@@ -92,7 +92,7 @@ void DateTimeTest::testJulian()
 	assertTrue (dt.millisecond() == 0);
 	assertTrue (dt.dayOfWeek() == 5);
 	assertTrue (dt.julianDay() == 2299160.5);
-
+/*
```
matejk, line 95: Is this commented out part replaced with `testInvalid()`?

Foundation/src/DateTime.cpp
```diff
@@ -145,6 +126,7 @@ DateTime& DateTime::operator = (const DateTime& dateTime)
 		_millisecond = dateTime._millisecond;
 		_microsecond = dateTime._microsecond;
 	}
+	checkValid();
```
matejk, line 129: This check is redundant, because the original is valid already.

Foundation/src/DateTime.cpp
```diff
@@ -154,6 +136,7 @@ DateTime& DateTime::operator = (const Timestamp& timestamp)
 	_utcTime = timestamp.utcTime();
 	computeGregorian(julianDay());
 	computeDaytime();
+	checkValid();
```
matejk, line 139: Is this check intended to verify the conversion of `Timestamp` to `DateTime` because this is the only way that `DateTime` can be invalid.

Foundation/src/DateTime.cpp
```diff
@@ -162,21 +145,13 @@ DateTime& DateTime::operator = (double julianDay)
 {
 	_utcTime = toUtcTime(julianDay);
 	computeGregorian(julianDay);
+	checkValid();
```
matejk, line 148: IMO this one is redundant unless it checks the conversion.

Foundation/src/DateTime.cpp
```diff
@@ -43,14 +44,9 @@ DateTime::DateTime(const tm& tmStruct):
...
 {
-	poco_assert (_year >= 0 && _year <= 9999);
-	poco_assert (_month >= 1 && _month <= 12);
-	poco_assert (_day >= 1 && _day <= daysOfMonth(_year, _month));
-	poco_assert (_hour >= 0 && _hour <= 23);
-	poco_assert (_minute >= 0 && _minute <= 59);
-	poco_assert (_second >= 0 && _second <= 60);
-
-	_utcTime = toUtcTime(toJulianDay(_year, _month, _day)) + 10*(_hour*Timespan::HOURS + _minute*Timespan::MINUTES + _second*Timespan::SECONDS);
+	checkValid();
```
matejk, line 47: Should `checkValid` be used after initialising `_utcTime`?

Foundation/src/DateTime.cpp
```diff
@@ -59,6 +55,7 @@ DateTime::DateTime(const Timestamp& timestamp):
 {
 	computeGregorian(julianDay());
 	computeDaytime();
+	checkValid();
```
matejk, line 58: Same comment as for the assignment.

matejk (commented):
I think that some of the checks are redundant.

matejk:
> I think we should support the full range of the Julian calendar starting Monday, January 1, 4713 BC (year -4712) as conversion from Julian day number may produce such dates.

I agree to this statement.

Foundation/src/DateTime.cpp
aleks-f: yes

Foundation/src/DateTime.cpp
aleks-f: it makes sure the passed julian day gives a valid DateTime

aleks-f:
> > I think we should support the full range of the Julian calendar starting Monday, January 1, 4713 BC (year -4712) as conversion from Julian day number may produce such dates.
> 
> I agree to this statement.

If min year is -4712, then we have a bug, because julian 0.0 produces year -4713

Foundation/src/DateTime.cpp
aleks-f: `checkValid` does not check utc and it is pointless to calculate utc if something is invalid

matejk:
> > > I think we should support the full range of the Julian calendar starting Monday, January 1, 4713 BC (year -4712) as conversion from Julian day number may produce such dates.
> > 
> > 
> > I agree to this statement.
> 
> If min year is -4712, then we have a bug, because julian 0.0 produces year -4713

According to https://aa.usno.navy.mil/calculated/calendardate?ID=AA&jd=0.0&submit=Get+Date, Julian date 0.0 is

```
Monday	4713 B.C. January 1	12:00:00.0	0.000000
```

However, according to https://apps.aavso.org/v2/tools/julian-date-converter/, Julian date 0.0 is:

```
Julian Date:
0.000000001
UTC Date:
-4713-11-24 12:00:00.000
```

PR #4645, opened 2024-08-19
[JSON] Add more information to unescaped character error message

nlw0:
The JSON parser usually offers little information to help finding errors in the input, I'm trying to make it a little more informative. One mistake I often do is to forget a closing '"' of a string, what causes a parsing error once we reach a newline character. This patch modifies the error message, that ultimately goes inside a thrown exception, to include the line number of the input and print the contents of the string up to the point where the invalid character was found.

I imagine there might be better approaches to this. I wrote this patch just to have something that minimally works for me in my work, and I'm looking forward to better ideas.

### DIFF
--- a/JSON/src/pdjson.c
+++ b/JSON/src/pdjson.c
@@ -505,7 +505,15 @@ read_string(json_stream *json)
                 return JSON_ERROR;
         } else {
             if (char_needs_escaping(c)) {
-                json_error(json, "%s", "unescaped control character in string");
+                int length = 0;
+                const char* val = json_get_string(json, &length);
+
+                char msgbuf[1024];
+                int msglen = (length < 1023) ? length : 1023;
+                memcpy(msgbuf, val, msglen);
+                msgbuf[msglen] = '\0';
+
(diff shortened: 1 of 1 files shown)


aleks-f:
this should go [upstream](https://github.com/skeeto/pdjson), not here

nlw0:
Fair point, thanks. On the other hand, the patch would be here if a better approach might be to add information to the `JSONException` object. That's the feedback I was looking for. Or maybe it's just a bad idea to begin with.

aleks-f:
> Fair point, thanks. On the other hand, the patch would be here if a better approach might be to add information to the `JSONException` object. That's the feedback I was looking for. Or maybe it's just a bad idea to begin with.

`JSONException` string will contain whatever is in the `json->errmsg` when `JSON_ERROR` is returned; the patch still has to go upstream. We plan to merge the upstream code for the release 1.14. this month, so if you want it released, see that it is accepted upstream. Also, make sure that address sanitizer does not complain about the buffer.

PR #4647, opened 2024-08-20
Use Int64 for TcpServerDispatcher::totalConnections()

svart-riddare:
Value was previously a 32-bit integer, overflowing after 2 billion connections. Seems a lot, but a 1000 connections/second, it's less than a month.

I did not change `TchServerDispatcher::refusedConnections()`, I do not expect this value to overflow (it if does, well the server is not very useful !).


banka50:
Maybe unsigned would be better? @matejk

svart-riddare:
Perhaps; when I make a change to a project which isnt' mine, I tend to make the smallest possible change.

PR #4652, opened 2024-08-22
enh(ScopedLockWithUnlock): make it more alike std::unique_lock

siren186:
```cpp
void test_std()
{
    std::mutex m;
    std::unique_lock<std::mutex> guard(m);
    guard.unlock(); // OK
    guard.lock(); // OK
}

void test_poco()
{
    Poco::Mutex m;
    Poco::ScopedLockWithUnlock<Poco::Mutex> guard(m);
    guard.unlock(); // OK
    guard.lock(); // Failed !!!
}
```

This PR is to make `Poco::ScopedLockWithUnlock` can lock/unlock for many time. Just same like `std::unique_lock` do.

### DIFF
--- a/Foundation/include/Poco/ScopedLock.h
+++ b/Foundation/include/Poco/ScopedLock.h
@@ -75,14 +75,16 @@ class ScopedLockWithUnlock
 	/// unlocking of the mutex.
 {
 public:
-	explicit ScopedLockWithUnlock(M& mutex): _pMutex(&mutex), _locked(true)
+	explicit ScopedLockWithUnlock(M& mutex): _pMutex(&mutex)
 	{
 		_pMutex->lock();
+		_locked = true;
 	}
 
-	ScopedLockWithUnlock(M& mutex, long milliseconds): _pMutex(&mutex), _locked(true)
+	ScopedLockWithUnlock(M& mutex, long milliseconds): _pMutex(&mutex)
 	{
 		_pMutex->lock(milliseconds);
+		_locked = true;
 	}
 
 	~ScopedLockWithUnlock()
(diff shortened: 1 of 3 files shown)


Foundation/include/Poco/ScopedLock.h
```diff
@@ -75,12 +75,12 @@ class ScopedLockWithUnlock
 	/// unlocking of the mutex.
 {
 public:
-	explicit ScopedLockWithUnlock(M& mutex): _pMutex(&mutex)
+	explicit ScopedLockWithUnlock(M& mutex): _pMutex(&mutex), _locked(true)
 	{
 		_pMutex->lock();
```
matejk, line 80: Technically, the mutex is not locked until _pMutec->lock() completes. I propose setting _locked to true after _pMutex->lock() succeeds.

Foundation/include/Poco/ScopedLock.h
```diff
@@ -97,17 +97,25 @@ class ScopedLockWithUnlock
...
+	{
+		poco_assert(_locked == false);
+		_pMutex->lock();
+		_locked = true;
+	}
+
 	void unlock()
 	{
-		if (_pMutex)
+		if (_locked)
```
matejk, line 109: Check for null _pMutex is removed. Would it make sense for both checks to stay? And do the check for nullptr in the ctor also?

Foundation/include/Poco/ScopedLock.h
```diff
@@ -97,17 +97,25 @@ class ScopedLockWithUnlock
...
 		{
 			_pMutex->unlock();
-			_pMutex = 0;
+			_locked = false;
 		}
 	}
 
 private:
 	M* _pMutex;
+	bool _locked;
```
matejk, line 118: Set default value to false.

PR #4656, opened 2024-08-28
Add possibility to load providers and use them in Poco

mapogu:
We have need to externally configure and load providers to OpenSSL 3.x and I noticed that current implementation of the Poco::Net::Context does not provide a way to do this. In my understanding the library context needs to be configured for OpenSSL to search for Providers at an appropriate location. In addition to this, the actual external providers need to be loaded and kept alive until shutdown. These two steps have to happen prior to creation of the SSL Context in order to use providers.

Therefore, I've created this patch as a suggestion to add this improvement to Poco.

### DIFF
--- a/NetSSL_OpenSSL/include/Poco/Net/Context.h
+++ b/NetSSL_OpenSSL/include/Poco/Net/Context.h
@@ -28,6 +28,7 @@
 #include "Poco/SharedPtr.h"
 #include "Poco/AutoPtr.h"
 #include <openssl/ssl.h>
+#include <openssl/types.h>
 #include <cstdlib>
 
 
@@ -140,6 +141,9 @@ public:
 	{
 		Params();
 			/// Initializes the struct with default values.
+
+		std::string providerName;
+		OSSL_LIB_CTX *libctx = nullptr;
 
 		std::string privateKeyFile;
 			/// Path to the private key file used for encryption.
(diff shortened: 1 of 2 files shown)


mapogu:
OK so two failed checks. Is it correct that CIFuzz and CodeQL build with OpenSSL version < 3.x ? In that case the correct way would be to add #if OPENSSL_VERSION_NUMBER >= 0x10100000L around changes, as the OpenSSL Provider API (replacement to Engine) is to my understanding a OpenSSL 3.x addition to OpenSSL ?

aleks-f:
@mapogu is it possible to add a unit test here? I'm not sure what provider(s) are reasonable to expect on different platforms

mapogu:
Sorry for the late response. I can try to add unit test. To my understanding the Provider API in OpenSSL should be available on all platforms from OpenSSL 3.x. It can be used to load for instance:

- Default Provider :  standard built-in OpenSSL algorithm
- Legacy Provider:  collection of legacy algorithms (md2, md4, ... BF, DES)
- FIPS Provider: algorithms conforming to FIPS standards
- Base Provider: small sub-set of non-cryptographic algorithms available in the default provider.
- Null Provider: "built-in" to libcrypto and contains no algorithm implementations. Guarantees that the default provider is not loaded.

A note in the docs seems to be that if a provider is loaded then the default provider is not automatically loaded by OpenSSL (has to be explicitly).

Hence I could try to write a test where we load for instance FIPS provider through Poco and check that the default provider is not available. 

I guess this addition could be useful when you want to run the combination Poco, OpenSSL 3.x and only FIPS approved algorithms.

mapogu:
Since this is explicitly for OpenSSL 3.x is there any good compiler flag for the unit test framework so that the test case doesn't run on OpenSSL 1.x ?

aleks-f:
> Since this is explicitly for OpenSSL 3.x is there any good compiler flag for the unit test framework so that the test case doesn't run on OpenSSL 1.x ?

@mapogu see [openssl3 CI jobs](https://github.com/pocoproject/poco/blob/64c751f3a31b978fc656f5e01b821d3d08a49cce/.github/workflows/ci.yml#L327)

PR #4663, opened 2024-09-04
Fixed incorrect SSL_CTX_set0_tmp_dh_pkey() usage

pkl97:
This simple program crashes POCO (tested under Red Hat Enterprise Linux 9.4):

```c++
#include <Poco/Net/Context.h>

int main()
{
    const Poco::Net::Context context(Poco::Net::Context::CLIENT_USE, "/tmp", Poco::Net::Context::VERIFY_STRICT, 9, false, "ALL");
    return 0;
}
```

The problem is an incorrect usage of SSL_CTX_set0_tmp_dh_pkey() in Context::initDH(). The return value is not evaluated and the key is freed even if it has been successfully transferred to the SSL Context.

The relevant part of the OpenSSL manpage https://docs.openssl.org/3.1/man3/SSL_CTX_set_tmp_dh_callback/:

> Ownership of the dhpkey value is passed to the SSL_CTX or SSL object as a result of this call, and so the caller should not free it if the function call is successful.

### DIFF
--- a/NetSSL_OpenSSL/include/Poco/Net/Context.h
+++ b/NetSSL_OpenSSL/include/Poco/Net/Context.h
@@ -136,9 +136,24 @@ public:
 		SECURITY_LEVEL_256_BITS = 5
 	};
 
+	enum KeyDHGroup
+	{
+		// MODP
+		//KEY_DH_GROUP_768  = 1,  // (768-bit)
+		KEY_DH_GROUP_1024 = 2,  // (1024-bit)
+		//KEY_DH_GROUP_1536 = 5,  // (1536-bit)
+		KEY_DH_GROUP_2048 = 14, // (2048-bit)
+		//KEY_DH_GROUP_3072 = 15, // (3072-bit)
+
+		// ECP
+		//KEY_DH_GROUP_256 = 19, // (256-bit random)
+		//KEY_DH_GROUP_384 = 20, // (384-bit random)
+		//KEY_DH_GROUP_521 = 21  // (521-bit random)
+	};
+
 	struct NetSSL_API Params
 	{
-		Params();
+		Params(KeyDHGroup dhBits = KEY_DH_GROUP_2048);
 			/// Initializes the struct with default values.
 
 		std::string privateKeyFile;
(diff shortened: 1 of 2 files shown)


aleks-f (changes requested):
well, you broke quite a bit of things with this, please see what else needs to be adressed; won't be merged unless CI is all green

chrisse74:
Hi all,

The error messages show that the newly introduced exception which is raised if SSL_CTX_set0_tmp_dh_pkey() returns with an error is triggering the failures in the CI.

> 1: CppUnit::TestCaller<HTTPSClientSessionTest>.testGetSmall
>     "Poco::Net::SSLContextException:
> SSL context exception: Context::initDH():SSL_CTX_set0_tmp_dh_pkey()
> 
> error:0A00018A:SSL routines::dh key too small"

OpenSSL rejects the given DH key because it is too small. Prior to this PR this rejection was not reported, leaving the client under the impression that the given DH keys were accepted.

My guess is that the machines running the failing tests are configured to use SECLEVEL=2 (see https://stackoverflow.com/questions/61626206/what-could-cause-dh-key-too-small-error ) and thereby do not support 1024-bit DH keys.

These errors go away if you change dhUse2048Bits from false to true in
https://github.com/pocoproject/poco/blob/b6b1fec14f3966ca2cf2b862f607e4c00093c964/NetSSL_OpenSSL/src/Context.cpp#L48C2-L48C23 

@aleks-f  Could changing dhUse2048Bits in Context::Params::Params() be the way forward?

PS: On my machine this change brings the errors down to two errors (probably unrelated?)
```
There were 2 errors:
 1: CppUnit::TestCaller<HTTPSClientSessionTest>.testProxy
    "Poco::Net::HTTPException:
HTTP Exception: Cannot establish proxy connection: Forbidden"
    in "<unknown>", line -1
 2: CppUnit::TestCaller<HTTPSStreamFactoryTest>.testProxy
    "Poco::Net::HTTPException:
HTTP Exception: Cannot establish proxy connection: Forbidden"
    in "<unknown>", line -1
```

aleks-f:
> @aleks-f Could changing dhUse2048Bits in Context::Params::Params() be the way forward?

Yes, but let's introduce the default values as parameters to the `Params::Params()`. And the key size should really be [the group number,](https://www.watchguard.com/help/docs/help-center/en-US/Content/en-US/Fireware/bovpn/manual/diffie_hellman_c.html) rather than a simple boolean switch between 1024/2048
 
> PS: On my machine this change brings the errors down to two errors (probably unrelated?)

yes, unrelated

aleks-f:
moved to #4753

Issue #4557, opened 2024-05-14
NestedDiagnosticContext should be thread local


siren186:
**Describe the bug**
[A clear and concise description of what the bug is.](https://github.com/pocoproject/poco/commit/d5a5ebc2d7af2cba520c29cc29075b3982eb099c#r141975904)

```cpp
NestedDiagnosticContext& NestedDiagnosticContext::current()
{
	static NestedDiagnosticContext ndc;
	return ndc;
}
```

It should use `Thread Local Storage` to store the NDC context. Not a singleton.

**Please add relevant environment information:**
 - Windows 10
 - poco-1.13.3

siren186:
```cpp
NestedDiagnosticContext& NestedDiagnosticContext::current()
{
	static NestedDiagnosticContext ndc;
	return ndc;
}
```

NestedDiagnosticContext are used for Multi-Thread, It may stored in threadlocal variables, But now it's a singleton without lock...
And it will crash in multi-thread

PR #4674, opened 2024-09-11
4557 ndc thread local

aleks-f:
replacement for #4667 , fixing #4557

PR #4682, opened 2024-09-12
4557 ndc thread local

siren186:
Replacement for PR: https://github.com/pocoproject/poco/pull/4674
1. Fix https://github.com/pocoproject/poco/issues/4557
2. Merged `dump(bool nameOnly)` function from [implemented here](https://github.com/pocoproject/poco/blob/df23923b01437031a96ddf1a69387a5f03feb2e3/Foundation/include/Poco/NestedDiagnosticContext.h#L120-L152)
3. Discard `backtrace` from [implemented here](https://github.com/pocoproject/poco/blob/df23923b01437031a96ddf1a69387a5f03feb2e3/Foundation/include/Poco/NestedDiagnosticContext.h#L120-L152) . Because it can be replaced with a more powerful implementation by [Boost.Stacktrace](https://www.boost.org/doc/libs/1_65_1/doc/html/stacktrace.html)

### DIFF
--- a/Foundation/include/Poco/NestedDiagnosticContext.h
+++ b/Foundation/include/Poco/NestedDiagnosticContext.h
@@ -89,9 +89,11 @@ public:
 		/// to the given stream. The entries are delimited by
 		/// a newline.
 
-	void dump(std::ostream& ostr, const std::string& delimiter) const;
+	void dump(std::ostream& ostr, const std::string& delimiter, bool nameOnly = false) const;
 		/// Dumps the stack (including line number and filenames)
 		/// to the given stream.
+		/// If nameOnly is false (default), the whole path to file is printed,
+		/// otherwise only the file name.
 
 	void clear();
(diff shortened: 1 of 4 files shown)


aleks-f:
> 3. Discard `backtrace` from [implemented here](https://github.com/pocoproject/poco/blob/df23923b01437031a96ddf1a69387a5f03feb2e3/Foundation/include/Poco/NestedDiagnosticContext.h#L120-L152) . Because it can be replaced with a more powerful implementation by [Boost.Stacktrace](https://www.boost.org/doc/libs/1_65_1/doc/html/stacktrace.html)

Only if you plan to rewrite it in poco style; we will not introduce boost dependency

siren186:
Rewrite `boost::stacktrace` in poco style is complex. I think current PR is not fit to solve the problem.
If there is a fate in the future, I will implement it and create another PR to solve the problem.

PR #4693, opened 2024-09-15
Serializable Isolation level for SQLite Databases 

aleks-f:

PR #4718, opened 2024-09-30
we should wait result of select before rollback

bas524:
In previous test-case code I was waiting stmt  after rallback, this is a root couse of bug...
but we saw error only on macos


bas524:
@aleks-f 
Hi!
Could you merge this PR and re-run #4693 ?

PR #4719, opened 2024-10-02
test: move fuzzing tests from google/oss-fuzz repository

tyler92:
Fuzz targets moved from [google/oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/poco) repository. SA https://github.com/google/oss-fuzz/pull/12506 for more details about advantages.

How does it work?

There is a new cmake option - ENABLE_FUZZING (for clang only). It enables fuzz targets (binaries with `LLVMFuzzerTestOneInput` as entrypoint). This option is used by `build/script/oss-fuzz-build.sh` script which will be called from `oss-fuzz` build process (I will prepare an MR for that). Fuzz targets with seed corpus and dictionaries are copied to `OUT` directory and oss-fuzz runs them. Please have a look [the documentation](https://google.github.io/oss-fuzz/) for details.

Why do we need `seed` in this repository?

It increases the efficiency of fuzzing significantly if the fuzz target is quite complex. Seed can contain any data (valid or malformed), the only goal is to provide fuzzers a hint of how to reach some code branches.

Can we put all seed files into a single tar archive?

Yes, we can do anything we want with the seed, there is only one requirement - `oss-fuzz-build.sh` has to prepare seed corpus with the name `$fuzzer_seed_corpus.zip` in `$OUT` directory where `$fuzzer` is the name of fuzz binary.

Is it possible to test it locally?

Sure. You can have a look at `build/script/oss-fuzz-build.sh` as a reference. Please note that this script is running on oss-fuzz environment where env like `CC`, `CXX`, etc have special values. But in general, you can do the following:

```sh
mkdir build && cd build

export LIB_FUZZING_ENGINE="-fsanitize=address,fuzzer --coverage"
export CXX_FLAGS="-fsanitize=address,fuzzer-no-link --coverage"
export CXX=clang++
export CC=clang

cmake -DENABLE_FUZZING=ON \
      -DBUILD_SHARED_LIBS=OFF \
      -DENABLE_TESTS=OFF \
      -DCMAKE_CXX_FLAGS=$CXX_FLAGS \
      -DCMAKE_C_FLAGS=$CXX_FLAGS \
      ..
cmake --build .
```

and run fuzzers. See the doc for reference: https://llvm.org/docs/LibFuzzer.html


tyler92:
The CI already contains `CIFuzz` job (I didn't know it) which uses that fuzz targets. So when this MR is merged the `CIFuzz` job will use fuzz targets directly from the Poco repository

aleks-f (changes requested):
@tyler92 can you please document the exact versions/requirements

```
$ cmake --version
cmake version 3.30.3

$ clang++ --version
Apple clang version 16.0.0 (clang-1600.0.26.3)
Target: arm64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

$ uname -a
Darwin M2.local 23.6.0 Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:04 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6020 arm64
```
```
$ export LIB_FUZZING_ENGINE="-fsanitize=address,fuzzer --coverage"
$ export CXX_FLAGS="-fsanitize=address,fuzzer-no-link --coverage"
$ export CXX=clang++
$ export CC=clang
$ cmake -DENABLE_FUZZING=ON \
>       -DBUILD_SHARED_LIBS=OFF \
>       -DENABLE_TESTS=OFF \
>       -DCMAKE_CXX_FLAGS=$CXX_FLAGS \
>       -DCMAKE_C_FLAGS=$CXX_FLAGS \
>       ..
CMake Error: Unknown argument --coverage
CMake Error: Run 'cmake --help' for all supported options.
```

tyler92:
@aleks-f I've corrected the description - there should be quotes for `-DCMAKE_CXX_FLAGS="$CXX_FLAGS"`. Now it should be OK, I rechecked it. I'm not sure about the minimal version of CMake and Clang but this is not something new and for your versions everything should work.

aleks-f (approved):

PR #4723, opened 2024-10-03
A reader writer lock on the file region

bas524:
This PR contains implementation of RWLock on the file descriptor
It allows multiple concurrent process-readers or one exclusive process-writer on the file region

example

```cpp
// open file for reading and writing
FileStream fs(filepath, std::ios::in | std::ios::out | std::ios::binary);
// write some information
Poco::Int32 i32 = 0;
fs.seekp(0, std::ios::beg);
fs.write((const char *)&i32, sizeof(i32));
fs.flushToDisk();
//...
// other process-reader
FileStreamRWLock fLock(fs, 0, sizeof(Poco::Int32));
ScopedFStreamReadRWLock readLock(fLock);
fs.seekg(0, std::ios::beg);
fs.read((char *)&counter, sizeof(counter));

//...
// other process-writer
FileStreamRWLock fLock(fs, 0, sizeof(Poco::Int32));
ScopedFStreamWriteRWLock writeLock(fLock);
fs.seekp(0, std::ios::beg);
fs.write((char *)&counter, sizeof(counter));
fs.flushToDisk();
```

!Warning
On windows you should open file by WINAPI and then use handle with poco, because Poco::FileStream doesn't open or create file with ```FILE_SHARE_READ | FILE_SHARE_WRITE``` and ```GENERIC_READ | GENERIC_WRITE```

You can do it with this code
```cpp
HANDLE openFileWithRWAccess(const std::string& path)
{
	DWORD access = GENERIC_READ | GENERIC_WRITE;
	DWORD shareMode = FILE_SHARE_READ | FILE_SHARE_WRITE;

	HANDLE handle = CreateFileA(path.c_str(), access, shareMode, NULL, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);

	if (handle == INVALID_HANDLE_VALUE)
		Poco::File::handleLastError(path);

	return handle;
}
```

and then

```cpp
FileStream fs;
fs.openHandle(openFileForSahredRWAccess(fl), std::ios::in | std::ios::out | std::ios::binary);
```


Foundation/testsuite/src/FileStreamRWLockTest.cpp
```diff
@@ -0,0 +1,264 @@
...
+	Poco::Path testAppPath = Poco::Path::current();
+	testAppPath.append(name + ext).makeFile();
+	return testAppPath.toString();
+}
+
+#if POCO_OS == POCO_OS_WINDOWS_NT
+
+#include "Poco/File_WIN32U.h"
+
+HANDLE createFileForSahredRWAccess(const std::string &path)
```
aleks-f, line 57: spelling

Foundation/src/FileStreamRWLock.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+
+FileStreamRWLock::FileStreamRWLock(const FileStream &fs, Poco::UInt64 offset, Poco::UInt64 size) :
+	FileStreamRWLockImpl(fs.nativeHandle(), offset, size)
+{
+}
+
+
+FileStreamRWLock::~FileStreamRWLock()
+{
```
aleks-f, line 35: there should be an `atomic_bool _locked` flag that unlocks the file if it is still locked on destruction

aleks-f (changes requested):

Foundation/testsuite/src/FileStreamRWLockTest.cpp
bas524: fixed (I hope)

Foundation/src/FileStreamRWLock.cpp
bas524: done

aleks-f:
@bas524 wasn't triggering al CI, moved to #4734

PR #4725, opened 2024-10-04
Exception (VS): static assert to make sure that _HAS_EXCEPTIONS is set

### LINKED ISSUE #4668: Cross-module exception issue with pre-compiler define _HAS_EXCEPTIONS=0 used in an application
Hi @obiltschnig 

I encountered a very subtle issue. It is about `Poco::Exception` in `PocoFoundation64.dll`. In Visual Studio 2019/2022 **release_shared** x64.
It is **very hard to reproduce**. Due to the large amount of code, I can't paste it all here, so I’ve only extracted a small portion.
```cpp
// This is a small portion I extracted from a large amount of code.
int main() {
    std::uint64_t val = 0x1122334455667788;
    
    try {
    	throw Poco::Exception("error"); 
        // If I do not use Poco::XXXXXXException, everything works fine.
        // throw std::exception("error"); // It's OK !!!
    } catch (const Poco::Exception& e) {
    }
    
    std::cout << val; // Now the val is 0x1122334400000000 !!!
    // It's Very Strange !!! It lost low 32 bit.
    return 0;
}
```

It took me two days to find the real reason.
```cpp
auto m = sizeof(Poco::Exception);
auto n = sizeof(std::exception);

// m=72, n=24: In PocoFoundation64.dll
// m=64, n=16: In my EXE
```

It was all caused by **`_HAS_EXCEPTIONS=0`** in my EXE compile options. This macro will change the class size and implementation of std::exception.
```cpp
// By: _HAS_EXCEPTIONS=1 (default in visual studio)
// sizeof(std::exception) == 24
namespace std { 
    class exception {
    public:
        virtual ~exception() {}
    private:
        char const* _What;
        bool        _DoFree;
    }
}


// By: _HAS_EXCEPTIONS=0
// sizeof(std::exception) == 16
namespace std { 
    class exception {
    public:
        virtual ~exception() {}
    protected:
        const char* _Ptr;
    }
}
```

I think it's necessary to **add some tips in POCO documentation** to avoid users wasting unnecessary time on this issue.

----
See alse: https://github.com/nodejs/node-addon-api/issues/85

matejk:
Resolves #4668.

### DIFF
--- a/Foundation/include/Poco/Exception.h
+++ b/Foundation/include/Poco/Exception.h
@@ -100,9 +100,15 @@ protected:
 private:
 	std::string _msg;
 	Exception*  _pNested;
-	int			_code;
+	int			_code;    
 };
 
+#if defined(_HAS_EXCEPTIONS)
+    // Size of Poco::Exception depends on the exception settings (like _HAS_EXCEPTIONS)
+    // that might influence size of std::exception from which Poco::Exception is derived from.
+    // It is expected that Poco libraries and application using Poco have the same settings.
+    static_assert(_HAS_EXCEPTIONS != 0);
+#endif
 
 //
 // inlines


obiltschnig (commented):
Isn't `_HAS_EXCEPTIONS` MSVC specific?

obiltschnig (changes requested):
Also, I'm pretty sure the <typeinfo> header is needed on some platforms.

matejk:
> Isn't `_HAS_EXCEPTIONS` MSVC specific?

I realised that quick and corrected it :)

matejk:
> Also, I'm pretty sure the `<typeinfo>` header is needed on some platforms.

Also corrected.

PR #4726, opened 2024-10-04
Explicitly define conditional LOB constructors for const (w)string &

### LINKED ISSUE #4722: libcxx: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
**Describe the bug**
With LLVM/Clang/libcxx 19.1.0 on Gentoo Linux, POCO fails to compile with the following error:

```
In file included from /var/tmp/portage/dev-libs/poco-1.13.3/work/poco-poco-1.13.3-release/Data/src/AbstractPreparation.cpp:15:
In file included from /var/tmp/portage/dev-libs/poco-1.13.3/work/poco-poco-1.13.3-release/Data/include/Poco/Data/AbstractPreparation.h:21:
In file included from /var/tmp/portage/dev-libs/poco-1.13.3/work/poco-poco-1.13.3-release/Data/include/Poco/Data/Data.h:23:
In file included from /var/tmp/portage/dev-libs/poco-1.13.3/work/poco-poco-1.13.3-release/Foundation/include/Poco/Foundation.h:94:
/usr/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
  820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
/usr/include/c++/v1/__type_traits/is_constructible.h:42:62: note: in instantiation of template class 'std::basic_string<unsigned char>' requested here
   42 |     : public integral_constant<bool, __is_constructible(_Tp, __add_rvalue_reference_t<_Tp>)> {};
      |                                                              ^
/usr/include/c++/v1/__type_traits/is_swappable.h:43:39: note: in instantiation of template class 'std::is_move_constructible<Poco::Data::LOB<unsigned char>>' requested here
   43 | using __swap_result_t = __enable_if_t<is_move_constructible<_Tp>::value && is_move_assignable<_Tp>::value>;
      |                                       ^
/usr/include/c++/v1/__type_traits/is_swappable.h:50:60: note: in instantiation of template type alias '__swap_result_t' requested here
   50 | inline _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX20 __swap_result_t<_Tp> swap(_Tp& __x, _Tp& __y)
      |                                                            ^
/var/tmp/portage/dev-libs/poco-1.13.3/work/poco-poco-1.13.3-release/Data/include/Poco/Data/LOB.h:243:14: note: while substituting explicitly-specified template arguments into function template 'swap' 
  243 |         inline void swap<Poco::Data::BLOB>(Poco::Data::BLOB& b1, Poco::Data::BLOB& b2) noexcept
      |                     ^
/usr/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
      |                             ^
```

This error occurs because of [this line](https://github.com/pocoproject/poco/blob/af3b3b19022d279bee648fd48411212827fa875b/Data/include/Poco/Data/LOB.h#L66) which instantiates std::basic_string\<unsigned char\> for BLOB which is not defined in libcxx.

I have a patch [here](https://940705.bugs.gentoo.org/attachment.cgi?id=904862) that disables that constructor in case basic_string\<T\> is not instantiable and fixes compilation.

However I'm not sure if it isn't my system that's hosed in some way since a complete compilation failure would have surely been caught already since libcxx is used for Mac builds, right? 

**To Reproduce**
Compile POCO with clang 19 linking against libcxx.

**Expected behavior**
POCO compiles

**Logs**
https://940705.bugs.gentoo.org/attachment.cgi?id=904816

 - Gentoo Linux (repository dated Tue, 01 Oct 2024 16:00:00 +0000)
 - POCO Version 1.13.3

**Additional context**
Gentoo [bug 940705](https://bugs.gentoo.org/940705)

2xsaiko:
This fixes compilation against libcxx 19, which removed the
non-standard specializations for std::char_traits including the one
for unsigned char, which would get instantiated for BLOB.
See https://reviews.llvm.org/D138307 and
https://reviews.llvm.org/D157058.

Closes #4722.

### DIFF
--- a/Data/include/Poco/Data/LOB.h
+++ b/Data/include/Poco/Data/LOB.h
@@ -63,7 +63,15 @@ public:
 	{
 	}
 
-	LOB(const std::basic_string<T>& content):
+	template<typename U = T>
+	LOB(const std::enable_if_t<std::is_same_v<T, U> && std::is_same_v<U, char>, std::string>& content):
+		_pContent(new std::vector<T>(content.begin(), content.end()))
+		/// Creates a LOB from a string.
+	{
+	}
+
+	template<typename U = T>
+	LOB(const std::enable_if_t<std::is_same_v<T, U> && std::is_same_v<U, wchar_t>, std::wstring>& content):
 		_pContent(new std::vector<T>(content.begin(), content.end()))
 		/// Creates a LOB from a string.
 	{


2xsaiko:
Looks like the initial approach disabled implicit conversions of the parameter when calling the constructor (i.e. `CLOB("foo")`). So here's the approach you probably had in mind in https://github.com/pocoproject/poco/issues/4722#issuecomment-2393345744.

Issue #4324, opened 2023-12-06
Poco:Data::ODBC - MSSQL (n)varchar(max) length issue


Spaky:
**Describe the bug**
The data binding of (n)varchar(max) field in Poco:Data::ODBC seems just to allow max 8000 bytes, but MSSQL Server allow allows up to 2^31-1 bytes. 

**To Reproduce**
Create a table with a varchar(max) or nvarchar(max) field, insert data with more then 8000 byte.

**Expected behavior**
I should be possible to insert data with more then 8000 bytes when SQL datatype is (n)varchar(max).

**Logs**
INSERT failed: exception: 'class Poco::Data::LengthExceededException' what: 'Data too long' message: 'Error binding column 6 size=67150, max size=8000)' 

**Please add relevant environment information:**
 - Windows 10, Windows Server 2019
 - POCO Version: 1.12.4
 - MS SQLServer 2019

**Additional context**
I think the retrieving of the type info of the DB needs a special handling for (n)varchar(max).

aleks-f:
From [SQLBindParameter](https://learn.microsoft.com/en-us/sql/relational-databases/native-client-odbc-api/sqlbindparameter?view=sql-server-ver16#binding-parameters-for-sql-character-types):

> If the SQL data type passed in is a character type, ColumnSize is the size in characters (not bytes). If the length of the data string in bytes is greater than 8000, ColumnSize should be set to SQL_SS_LENGTH_UNLIMITED, indicating that there is no limit to the size of the SQL type.

> For instance, if the SQL data type is SQL_WVARCHAR, ColumnSize should not be greater than 4000. If the actual data length is greater than 4000, then ColumnSize should be set to SQL_SS_LENGTH_UNLIMITED so that nvarchar(max) will be used by driver.

The `SQL_SS_LENGTH_UNLIMITED` is not standard ODBC. If you want this feature, add the SQLServer-specific session feature like eg. `multiActiveResultset`:
https://github.com/pocoproject/poco/blob/e868ecb2472511cfe25e0f71d5839cb766cb45c3/Data/ODBC/src/SessionImpl.cpp#L134-L138
and send pull request.

aleks-f:
See also [DataFormat](https://learn.microsoft.com/en-us/sql/relational-databases/native-client/odbc/large-clr-user-defined-types-odbc?view=sql-server-ver16):

[...]

PR #4738, opened 2024-10-15
fix(ODBC): Poco:Data::ODBC - MSSQL (n)varchar(max) length issue #4324

aleks-f:


aleks-f:
@matejk please add MS SQLServer libs in the CI environment, so this takes effect there:
https://github.com/pocoproject/poco/blob/6e09ec2fbd89fc0fa92b8853da29b2a79923e258/Data/ODBC/include/Poco/Data/ODBC/ODBC.h#L58

I tested locally on Debian12 and this should be sufficient to find the header:
https://github.com/pocoproject/poco/blob/6e09ec2fbd89fc0fa92b8853da29b2a79923e258/Data/ODBC/ODBC.make#L9

Data/ODBC/include/Poco/Data/ODBC/ODBC.h
```diff
@@ -58,6 +58,20 @@
...
+		//
+		// - this setting works in conjunction with
+		//   the session "maxFieldSize" property, which
+		//   ultimately determines the max string length.
+		//
+		// - increasing the "maxFieldSize" property may
+		//   affect performance (more memory preallocated
+		//   for prepared statements in order to safely
+		//   accomodate data returned at execution)
+		#define POCO_DATA_SQL_SERVER_BIG_STRINGS 1
```
matejk, line 74: Shall this be configurable in CMake?

Data/ODBC/ODBC.make
```diff
@@ -12,6 +12,16 @@ POCO_ODBC_INCLUDE = /opt/homebrew/include
...
+ifndef POCO_MSSQL_INCLUDE
+ifeq (0, $(shell test -e /usr/include/msodbcsql.h; echo $$?))
+POCO_MSSQL_INCLUDE = /usr/include
+else ifeq (0, $(shell test -e /opt/microsoft/msodbcsql18/include/msodbcsql.h; echo $$?))
+POCO_MSSQL_INCLUDE = /opt/microsoft/msodbcsql18/include
+else ifeq (0, $(shell test -e /opt/homebrew/include/msodbcsql18/msodbcsql.h; echo $$?))
+POCO_MSSQL_INCLUDE = /opt/homebrew/include/msodbcsql18
+endif
+endif
+
```
matejk, line 24: Shall these settings be added to CMake also?

matejk (commented):
Code changes seem OK to me. I just have comments whether it shall be possible to configure MSSQL specialities in CMake also.

Data/ODBC/include/Poco/Data/ODBC/ODBC.h
aleks-f: It could be, but then it should be wrapped in `#ifndef POCO_DATA_SQL_SERVER_BIG_STRINGS` here

Data/ODBC/ODBC.make
aleks-f: Yes, in order for SQL Server big strings to work, at compile time the path to `mssqlodbc.h` should be in compiler include directories list. On some systems I've seen that header in `/usr/include`, but mostly it's not there it seems. We try our best to find it automatically, but we also provide the way to explicitly specify it:

https://github.com/pocoproject/poco/blob/a97fd9eaa5c028350c6044c8c648ccc19eac117d/configure#L212-L213

https://github.com/pocoproject/poco/blob/a97fd9eaa5c028350c6044c8c648ccc19eac117d/configure#L384-L385

At this time, we only need [SQL_SS_LENGTH_UNLIMITED](https://github.com/pocoproject/poco/blob/a97fd9eaa5c028350c6044c8c648ccc19eac117d/Data/ODBC/src/Binder.cpp#L170), but even that is a symbol not guaranteed to have the same numeric value forever, and there may be more such things in the future (there are always some nonstandard specifics that leak through the abstraction)

Issue #4395, opened 2024-01-15
There is no way to resolve host in advance and connect to HTTPS server with SNI.


alexey-milovidov:
This is a continuation of #3456, which was mistakenly closed by the offensive "stale" bot.
This is a continuation of #3123, which was mistakenly closed by the offensive "stale" bot.

It does not matter if four years have passed. The issue is still relevant.

micheleselea:
I think it's possible just using this

Poco::Net::SocketAddress sa(IP-ADDRESS, PORT);
Poco::Net::SecureStreamSocket ss(sa, "DNS FOR SNI");

aleks-f:
>  mistakenly closed by the offensive "stale" bot.

It is not a mistake and the bot is there not to offend anyone, but rather to (1) remind OP what's going on with their issue, and (2) to help us remain sane by keeping the number of opened issues reasonable.

> It does not matter if four years have passed. The issue is still relevant.

It does not matter how many years passes - if nobody cares enough to send a fix, the issue will eventually be automatically staled and closed. And I mean a proper fix, not a workaround hack.

Your turn - [send the fix](https://github.com/pocoproject/poco/blob/poco-1.13.1/CONTRIBUTING.md#code-contributions) and we'll include it in a future release.

aleks-f:
> I think it's possible just using this
> 
> Poco::Net::SocketAddress sa(IP-ADDRESS, PORT); Poco::Net::SecureStreamSocket ss(sa, "DNS FOR SNI");

That's not what the issue is about; the problem happens on [reconnect](https://github.com/pocoproject/poco/issues/3573#issuecomment-1525852750)

PR #4751, opened 2024-10-30
fix(HTTPSClientSession): There is no way to resolve host in advance a…

aleks-f:
…nd connect to HTTPS server with SNI. #4395


matejk (approved):
Code changes seem OK to me.

PR #4753, opened 2024-10-31
Openssl DH key size

aleks-f:
replacement for #4663 

This simple program crashes POCO (tested under Red Hat Enterprise Linux 9.4):

```c++
#include <Poco/Net/Context.h>

int main()
{
    const Poco::Net::Context context(Poco::Net::Context::CLIENT_USE, "/tmp", Poco::Net::Context::VERIFY_STRICT, 9, false, "ALL");
    return 0;
}
```

The problem is an incorrect usage of SSL_CTX_set0_tmp_dh_pkey() in Context::initDH(). The return value is not evaluated and the key is freed even if it has been successfully transferred to the SSL Context.

The relevant part of the OpenSSL manpage https://docs.openssl.org/3.1/man3/SSL_CTX_set_tmp_dh_callback/:

> Ownership of the dhpkey value is passed to the SSL_CTX or SSL object as a result of this call, and so the caller should not free it if the function call is successful.

@pkl97


matejk (approved):

NetSSL_OpenSSL/src/Context.cpp
```diff
@@ -921,18 +957,22 @@ void Context::initDH(bool use2048Bits, const std::string& dhParamsFile)
...
 			dh->length = 256;
 		}
-		else
+		else if (keyDHGroup == KEY_DH_GROUP_1024)
 		{
 			dh->p = BN_bin2bn(dh1024_p, sizeof(dh1024_p), 0);
 			dh->g = BN_bin2bn(dh1024_g, sizeof(dh1024_g), 0);
 			dh->length = 160;
 		}
+		{
```
SinghRajenM, line 972: This must be in else: https://github.com/pocoproject/poco/commit/c4f66d518891736d08636fa0fa7758303587899c#r152747523

Issue link: https://github.com/pocoproject/poco/issues/4877

PR #4759, opened 2024-11-06
DB logger sample

### LINKED ISSUE #4750: DBLogger sample
- [ ] create a Util::ServerApplication-based DBLogger sample in Poco::Data which:
  - connects to a database
  - logging thread produces some logs via `SQLChannel` in `store-and-forward` mode
  - forwarding thread:
    - monitors the directory where files with *.sql extension (containing SQL inserts statements) are saved by `SQLChannel` in store-and-forward mode
    - executes the queries in the *.sql files
      - in case of success, deletes the *.sql file
      - in case of error, renames the offending file from *.sql to *.err
    - any files with extension different from *.sql are ignored
  - use sqlite db for the sample (like other `Poco::Data` samples)

matejk:
Resolves #4750


Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	void onItemRemoved(const Poco::DirectoryWatcher::DirectoryEvent& ev)
+	{
+		logger().trace("Removed: %s", ev.item.path());
+	}
+
+	void processFile(const Poco::File& file)
+	{
+		// TODO: What to do with symbolic links?
```
aleks-f, line 222: we don't want to bother with links to files; I think directory link should not be a problem

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+				auto& s = (*_dataSession);
+
+				s << sql, now;
+
+				s.isGood();
+
+				f.remove();
+				++_processed;
+			}
+			// TODO: What if sql is empty?
```
aleks-f, line 266: log error and rename file to `.err`

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+	{
+		logger().trace("Moved to: %s", ev.item.path());
+	}
+
+	void onItemRemoved(const Poco::DirectoryWatcher::DirectoryEvent& ev)
+	{
+		logger().trace("Removed: %s", ev.item.path());
+	}
+
+	void processFile(const Poco::File& file)
```
aleks-f, line 220: The entire processing sequence should be broken down into (a) detection, (b) forwarding to DB and (c) deleting the file or marking it as error. This should be done in threads (or `Poco::Task`s), with one scanner doing (a) and multiple (number configurable) workers doing (b),(c). 

So, let's leave just the Application functionality  in the main thread, then have (1) one (optional) sample message creation thread, (2) one thread just for monitoring the directory and (3) multiple worker threads for processing files. 

The monitoring thread just detects new files, and puts the file names into a shared `NotificationQueue`. Working threads are removing paths from the queue and doing the actual work. Number of working threads should be configurable so app can scale. Working threads should be capable to graciously deal with the condition when file does not exist (eg. file deleted while notification was waiting in the queue).

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+using namespace Poco::Data::Keywords;
+using Poco::Util::Application;
+using Poco::Util::Option;
+using Poco::Util::OptionSet;
+using Poco::Util::HelpFormatter;
+using Poco::Util::OptionCallback;
+
+using namespace std::string_literals;
+
+class DBLogger: public Application
```
aleks-f, line 38: Should be `ServerApplication`

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+				.callback(OptionCallback<DBLogger>(this, &DBLogger::handleHelp)));
+
+		options.addOption(
+			Option("?", "?", "display help information on command line arguments")
+				.required(false)
+				.repeatable(false)
+				.callback(OptionCallback<DBLogger>(this, &DBLogger::handleHelp)));
+
+		options.addOption(
+			Option("dir", "d", "directory path to scan for SQL log files")
```
aleks-f, line 134: these options should also be configurable from `DBLogger.properties` file

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,337 @@
...
+			std::stringstream buffer;
+			buffer << is.rdbuf();
+			const auto& sql { buffer.str() };
+			if (!sql.empty())
+			{
+				auto& s = (*_dataSession);
+
+				s << sql, now;
+
+				s.isGood();
```
aleks-f, line 261: no need for this. If sql succeeded, it's all fine. If it doesn't succeed, it will throw.

aleks-f (changes requested):

Data/samples/DBLogger/src/DBLogger.cpp
matejk: This was left in by mistake.

Data/samples/DBLogger/src/DBLogger.cpp
matejk: With exception of initial directory scan the changes in the directory are monitored with DirectoryWatcher. So far it seems that it works properly.

Will add configurable workers to populate the database.

matejk: Reimplemented to use DirectoryIterator and NotificationQueue, however something does not work right because processing gets stuck after few iterations.

I'll find a different way to process the files.

matejk: Another refactoring using STL.

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,476 @@
...
+			auto entry = popEntry();
+			processFile(entry);
+		}
+	}
+
+	int main(const std::vector<std::string>& args) override
+	{
+		if (!_helpRequested)
+		{
+			logger().information("Press any key to stop scanning.");
```
aleks-f, line 433: This is a daemon, can't do this. Use `waitForTerminatonRequest()`

aleks-f (changes requested):
Functionality ok, but it should be tidied up a bit, there's too much functionality crammed into a single class. Something along these lines:
https://github.com/pocoproject/poco/blob/c8dade87a68643b1a07245742354ac294008965b/Util/samples/SampleServer/src/SampleServer.cpp#L108-L112

matejk:
> Functionality ok, but it should be tidied up a bit, there's too much functionality crammed into a single class. Something along these lines:
> 
> https://github.com/pocoproject/poco/blob/c8dade87a68643b1a07245742354ac294008965b/Util/samples/SampleServer/src/SampleServer.cpp#L108-L112

Extracted scanning and inserting functionality to a separate class `SQLLogInserter`.

Data/samples/DBLogger/src/DBLogger.cpp
```diff
@@ -0,0 +1,268 @@
...
+		{
+			return;
+		}
+
+		loadConfiguration(); // load default configuration files, if present
+		Application::initialize(self);
+
+		Poco::Data::SQLite::Connector::registerConnector();
+
+		// TODO: Only delete and create table when creating demo messages?
```
aleks-f, line 61: yes

Data/samples/DBLogger/src/SQLLogInserter.h
```diff
@@ -0,0 +1,138 @@
...
+using namespace Poco::Data::Keywords;
+using namespace std::string_literals;
+
+class SQLLogInserter
+{
+public:
+
+	using WorkSet = std::unordered_set<std::string>;
+
+	SQLLogInserter()
```
aleks-f, line 31: move to .cpp

Data/samples/DBLogger/src/DBLogger.cpp
matejk: Implemented changes to provide options via configuration file and command line.

aleks-f (changes requested):
Need to have `DBLogger.properties` file, with all the commend option settings and logger settings, something like this:
```
logging.loggers.root.channel = appSplitter
logging.loggers.root.level = information

logging.channels.appFile.class = FileChannel
logging.channels.appFile.pattern = %L%Y-%m-%d %H:%M:%S.%i [%p] %s<%I>: %t
logging.channels.appFile.path = ${application.baseName}.log
logging.channels.appFile.times = local
logging.channels.appFile.rotation = daily
logging.channels.appFile.archive = number
logging.channels.appFile.purgeCount = 10

logging.channels.sql.class = SQLChannel
logging.channels.sql.name = DBLogger
logging.channels.sql.connector = SQLite
logging.channels.sql.connect = DBLogger.db3
logging.channels.sql.timeout = 3000
logging.channels.sql.minBatch = 5
logging.channels.sql.maxBatch = 1000
logging.channels.sql.bulk = true
logging.channels.sql.table = T_LOG
logging.channels.sql.async = false
logging.channels.sql.throw = false
logging.channels.sql.file = ${application.baseName}.sql

logging.channels.console.class = ColorConsoleChannel
logging.channels.console.pattern = %L%Y-%m-%d %H:%M:%S.%i [%p] %s<%I>: %t

logging.channels.appSplitter.class = SplitterChannel
logging.channels.appSplitter.channels = console, appFile, sql
```

aleks-f (changes requested):
- [x] Sample should work out of the box, but it doesn't:
```
$ bin/Linux/aarch64/DBLogger
Directory: 
Database: , 
Number of workers: 2
Not found
Created 0 messages, processed 0 messages in 0 ms.
```

- [x] There should be a reasonable initial default setup so that workers can keep up with sql file generation; as it is now, the filesystem gets clogged with sql files. 

- [x] logging configuration, as mentioned [here](https://github.com/pocoproject/poco/pull/4759#pullrequestreview-2428339345)
- [x] SQL statements in the files should insert more than a single row; should be achievable with `minBatch` property

matejk:
> 
>     * [ ]  SQL statements in the files should insert more than a single row; should be achievable with `minBatch` property

@aleks-f :

Is it a matter of SQL channel configuration to insert multiple SQL statements into the same file or shall DB logger have its own configuration to merge multiple sql files together?

aleks-f:
> > ```
> > * [ ]  SQL statements in the files should insert more than a single row; should be achievable with `minBatch` property
> > ```
> 
> @aleks-f :
> 
> Is it a matter of SQL channel configuration to insert multiple SQL statements into the same file or shall DB logger have its own configuration to merge multiple sql files together?

It should be controllable with `minBatch` property, when >1 the channel should concatenate multiple rows so they are inserted in one execution:

https://github.com/pocoproject/poco/blob/cb91880bfb63b7ecdff37efc90ce14a2c8c96bd0/Data/src/SQLChannel.cpp#L56-L58

The `maxSQL` property should be set not to exceed the max SQL statement length allowed by back end:

https://github.com/pocoproject/poco/blob/cb91880bfb63b7ecdff37efc90ce14a2c8c96bd0/Data/src/SQLChannel.cpp#L400-L408

See eg. `Batch size` for [SQL Server](https://learn.microsoft.com/en-us/sql/sql-server/maximum-capacity-specifications-for-sql-server?view=sql-server-ver16&redirectedfrom=MSDN)

matejk:
> > > ```
> > > * [ ]  SQL statements in the files should insert more than a single row; should be achievable with `minBatch` property
> > > ```
> > 
> > 
> > @aleks-f :
> > Is it a matter of SQL channel configuration to insert multiple SQL statements into the same file or shall DB logger have its own configuration to merge multiple sql files together?
> 
> Number of rows inserted in one statement should be controllable with `minBatch`, `maxBatch` and `maxSQL` property, when >1 the channel should concatenate multiple rows so they are inserted in one execution:
> 

Thanks, the latest code uses `SQLLogger` demo channel configuration from `DBLogger.properties` where all of its properties can be tuned.

aleks-f (approved):

Issue #4202, opened 2023-10-19
[Bug] if path does not exist, then `LogFileImpl::createFile()` fails to open the file


Smurf-IV:
V 1.12.4

I think the that the `LogFileImpl::createFile()` should check for the existence of the path and try to create it if it does not exist.

Smurf-IV:
I am not working on the project that uses the Poco source anymore, But Just search the code base for the `bit` I posted.

matejk:
@capak07, do you still plan to work on this?

matejk:
Fixed in main (commit 7b58bc10e) via PR #4959. Will be included in release 1.15.0.

PR #4783, opened 2024-11-25
LogFileImpl::createFile() error

capak07:
#4202 Checked if the path exists, otherwise create one


matejk:
Created branch [4873_file-create-path](https://github.com/pocoproject/poco/tree/refs/heads/4873_file-create-path) in poco repository with these changes to run builds and tests on github.

PR #4787, opened 2024-11-28
NetSSL_OpenSSL and NetSSL_Win: non-blocking support, shutdown behavior fix

obiltschnig:
This fixes a whole lot of issues in NetSSL_OpenSSL and NetSSL_Win.
Specifically:
  - we now have full support for non-blocking TLS sockets in both NetSSL_OpenSSL and NetSSL_Win
  - fixed TLS shutdown behavior, hopefully once and for all in both NetSSL_OpenSSL and NetSSL_Win
  - cleaned up and fixed NetSSL_Win, including workaround for Schannel bugs

### DIFF
--- a/Net/include/Poco/Net/SocketImpl.h
+++ b/Net/include/Poco/Net/SocketImpl.h
@@ -170,12 +170,22 @@ public:
 	virtual void shutdownReceive();
 		/// Shuts down the receiving part of the socket connection.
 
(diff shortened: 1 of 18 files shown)


matejk (approved):
Honestly I don't feel fully competent for review, because I don't know all of the background. Some parts are complete rewrites.

PR #4799, opened 2024-12-02
Attempting to load a passphrase-protected key should not prompt cin

### LINKED ISSUE #4627: Attempting to load a passphrase-protected key should not prompt cin
**Describe the bug**

Assuming a private key file protected by a password exists, it can be loaded with the `EVPPkey` constructor accepting empty pubkey name, private key name, and passphrase. If The passphrase is empty, the program will prompt the user to type in the key
with a message like `Enter PEM pass phrase:`. This is unsuitable for library code, since this cannot be reacted to unless the program is running in an interactive terminal.
It would be better to raise an exception, which happens e.g. when the wrong password is supplied.

**To Reproduce**

Attached an MVP with an encrypted openssl key

[poco.zip](https://github.com/user-attachments/files/16511333/poco.zip)

**Expected behavior**

Poco raises an exception when the key requires a passphrase, but none was supplied

**Logs**
n/a

**Screenshots**

CleanShot 2024-08-06 at 13 58 31

 - OS Type and Version: macos ARM 12.7.5
 - POCO Version: 1.13.3 (homebrew)

**Additional context**

I know this probably comes from OpenSSL, but can this be worked around somehow?

nyashbox:
Fixes #4627

### DIFF
--- a/Crypto/include/Poco/Crypto/EVPPKey.h
+++ b/Crypto/include/Poco/Crypto/EVPPKey.h
@@ -237,7 +237,7 @@ private:
 				if (pFile)
 				{
 					pem_password_cb* pCB = pass.empty() ? (pem_password_cb*)0 : &passCB;
-					void* pPassword = pass.empty() ? (void*)0 : (void*)pass.c_str();
+					void* pPassword = (void*)pass.c_str();
 					if (readFunc(pFile, &pKey, pCB, pPassword))
 					{
 						fclose(pFile); pFile = 0;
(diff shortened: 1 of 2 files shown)


matejk:
This PR seems to solve the problem with empty password on macOS (OpenSSL 3.4.0).

However, on Ubuntu 24.10 (OpenSSL 3.3.1), the example prompts for the password with or without the code change.

obiltschnig:
Let's move this to 1.15.

matejk:
Superseded by #5128, which provides a more complete fix that works on both macOS and Linux.

The issue with PR #4799 was that it only changed `pPassword` but kept `pCB` (callback) as NULL when password was empty. On Linux, when the callback is NULL, OpenSSL ignores `pPassword` entirely and still prompts stdin.

The fix in #5128 always provides the callback, which correctly handles empty passwords by returning 0 (no password), causing OpenSSL to throw a proper error.

Issue #3304, opened 2021-06-09
Windows Eventviewer not able to find PocoFoundation DLL


KevDi:
I want to log from my Application to the Windows Event Log.
When i run the Application as an Administrator it creates the Registry Keys under the `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\EventLog` Path. Without Admin Rights the key is not created.
I have placed the `PocoFoundation64.dll` inside the `System32` Folder but in the Eventlog it still gives me the message:

`the description for event id from source cannot be found`

I'm using Poco in Version 1.10.1

KevDi:
@obiltschnig  I further investigate on that problem. If i make a visual studio build directly from the provided Visual Studio Solution file inside the Poco Foundation Folder and use that created lib and dll File it works. 
But if i made a CMake Build or use the Conan Package Manager there seems to be two Problems:

1. CMake/Conan gives me a `PocoFoundationd.dll` instead of the `PocoFoundation64d.dll` which was created by the Visual Studio Solution File.
2. The `setUpRegistry()` Function in the `EventlogChannel.cpp` could not find the `dll` because of the Naming missmatch. As shown here: 
```
if (disp == REG_CREATED_NEW_KEY)
	{
		std::wstring path;
		#if defined(POCO_DLL)
			#if defined(_DEBUG)
				#if defined(_WIN64)
					path = findLibrary(L"PocoFoundation64d.dll");
				#else
					path = findLibrary(L"PocoFoundationd.dll");
				#endif
			#else
				#if defined(_WIN64)
					path = findLibrary(L"PocoFoundation64.dll");
				#else
					path = findLibrary(L"PocoFoundation.dll");
				#endif
			#endif
		#endif
```
Also manually adding the Registry Keys and pointing them to the `PocoFoundationd.dll` does not work.
This is also the same for the Release Build.

obiltschnig:
Thanks for looking into it. The solution is then probably to pass an additional preprocessor macro when building POCO, indicating that CMake is being used, so that the expected DLL name can be set accordingly.

KevDi:
where should it be placed? so maybe i can fix this and send you a merge request to fix this

obiltschnig:
Probably add a definition like:
```
target_compile_definitions(Foundation PUBLIC POCO_CMAKE)
```
to `Foundation/CMakeLists.txt` and then change `EventlogChannel.cpp` accordingly.

KevDi:
[...]

PR #4803, opened 2024-12-02
Updated CMake to also include the generated pocomessage.rc File

KevDi:
I have added the `pocomessage.rc` in the CMakeLists. Currently it is missing and thats leads to the issue that the `PocoFoundation.dll` does not contains the Message Format for the Eventlog Channel in Windows. So the Eventviewer will show a warning about the missing Information. 
I'm not the CMake Expert so i tried my best to solve it. If there is a better way then please let me know :). 
At the moment i first need to create the File as an empty one because the `custom_command` is first called during the Build. So when running `cmake ..` it is unable to find the file because it is not there. 
This will solve #3304

### DIFF
--- a/Foundation/CMakeLists.txt
+++ b/Foundation/CMakeLists.txt
@@ -15,10 +15,14 @@ POCO_SOURCES_AUTO_PLAT(SRCS WIN32
 	src/EventLogChannel.cpp
 	)
 
+
 # Version Resource
 if(MSVC AND BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


obiltschnig:
For consistency, this should be done in the `POCO_MESSAGE` CMake macro.

obiltschnig:
Proper fix in ff26c8e

KevDi:
@obiltschnig i was unsure what is the best place for this but looks good for me also to put it into the `POCO_MESSAGES` Macro

PR #4804, opened 2024-12-02
Let user can config a ld load flags when use class loader

XinyuKhan:


Foundation/include/Poco/ClassLoader.h
```diff
@@ -139,7 +139,7 @@ class ClassLoader
 		}
 	}
 
-	void loadLibrary(const std::string& path, const std::string& manifest)
+	void loadLibrary(const std::string& path, const std::string& manifest, int flags = 0)
```
aleks-f, line 142: this should default to `SharedLibrary::SHLIB_GLOBAL`

https://github.com/pocoproject/poco/blob/d13fac600a50cdb31bcb3f688a4e9bf3bbf04098/Foundation/include/Poco/SharedLibrary.h#L45

aleks-f (changes requested):

matejk:
@XinyuKhan, do you plan to correct the PR as suggested by @aleks-f ?

matejk:
Replaced with #5105.

PR #4805, opened 2024-12-02
fix(cmake): correct typo AVHAI -> AVAHI

stkw0:
Small typo. It should be AVAHI instead of AVHAI


matejk:
Cherry-picked to poco-1.14.1.

PR #4807, opened 2024-12-03
Poco::Zip - Check archive consistency before extracting 

### LINKED ISSUE #4623: Poco::Zip library should check consistency of archives before extracting
Specifically, file sizes and CRCs in the central directory should be checked for consistency with entries in the file headers, and inconsistent archives shall not be processed.

nyashbox:
Fixes #4623

This is by far the best solution I've come up with. When decompressing all files, archive will be checked for consistency, but when decompressing single file, user should check archive manually. Example of the single file decompression with consistency check can be seen below (Original taken from the [POCO Zip User Guide](https://docs.pocoproject.org/current/ZipUserGuide.html))

```c++
std::ifstream inp("test.zip", std::ios::binary);
poco_assert (inp);
ZipArchive arch(inp);

// Check consistency before going further
try 
{
    arch.checkConsistency();
}
catch (const ZipException &e) 
{
    // Handle error here
}

ZipArchive::FileHeaders::const_iterator it = arch.findHeader("data/hello.txt");
poco_assert (it != arch.headerEnd());
inp.clear();
ZipInputStream zipin(inp, it->second);
std::ostringstream out(std::ios::binary);
Poco::StreamCopier::copyStream(zipin, out);
```

There are some other solutions I've considered: 

- **Put consistency check in ZipArchive's constructors.** But I've decided to refrain from it due to possible unforeseen consequences (and because it's less explicit than calling actual method).
- **Add higher-level API**. This could simplify, for example, extracting single file (IMHO wiki example looks a little bit too long and dubious for such a common task), and also hide consistency checks.


aleks-f (approved):

matejk (commented):
Would it make sense to make it possible to ignore consistency check in case a user would desire to get the file out from the archive anyway?

nyashbox:
I think it absolutely makes sense. Somehow I've missed this idea :(

matejk:
@nyashbox, will you update the PR with optional consistency check?

matejk (approved):

PR #4811, opened 2024-12-03
fix(cmake): remove libatomic dependency

aleks-f:


matejk (changes requested):
Linking `libatomic` came in with a patch that Debian applies in their build system.

I think that it is only required on 32-bit platforms and not 64-bit. Most probably linking can be made private as well for `libatomic`.

aleks-f:
If there is a choice, I'd say better not to fail on current platforms, then fail for the sake of maintaining the old ones. `<atomic>` is 14 years old

matejk (approved):

matejk:
Cherry-picked to poco-1.14.1.

PR #4815, opened 2024-12-07
feat(Process): Process::isRunning returns true for defunct tagged pro…

aleks-f:
…cess #1097 (WiP)

### DIFF
--- a/Foundation/include/Poco/Process_UNIX.h
+++ b/Foundation/include/Poco/Process_UNIX.h
@@ -19,11 +19,14 @@
 
 
 #include "Poco/Foundation.h"
+#include "Poco/Event.h"
+#include "Poco/Mutex.h"
 #include "Poco/RefCountedObject.h"
 #include <unistd.h>
 #include <vector>
 #include <map>
 #include <atomic>
+#include <optional>
 
 
 namespace Poco {
@@ -40,10 +43,13 @@ public:
 
 	pid_t id() const;
 	int wait() const;
+	int wait(int options) const;
 	int tryWait() const;
 
 private:
-	std::atomic<pid_t> _pid;
+	const std::atomic<pid_t> _pid;
+	mutable Event _event;
+	mutable std::optional<std::atomic<int>> _status;
 };
 
 
--- a/Foundation/src/Process_UNIX.cpp
+++ b/Foundation/src/Process_UNIX.cpp
@@ -40,7 +40,8 @@ namespace Poco {
 // ProcessHandleImpl
 //
 ProcessHandleImpl::ProcessHandleImpl(pid_t pid):
-	_pid(pid)
+	_pid(pid),
+	_event(Event::EVENT_MANUALRESET)
 {
 }
 
@@ -58,20 +59,40 @@ pid_t ProcessHandleImpl::id() const
 
 int ProcessHandleImpl::wait() const
 {
+	if (wait(0) != _pid)
+		throw SystemException("Cannot wait for process", NumberFormatter::format(_pid));
+
+	return WEXITSTATUS(_status.value());
+}
+
+
+int ProcessHandleImpl::wait(int options) const
+{
+	if (_status.has_value()) return _pid;
 	int status;
 	int rc;
 	do
 	{
-		rc = waitpid(_pid, &status, 0);
+		rc = waitpid(_pid, &status, options);
 	}
 	while (rc < 0 && errno == EINTR);
-	if (rc != _pid)
-		throw SystemException("Cannot wait for process", NumberFormatter::format(_pid));
+	if (rc == _pid)
+	{
+		_status = status;
+		_event.set();
+	}
+	else if (rc < 0 && errno == ECHILD)
+	{
+		// Looks like another thread was lucky and it should update the status for us shortly
+		_event.wait();
 
-	if (WIFEXITED(status)) // normal termination
-		return WEXITSTATUS(status);
-	else // termination by a signal
-		return 256 + WTERMSIG(status);
+		if (_status.has_value())
+		{
+			rc = _pid;
+		}
+	}
+
+	return rc;
 }
 
 
--- a/Foundation/testsuite/src/ProcessTest.cpp
+++ b/Foundation/testsuite/src/ProcessTest.cpp
@@ -16,6 +16,7 @@
 #include "Poco/PipeStream.h"
 #include "Poco/Path.h"
 #include "Poco/Format.h"
+#include "Poco/Thread.h"
 
 
 using namespace std::string_literals;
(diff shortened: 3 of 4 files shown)


Foundation/src/Process_UNIX.cpp
```diff
@@ -318,26 +358,33 @@ void ProcessImpl::killImpl(PIDImpl pid)
...
 
 bool ProcessImpl::isRunningImpl(PIDImpl pid)
 {
-	if (kill(pid, 0) == 0)
+	int status;
+	int rc;
+	do
 	{
-		return true;
+		rc = ::waitpid(pid, &status, WNOHANG);
```
matejk: **Bug (critical):** This `waitpid` call **reaps** the zombie child and discards the exit status. Since `isRunningImpl(PIDImpl)` is a static method with no access to `ProcessHandleImpl`'s `_hasStatus` / `_event` / `_status` fields, the status is permanently lost.

If user code does:
```cpp
ProcessHandle ph = Process::launch(...);
while (Process::isRunning(ph.id())) { Thread::sleep(100); }
int rc = ph.wait(); // DEADLOCK
```

`handle.wait()` will call `waitpid` → get `ECHILD` (already reaped) → enter the blocking ECHILD branch → `_event.wait()` blocks forever because nobody sets the event.

This is the exact pattern `ProcessRunner` uses (it polls `isRunning(pid)`).

**Suggested fix — use `waitid()` with `WNOWAIT` to peek without reaping:**
```cpp
bool ProcessImpl::isRunningImpl(PIDImpl pid)
{
    siginfo_t info{};
    int rc = ::waitid(P_PID, pid, &info, WEXITED | WNOHANG | WNOWAIT);
    if (rc == 0 && info.si_pid == pid)
        return false;  // zombie/exited, but NOT reaped
    if (rc == 0 && info.si_pid == 0)
        return true;   // still running
    // Not our child or error; fall back to kill check
    if (::kill(pid, 0) == 0) return true;
    return errno == EPERM;
}
```

`WNOWAIT` is POSIX (supported on Linux, macOS, FreeBSD, QNX). This checks zombie status without consuming the waitable state, so a subsequent `handle.wait()` still works.

Foundation/src/Process_UNIX.cpp
```diff
@@ -56,42 +57,81 @@ pid_t ProcessHandleImpl::id() const
...
+				rc = _pid;
+			else
+				rc = 0;
+		}
+		else
+		{
+			_event.wait();
+			if (_hasStatus.load())
+				rc = _pid;
+		}
```
matejk, line 119: **Bug (latent):** If `_hasStatus.load()` is false after `_event.wait()` returns, `rc` is left at the negative value from the failed `waitpid` call. The caller `wait()` then throws a misleading `SystemException("Cannot wait for process")`.

Currently safe because `_hasStatus.store(true)` is sequenced before `_event.set()` (both seq_cst), but there's no `else` clause to catch the impossible case. A future code change reordering those two stores would silently break this.

```suggestion
		else
		{
			_event.wait();
			if (_hasStatus.load())
				rc = _pid;
			else
				throw SystemException("Lost process status (internal error)", NumberFormatter::format(_pid));
		}
```

Foundation/src/Process_UNIX.cpp
```diff
@@ -56,42 +57,81 @@ pid_t ProcessHandleImpl::id() const
...
 	{
-		rc = waitpid(_pid, &status, WNOHANG);
+		rc = ::waitpid(_pid, &status, options);
 	}
 	while (rc < 0 && errno == EINTR);
+	if (rc == _pid)
+	{
+		_status.store(status);
+		_hasStatus.store(true);
+		_event.set();
```
matejk, line 97: **Fragile ordering:** The correctness of the ECHILD handler (lines 100-112) depends on `_hasStatus` being stored **before** `_event.set()`. Specifically, the WNOHANG path at line 102 does `_event.tryWait(0) && _hasStatus.load()` — if these two lines were swapped, that check would have a race window where the event is set but `_hasStatus` is still false.

Please add a comment:
```cpp
_status.store(status);
_hasStatus.store(true);
// N.B. _hasStatus must be stored before _event.set()
// because the ECHILD/WNOHANG path checks tryWait(0) && _hasStatus
_event.set();
```

Alternatively, consider collapsing `_status` + `_hasStatus` into a single `std::atomic<int>` with a sentinel (e.g. `INT_MIN` = "no status yet"). waitpid status is a 16-bit encoded value that can never be `INT_MIN`. This eliminates the ordering dependency entirely and removes one atomic.

Foundation/include/Poco/Process_UNIX.h
```diff
@@ -40,10 +41,16 @@ class Foundation_API ProcessHandleImpl: public RefCountedObject
...
 	pid_t id() const;
 	int wait() const;
+	int wait(int options) const;
 	int tryWait() const;
+	bool isRunning() const;
 
 private:
-	std::atomic<pid_t> _pid;
+	static int statusToExitCode(int status);
+	const std::atomic<pid_t> _pid;
```
matejk: **Nit:** `const std::atomic<pid_t>` is unusual. Since `_pid` is set once in the constructor and never mutated, the `atomic` wrapper serves no purpose — `load()` happens to be const-qualified, but there are no concurrent writers to synchronize with. A plain `const pid_t` would be clearer and avoids unnecessary atomic load barriers on every access.

```suggestion
	const pid_t _pid;
```

Foundation/src/Process_UNIX.cpp
```diff
@@ -56,42 +57,81 @@ pid_t ProcessHandleImpl::id() const
 }
 
 
+int ProcessHandleImpl::statusToExitCode(int status)
+{
+	if (WIFEXITED(status))
+		return WEXITSTATUS(status);
+	else
+		return 256 + WTERMSIG(status);
```
matejk, line 68: **Docs:** The PR description says signal-killed processes return `-WTERMSIG` but this code returns `256 + WTERMSIG`. The code is correct (consistent with the old implementation), the description should be updated.

Also consider adding a brief doc comment since callers need to know the convention:
```cpp
/// Converts a raw waitpid status to a Poco exit code.
/// Normal exit: WEXITSTATUS (0-255).
/// Signal death: 256 + signal_number.
```

Foundation/src/Process_UNIX.cpp
```diff
@@ -56,42 +57,81 @@ pid_t ProcessHandleImpl::id() const
...
-		return 256 + WTERMSIG(status);
+	return statusToExitCode(_status.load());
+}
+
+
+bool ProcessHandleImpl::isRunning() const
+{
+	if (_hasStatus.load())
+		return false;
+	return wait(WNOHANG) == 0;
```
matejk, line 144: **Minor race:** Two threads calling `isRunning()` simultaneously can produce a transient false positive. Thread A reaps the child via `waitpid` and is between `_hasStatus.store(true)` and `_event.set()`. Thread B gets `ECHILD`, calls `_event.tryWait(0)` which returns false → `rc = 0` → `isRunning()` returns `true` even though the process has exited.

Self-correcting on the next poll (since `_hasStatus` is now true), so acceptable for a polling API. Worth a brief comment.

PR #4834, opened 2025-01-02
Missing parameter for enabling FTS5

vrabac6:


matejk (commented):
Do CMake files already have an option to set FTS5?

vrabac6:
> Do CMake files already have an option to set FTS5?

Yes, the CMake files provide an option to enable FTS5 for SQLite. The option `ENABLE_DATA_SQLITE_FTS5` is disabled by default but can be enabled by setting `-DENABLE_DATA_SQLITE_FTS5=ON` during configuration.

PR #4838, opened 2025-01-06
fix(Net) bad mask with odd number of bytes

d3matt:
Receiving an odd number of bytes in non-blocking mode results in using the wrong bytes to unmask the payload. Keep track of the number of payload bytes that have been unmasked.

### DIFF
--- a/Net/src/WebSocketImpl.cpp
+++ b/Net/src/WebSocketImpl.cpp
@@ -262,14 +262,14 @@ void WebSocketImpl::setMaxPayloadSize(int maxPayloadSize)
 }
 
 
-int WebSocketImpl::receivePayload(char *buffer, int payloadLength, char mask[MASK_LENGTH], bool useMask)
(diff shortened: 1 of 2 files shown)


obiltschnig:
I would take `_receiveState.maskOffset` modulo `MASK_LENGTH` to prevent an overflow issue with frame sizes that don't fit into `int`.

Net/src/WebSocketImpl.cpp
```diff
@@ -297,7 +297,7 @@ int WebSocketImpl::receiveBytes(void* buffer, int length, int)
 
 		skipHeader(_receiveState.headerLength);
 
-		if (receivePayload(reinterpret_cast<char*>(buffer), payloadLength, _receiveState.mask, _receiveState.useMask) != payloadLength)
+		if (receivePayload(reinterpret_cast<char*>(buffer), payloadLength, _receiveState.mask, _receiveState.useMask, 0) != payloadLength)
```
austin-beer, line 300: For consistency I would pass in `_receiveState.maskOffset` here as well instead of 0.

PR #4863, opened 2025-02-07
Fix StreamSocket::sendFile()

### LINKED ISSUE #4831: httpS server seems to not work in v1.14
upgraded my existing application to use poco 1.14 with openssl 3.0.15
when using firefox to get a page from the server over ssl it shows a "SSL_ERROR_RX_RECORD_TOO_LONG" error.
chorme did not show an error msg, but was unable to render,  it looks like it kept retrying to get the page.

when using poco 1.13.3 application works.

obiltschnig:
`StreamSocket::sendFile()` is always available and uses a fallback if native `sendfile()` (Linux, macOS) or `TransmitFile()` (WIN32) is not available. Furthermore, optimized methods are not used in case of a secure socket as these functions do not deal with TLS.

Fixes #4831.


matejk (approved):
The changes make a lot of sense to me. I do not have any comments.

PR #4887, opened 2025-02-28
Fix dynamic var conversion precision check

### LINKED ISSUE #4886: Poco::Dynamic::Var conversion to floating point does not handle precision checks correctly for negative numbers.
**Describe the bug**
When converting a negative integer to a floating point value, a Poco::RangeException is trown for values that do not lose precision when converted to float/double. This seems to be a bug in Dynamic/VarHolder.h::numValDigits, called from isPrecisionLost in the same file:
```
template <typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
static constexpr int numValDigits(const T& value)
{
	using U = std::make_unsigned_t<T>;
	if (value == 0) return 0;
	int digitCount = 0;
	U locVal = value; // to prevent sign preservation
	while (locVal >>= 1) ++digitCount;
	return digitCount;
}
```
The line `U locVal = value; // to prevent sign preservation` assigns the two's compliment value to locVal, and as such does not return the correct number of significant digits. 

**To Reproduce**

```
#include "Poco/Dynamic/Var.h"

int main(void)
{
  Poco::Dynamic::Var value = (int64_t)-10;
  double             answer = value.convert<double>();
  return (0);
}
```
Running this code results in:
```
terminate called after throwing an instance of 'Poco::RangeException'
  what():  Out of range
```

**Expected behavior**
The value stored in the Poco:Dynamic::Var should be converted to the equivalent double value if the original value fits within a double without precision loss and not throw an unnecessary exception. In the above example, -10 should be stored in the answer variable.

**Logs**
n/a

**Screenshots**
n/a

 - OS: Debian 12 (more specifically the vscode Debian c++ development container mcr.microsoft.com/vscode/devcontainers/cpp:1.2.7-debian-12) 
 - POCO Version: 1.14.1

**Additional context**
n/a

MduPlessis-1:
Take the absolute value of the value in numValDigits to fix the precision loss check for negative values. Fixes #4886.

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -475,7 +475,7 @@ private:
 		using U = std::make_unsigned_t<T>;
 		if (value == 0) return 0;
 		int digitCount = 0;
-		U locVal = value; // to prevent sign preservation
+		U locVal = llabs(value); // to prevent sign preservation
 		while (locVal >>= 1) ++digitCount;
 		return digitCount;
 	}


aleks-f (changes requested):
@MduPlessis-1 please add a test for this

MduPlessis-1:
> instead of `llabs()`, use `std::abs()`

Using std::abs() here causes a compilation error for ambiguous overloads of abs() since this template can be called on unsigned types. So the options are:

1. Keep it as-is
2. U locVal = -static_cast<unsigned long>(value);  (refer to [stackoverflow](https://stackoverflow.com/questions/17313579/is-there-a-safe-way-to-get-the-unsigned-absolute-value-of-a-signed-integer-with) for details)
3. Some way to have special code for signed/unsigned types.

matejk:
> > instead of `llabs()`, use `std::abs()`
> 
> Using std::abs() here causes a compilation error for ambiguous overloads of abs() since this template can be called on unsigned types. So the options are:
> 
> 1. Keep it as-is
> 2. `U locVal = -static_cast<U>(value);`  (refer to [stackoverflow](https://stackoverflow.com/questions/17313579/is-there-a-safe-way-to-get-the-unsigned-absolute-value-of-a-signed-integer-with) for details)
> 3. Some way to have special code for signed/unsigned types.

One solution would also be to use:
```
U locVal = (value < 0) ? -value : value;
```

obiltschnig:
Something involving `std::is_signed_t<>` and `std::make_signed_t<>` (to resolve the ambiguity) maybe?

matejk:
Replaced with PR #4894

PR #4893, opened 2025-03-10
Prebuild does not work as intended

vrabac6:


aleks-f (changes requested):
`pocoproject.org` IP address changed
```
There was 1 failure: 
 1: class CppUnit::TestCaller<class SocketAddressTest>.testSocketAddress
    "sa4.host().toString() == "54.93.62.90""
    in "D:\a\poco\poco\Net\testsuite\src\SocketAddressTest.cpp", line 76
$ nslookup pocoproject.org
Server:		192.168.0.1
Address:	192.168.0.1#53

Non-authoritative answer:
Name:	pocoproject.org
Address: 157.90.17.168
```

PR #4894, opened 2025-03-11
Fix dynamic var conversion precision check (#4886)

### LINKED ISSUE #4886: Poco::Dynamic::Var conversion to floating point does not handle precision checks correctly for negative numbers.

matejk:
Fixes #4886, replaces #4887.

### DIFF
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -475,7 +475,7 @@ private:
 		using U = std::make_unsigned_t<T>;
 		if (value == 0) return 0;
 		int digitCount = 0;
-		U locVal = llabs(value); // to prevent sign preservation
+		U locVal = (std::is_signed_v<T> && value < 0) ? -value : value; // to prevent sign preservation
 		while (locVal >>= 1) ++digitCount;
 		return digitCount;
 	}
--- a/Foundation/include/Poco/Dynamic/VarHolder.h
+++ b/Foundation/include/Poco/Dynamic/VarHolder.h
@@ -475,7 +475,7 @@ private:
 		using U = std::make_unsigned_t<T>;
 		if (value == 0) return 0;
 		int digitCount = 0;
-		U locVal = value; // to prevent sign preservation
+		U locVal = llabs(value); // to prevent sign preservation
 		while (locVal >>= 1) ++digitCount;
 		return digitCount;
 	}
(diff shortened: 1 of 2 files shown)


aleks-f:
this is causing warnings on windows:
```
C:\Users\alex\source\repos\pocoproject\poco\Foundation\include\Poco\Dynamic\VarHolder.h(478,51): warning C4146: unary minus operator applied to unsigned type, result still unsigned
1>(compiling source file 'src/JSONConfiguration.cpp')
1>    C:\Users\alex\source\repos\pocoproject\poco\Foundation\include\Poco\Dynamic\VarHolder.h(478,51):
1>    the template instantiation context (the oldest one first) is
1>        C:\Users\alex\source\repos\pocoproject\poco\Foundation\include\Poco\Dynamic\VarHolder.h(2124,3):
1>        see reference to function template instantiation 'void Poco::Dynamic::VarHolder::convertToFP<const Poco::UInt32,float,true,true>(F &,T &)' being compiled
1>        with
1>        [
1>            F=const Poco::UInt32,
1>            T=float
1>        ]
```

Issue #4892, opened 2025-03-10
TCPServer continues to accept connections after stop()


aleks-f:
**Is your feature request related to a problem? Please describe.**
Perhaps I'm missing something, but when there are connection attempts after the `TCPServer::stop()`, the connections continue to be accepted, without a way to prevent those connections to be established and run its course.

**Describe the solution you'd like**
- [x] Close the socket on stop, set `_stopped` to true, and ignore the exceptions in `run()` after that.
- [x] add `stop()` to `TCPServerConnectionFactory` and allow it to return `nullptr`, in which case any accepted socket after TCPServer stop will die on their own

micheleselea:
I understand that TCPServer::stop() put 
_stopped = true;
_thread.join();
So I believe the thread that accept new connection is stopped and the server socket is not accepting new connection, Am I missing something?

aleks-f:
> Am I missing something?

yes, the case when the thread is here:
https://github.com/pocoproject/poco/blob/91244ac9d9b682b8f271b7373ac5d691fa520d4e/Net/src/TCPServer.cpp#L137

when `TCPServer::stop()` is called.

PR #4896, opened 2025-03-11
fix(TCPServer): continues to accept connections after stop() #4892

aleks-f:


Net/src/TCPServer.cpp
```diff
@@ -150,24 +151,30 @@ void TCPServer::run()
 				}
 				catch (Poco::Exception& exc)
 				{
-					ErrorHandler::handle(exc);
+					if (!_stopped)
```
matejk, line 154: This if is always true: _stopped is true at this point.

aleks-f: > This if is always `false`: `_stopped` is `true` at this point.

`acceptConnection()` can throw for many reasons other than `_socket.close()` being called on `stop()`

aleks-f:
looks like the only way to do this is to have a thread-safe sockfd wrapper, or take the other route and allow nullptr return from connection factory

obiltschnig:
In `TCPServer::run()`, when `_socket.poll()` returns `true`, check `_stopped` again before calling `_socket.acceptConnection()`, and if it's set, `return`. That should prevent any new connections from being accepted after `stop()` has been called.

obiltschnig:
Also, first calling `TCPServer::stop()` and then killing existing connections may prevent aggressive clients from reconnecting.

obiltschnig:
If you handle this in the factory, by the time the factory returns null the connection has already been accepted and put in the dispatcher queue. Not calling accept() in the first place is preferable.

aleks-f:
> If you handle this in the factory, by the time the factory returns null the connection has already been accepted and put in the dispatcher queue. Not calling accept() in the first place is preferable.

I am not opposed to having both. But doing it on accept is still not 100% bulletproof. If factory returns null, the accepted socket will die with the notification. Additionally, it gives more control to the factory, eg. it can limit the number of connections it is willing to create or refuse connections from certain clients etc.

obiltschnig:
I'm ok with allowing the `TCPServerConnectionFactory` return null.

aleks-f:
@obiltschnig do you want me to leave the sockfd atomic? there may be some benefit from it, but it is not nearly a proper solution for socket thread safety

PR #4903, opened 2025-03-19
fix(VarHolder): MSVC warning #4886

aleks-f:


Foundation/include/Poco/Dynamic/VarHolder.h
```diff
@@ -469,23 +469,32 @@ class Foundation_API VarHolder
 
 private:
 
-	template <typename T, std::enable_if_t<std::is_integral_v<T>&& std::is_unsigned_v<T>, bool> = true>
+	template <typename T, std::enable_if_t<std::is_signed_v<T>, bool> = true>
```
matejk, line 472: `unpreserveSign` can be implemented with a single function using `constexpr `like this:

```
    template <typename T>
	static constexpr T unpreserveSign(const T& value)
	{
		if constexpr (std::is_signed_v<T>)
		{
			return (value < 0) ? -value : value;
		}
		else
		{
			return value;
		}
	}
```

aleks-f: that's fine if no compiler warns. I can't spend more time on it right now, but if you want to change it, go ahead

matejk: No warnings.

PR #4937, opened 2025-04-17
CMake: make utf8proc, PCRE2, zlib and expat private dependencies

lopsided98:
In the unbundled build, utf8proc, PCRE2, zlib and expat were all included as part of the CMake link interface, and therefore the CMake config file required `find_dependency()` calls for each.

These libraries are internal implementation details of Poco and are not exposed in any public API (as of https://github.com/pocoproject/poco/pull/4724). Therefore, we can make them private dependencies and avoid the find_dependency() calls and their associated find scripts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -562,11 +562,6 @@ install(
 		Devel
 )
 
-if(POCO_UNBUNDLED)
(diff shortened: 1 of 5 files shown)


matejk:
Thanks for this contribution.

PR #4944, opened 2025-05-05
fix(Mutex_POSIX): prevent deadlocks when logging inside signal handlers

### LINKED ISSUE #4939: Deadlock when loging during an signal handler
**Describe the bug**
We use Linux signals to coordinate our application shutdown.
The signal handler sets the appropriate flags and also logs the received signal.

It's possible that the application is already logging a message when the signal is received and the `FileChannel` and  `Channel` is already in use. 
Is this is the case, a deadlock can occurs.
This can be avoided when using an rekursiv mutex 

**To Reproduce**
Log messages when receiving an Linux Signal

**Expected behavior**
No deadlock

 - All Linux systems
 - Poco: 1.13.1

ssooffiiaannee:
Fixes #4939.


obiltschnig:
I would not want that additional overhead on all calls to lock().

ssooffiiaannee:
Is there anything I should do the CIFuzz yaml file ? v1.13.1 contains artifact@v3.
 I see Error: Missing download info for actions/upload-artifact@v3 and yaml is v4 since 3 months ago.

Issue #4978, opened 2025-07-07
ODBC MSSQL bulk insert UNIQUEIDENTIFIER type does not work


solarobot:
HI, 

I am using poco data to do bulk insert, while it falls into error when facing UNIQUEIDENTIFIER data type,
Tried std::string and Poco::UUID at code side, however not helped.
	```
int rows = 100000;
	std::vector<std::string> oids(rows);

	for (int i = 0; i < rows; ++i) {
		oids[i] = Poco::UUIDGenerator::defaultGenerator().createRandom().toString();
	}
	Statement stmt(*m_pSession);
	std::string query("CREATE TABLE ##Person (oid UNIQUEIDENTIFIER, FirstName NVARCHAR(MAX), BORN DATETIME, Age TINYINT)");
	stmt.executeDirect(query);

	int age = 42;
	std::vector<int> ageVec(rows, age);
	int count = -1;
	m_pSession->setFeature("autoBind", true);
	m_pSession->setFeature("autoExtract", true);
	*m_pSession << "select count(*) from ##Person;"s, Keywords::into(count), Keywords::now;
	*m_pSession << "INSERT INTO ##Person (oid) VALUES (?)"s,
		use(oids, bulk), Keywords::now;
	*m_pSession << "select count(*) from #Person;"s, Keywords::into(count), Keywords::now;

```

**Comment by solarobot** [comment]
HI,
I am trying to specialize the binding 
```
#pragma once
#include <Poco/Data/AbstractBinding.h>
#include <Poco/Data/TypeHandler.h>
#include <Poco/UUID.h>
#include <Poco/Data/BulkBinding.h>
#include <Poco/Data/ODBC/ODBCStatementImpl.h>

#include <vector>
#include <stdexcept>
#include <sqltypes.h>
#include <Poco/Data/ODBC/Binder.h>


using namespace Poco::Data::ODBC;

namespace Poco {
	namespace Data {

		template <>
		class BulkBinding<std::vector<Poco::UUID>> : public AbstractBinding
			/// A BulkBinding maps a value to a column.
			/// Bulk binding support is provided only for std::vector.
		{
		public:
			BulkBinding(const std::vector<Poco::UUID>& val, Poco::UInt32 bulkSize, const std::string& name = "", Direction direction = PD_IN) :
				AbstractBinding(name, direction, bulkSize),
				_val(val),
				_bound(false)
				/// Creates the BulkBinding.
			{
				if (0 == _val.size())
[...]

PR #4979, opened 2025-07-09
Data/ODBC: Add bulk binding for UUID containers

solarobot:
add implementation of bulkbinding uuid containers for odbc binder.
issue #4978

### DIFF
--- a/Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
+++ b/Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
@@ -1143,6 +1143,7 @@ CppUnit::Test* ODBCSQLServerTest::suite()
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testPrepare);
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testBulk);
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testNullBulk);
+		CppUnit_addTest(pSuite, ODBCSQLServerTest, testUUIDsBulk);
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testBulkPerformance);
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testSetSimple);
 		CppUnit_addTest(pSuite, ODBCSQLServerTest, testSetComplex);


aleks-f:
@solarobot  we can't accept this. you should branch off and send PR towards the [main](https://github.com/pocoproject/poco/tree/main) branch

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
```diff
@@ -417,6 +417,30 @@ catch(Poco::Exception& ex)
 }
 }
 
+void ODBCSQLServerTest::testUUIDsBulk()
```
aleks-f, line 420: ok, but this test case will never run, unless it is added [here](https://github.com/pocoproject/poco/pull/4979/files#diff-b3e0f02c34e11ef60b0cdc953127453088585369e052aab5025361a513e96a1dL1120)

aleks-f (changes requested):

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
solarobot: just added the missing line.

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
```diff
@@ -417,6 +417,30 @@ catch(Poco::Exception& ex)
...
+{
+	try
+	{
+		if (!_pSession) fail("Test not available.");
+
+		_pSession->setFeature("autoBind", true);
+		_pSession->setFeature("autoExtract", true);
+
+		recreateUUIDsTable();
```
aleks-f, line 429: where exactly is this function? CI does not build or run ODBC tests, you have to do it yourself and make sure it works as intended

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
```diff
@@ -417,6 +417,30 @@ catch(Poco::Exception& ex)
...
+		int rows = 1000;
+		std::vector<Poco::UUID> uuids(rows);
+		for (int i = 0; i < rows; ++i) {
+			uuids[i]= Poco::UUIDGenerator::defaultGenerator().createRandom();
+		}
+
+		*_pSession << "INSERT INTO Strings VALUES (?)"s, use(uuids, bulk), Poco::Data::Keywords::now;
```
aleks-f, line 436: what does this exactly do? Please make sure your PR compiles and executes the test (which should actually test the new functionality).

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
solarobot: while the change was compiled here on my machine and it worked as expected in my production environment.
the code above is to create the test database table with a single one column which is uniqueidentifer column type.
recreateUUIDsTable was existing there already, i just reuse it for this specific purpose.

Data/ODBC/testsuite/src/ODBCSQLServerTest.cpp
solarobot: use(uuids, bulk) tells the binder to do bulkbinding, which will call the added implementation for UUID data type.

PR #4983, opened 2025-07-20
New SQLite Only Unbundled Option

andrewauclair:
A new option to use an unbundled (external) SQLite but all other dependencies internal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,9 @@ option(ENABLE_FUZZING
 option(POCO_UNBUNDLED
 	"Set to OFF|ON (default is OFF) to control linking dependencies as external" OFF)
 
+option(POCO_SQLITE_UNBUNDLED
+	"Set to OFF|ON (default is OFF) to control linking sqlite dependency as external" OFF)
+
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -267,6 +270,9 @@ endif()
 
 if(POCO_UNBUNDLED)
```
matejk, line 275: POCO_UNBUNDLED could set POCO_SQLITE_UNBUNDLED when it is not set explicitly. That would make a few conditions that follow a bit simpler and would give a base for introducing more fine grained unbundled options.

andrewauclair: Done. That was a little more annoying than I thought it might be. The option gets a value no matter what and cannot be changed. But that can be worked around by setting a variable with the same name to ON before the option gets set. Looks to work well. I am going to test it all out again tomorrow, assuming I can get all the dependencies setup externally.

andrewauclair:
Looks good in both unbundled modes. CI fails for linux-gcc-make-odbc, but it appears to fail for other PRs as well.

PR #4996, opened 2025-08-11
Reorganise external libs to separate subdirectory

### LINKED ISSUE #4358: Reorganise source code of external libraries (pre-requisite for #4354)

matejk:
Closes #4358.

Source code of external libraries is moved to subdirectory `dependencies` where each of the libraries has its own subdirectory.

When creating a bundled build, a static library is created and then linked where appropriate. For unbundled builds system provided library is linked.

This mechanism allows cleaner separation of Poco and other code, easier upgrade and the possibility to have fine-grained bundled options per external library.


matejk:
- External sources moved to `dependendies`
- static (object) libraries built and linked to Poco libraries
- Poco code and CMake files updated
- Most CI tests converted to build with CMake instead of make/VS
- fixed a few issues detected with thread sanitizer

dependencies/hpdf/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+target_include_directories(_BUNDLED_HPDF
+	PUBLIC
+		$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+)
+
+if (POCO_UNBUNDLED)
+	target_link_libraries(_BUNDLED_HPDF PRIVATE PNG::PNG)
+else()
+	target_link_libraries(_BUNDLED_HPDF PRIVATE "$<BUILD_LOCAL_INTERFACE:PNG::PNG>")
```
andrewauclair, line 23: Is there an alternate solution to the BUILD_LOCAL_INTERFACE generator expression here? I think it would be good if Poco kept the cmake minimum to 3.15, or at least not jumping all the way to 3.26.

matejk: I was not able to find other solution that does not involve writing own custom CMake scripts. And CMake 3.26 is the first version that has clean solution to link private Poco library and not get undefined export errors.

matejk:
Replaced all Make and VS jobs in the CI with CMake.

PR is ready to be merged.

Issue #5002, opened 2025-08-22
Oracle DB Integers


aleks-f:
Oracle essentially has no integer type; we tried to [work around it](https://github.com/pocoproject/poco/blob/dc1c8e1cd5d1917cf9deb439ac3936a4d8788e35/Data/ODBC/src/ODBCMetaColumn.cpp#L134-L143) so that we don't have to use double for any numeric value, but newer ODBC drivers (v23 in particular) take an issue with it:

Binding is ok:

```
[ODBC][22923][1755812435.900928][SQLBindCol.c][245]
		Entry:
			Statement = 0xaaaae95f8370
			Column Number = 1
			Target Type = -25 SQL_C_SBIGINT
			Target Value = 0xaaaae9601598
			Buffer Length = 8
			StrLen Or Ind = 0xaaaae9600b80
[ODBC][22923][1755812435.900930][SQLBindCol.c][353]
		Exit:[SQL_SUCCESS]
```

But fetching the data is not:
```
[ODBC][22923][1755812435.901763][SQLFetch.c][352]
		Exit:[SQL_ERROR]
[ODBC][22923][1755812435.901767][SQLGetDiagRec.c][681]
		Entry:
			Statement = 0xaaaae95f8370
			Rec Number = 1
			SQLState = 0xfffffc65a0a0
			Native = 0xfffffc65a2a8
			Message Text = 0xfffffc65a0a6
			Buffer Length = 513
			Text Len Ptr = 0xfffffc65a098
		DIAG [HY004] [Oracle][ODBC]Invalid SQL data type <-25>.
```

And so, I guess all numbers in Oracle will have to be treated as double.

PR #5003, opened 2025-08-22
fix(Data/ODBC): Oracle DB Integers #5002

aleks-f:


matejk:
I don't know the underlying problem well enough to be able to provide meaningful comments on the changes.

Data/DataTest/src/SQLExecutor.cpp
```diff
@@ -2849,8 +2849,16 @@ void SQLExecutor::internalExtraction()
 		}
 		catch(BadCastException&)
 		{
-			Poco::Int64 l = rset.value<Poco::Int64>(0,0);
-			assertTrue (1 == l);
+			try
```
matejk, line 2852: @aleks-f, is it possible to know in advance that the database/driver does not support integers and adapt accordingly without using exceptions?

aleks-f: It is possible to find out what back-end we are dealing with, but I would not spend time on it here - I think this is good enough for testing purposes

https://github.com/pocoproject/poco/blob/267cb41daaea45001d9761c9128ba1e05c8a1089/Data/ODBC/include/Poco/Data/ODBC/Utility.h#L177-L187

PR #5019, opened 2025-09-10
Fix Can not resolve "localhost" if not Network adapter exists #4988

### LINKED ISSUE #4988: Can not resolve "localhost" if not Network adapter exists
**Describe the bug**
On a Windowos PC without network adapter (tried in Windows Sandbox with disabled network adapter) poco throws a "Host not found" exception.  
for: 

* Poco::Net::DNS::resolve("localhost");
* Poco::Net::DNS::thisHost();
 
**To Reproduce**
Execute application in Windows Sandbox with deacitivated network adapter. (might be reproducable in docker)
```
    try
    {
        auto c = Poco::Net::DNS::resolve("localhost");
    }
    catch(Poco::Exception & e)
    {
        std::cout << "resolve: " << e.what() << std::endl;
    }
    try
    {
        auto c = Poco::Net::DNS::thisHost();
    }
    catch(Poco::Exception & e)
    {
        std::cout << "thisHost: " << e.what() << std::endl;
    }
```
resolve: Host not found
thisHost: Host not found

**Expected behavior**
Localhost should be resolvable. E.g. to 127.0.0.1 or ::1

 - Wind 11
 - POCO 1.13.4

**Additional context**
```
ipconfig /all
Windows IP Configuration
   Host Name . . . . . . . . . . . . : <>
   Primary Dns Suffix  . . . . . . . :
   Node Type . . . . . . . . . . . . : Mixed
   IP Routing Enabled. . . . . . . . : No
   WINS Proxy Enabled. . . . . . . . : No
``` 

* GetAdaptersInfo() returns 232 (ERROR_NO_DATA No adapter information exists for the local computer.)
* getaddrinfo() returns 11004  ( WSANO_DATA  [see](https://learn.microsoft.com/en-us/windows/security/application-security/application-isolation/windows-sandbox/windows-sandbox-configure-using-wsb-file) )

diridari:
Fill HostEntry with localhost and a default loopback address if 'getaddrinfo' returns 'WSANO_DATA'

### DIFF
--- a/Net/src/DNS.cpp
+++ b/Net/src/DNS.cpp
@@ -82,7 +82,7 @@ HostEntry DNS::hostByName(const std::string& hostname, unsigned
 		freeaddrinfo(pAI);
 		return result;
 	}
-#if defined(_WIN32)
+#if defined(POCO_OS_FAMILY_WINDOWS)
 	else if(rc == WSANO_DATA && hostname == "localhost") // no data record found and is local host 
 	{
  #if defined(POCO_HAVE_IPv6)
@@ -141,7 +141,7 @@ HostEntry DNS::hostByAddress(const IPAddress& address, unsigned
 			freeaddrinfo(pAI);
 			return result;
 		}
-#if defined(_WIN32)
+#if defined(POCO_OS_FAMILY_WINDOWS)
 		else if(rc == WSANO_DATA && address.isLoopback()) // no data record found and is local host 
 		{
 			return HostEntry("localhost", address);
--- a/Net/src/DNS.cpp
+++ b/Net/src/DNS.cpp
@@ -25,7 +25,6 @@
 #include "Poco/Unicode.h"
 #include <cstring>
 
-
 #if defined(POCO_HAVE_LIBRESOLV)
 #include <resolv.h>
 #endif
@@ -83,6 +82,16 @@ HostEntry DNS::hostByName(const std::string& hostname, unsigned
 		freeaddrinfo(pAI);
 		return result;
 	}
+#if defined(_WIN32)
+	else if(rc == WSANO_DATA && hostname == "localhost") // no data record found and is local host 
+	{
+ #if defined(POCO_HAVE_IPv6)
+		return HostEntry("localhost", Poco::Net::IPAddress("::1"));
+ #else
+		return HostEntry("localhost",Poco::Net::IPAddress("127.0.0.1"));
+#endif // ipv6
+	}
+#endif // _WIN32
 	else
 	{
 		aierror(rc, hostname);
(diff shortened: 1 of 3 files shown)


Net/src/DNS.cpp
```diff
@@ -83,6 +82,16 @@ HostEntry DNS::hostByName(const std::string& hostname, unsigned
 		freeaddrinfo(pAI);
 		return result;
 	}
+#if defined(_WIN32)
```
matejk: Use POCO_OS_FAMILY_WINDOWS instead.

Net/src/DNS.cpp
```diff
@@ -132,6 +141,13 @@ HostEntry DNS::hostByAddress(const IPAddress& address, unsigned
 			freeaddrinfo(pAI);
 			return result;
 		}
+#if defined(_WIN32)
```
matejk: Use POCO_OS_FAMILY_WINDOWS

Net/src/DNS.cpp
```diff
@@ -132,6 +141,13 @@ HostEntry DNS::hostByAddress(const IPAddress& address, unsigned
...
+		else if(rc == WSANO_DATA && address.isLoopback()) // no data record found and is local host
```
matejk, line 141: Does it make sense putting this condition first, before checking the "localhost" hostname?

diridari: i would say checking the rc value is much cheaper than checking for localhost or loopback and is less common.

matejk: I apologise for not being clear enough.

In line 86 (above) it is first checked if the hostname is localhost and if that is not true then in line 145 is is checked if the device is a loopback.

From the logic point of view, what shall be the order? localhost first of loopback first? Or it does not matter?

Net/src/DNS.cpp
```diff
@@ -83,6 +82,16 @@ HostEntry DNS::hostByName(const std::string& hostname, unsigned
 		freeaddrinfo(pAI);
 		return result;
 	}
+#if defined(POCO_OS_FAMILY_WINDOWS)
+	else if(rc == WSANO_DATA && hostname == "localhost") // no data record found and is local host 
+	{
+ #if defined(POCO_HAVE_IPv6)
```
aleks-f: I don't think it's a good idea to default to IPv6 solely based on it being present on the system where poco was compiled. Not even the presence of IPv6 on the runtime system would warrant it being default, given the prevalence of IPv4.

Bottom line, I would prefer this to be decided **at runtime**, based on an **additional function argument**, defaulting to **IPv4**.

aleks-f (changes requested):

Net/src/DNS.cpp
diridari: @aleks-f  Since this is an edge case i am not sure if a addition argument wouldn't confuse users. 
Does an function exists, that determines if ipv6 is available? 
Otherwise i would use 127.0.0.1

Net/src/DNS.cpp
diridari: @matejk  Those are different methods: 
First is 'DNS::hostByName()' line 63-114, second is 'DNS::hostByAddress' from line 117-176

obiltschnig:
I don't think Poco should do any name resolution at all outside of that provided by the operating system's network stack.

diridari:
@obiltschnig i can not respond to you comment. 
I agree with your statement, but in this case Linux and Windows do behave differently. 
On linux systems 127.0.0.1 and localhost can be resolved on windows an exception is thrown. See #4988

Issue #5011, opened 2025-09-02
Poco::File crashed when handle windows very long path(\\?\C:)


siren186:
**Describe the bug**

```cpp
Poco::File f1("C:");
Poco::File f2("\\\\?\\C:"); // long path
bool b1 = f1.exists(); // ok
bool b2 = f2.exists(); // crash !!!!!!

std::string path = "C:\\abc\\def\\"; // folder [abc] and [def] are all not exists
Poco::File f3("\\\\?\\" + path); // long path
Poco::File f4(path);
f3.createDirectories(); // crash !!!!!!
f4.createDirectories(); // ok
```

**Error occured here**

https://github.com/pocoproject/poco/blob/5886ee818d89967562185a4b2715c508515be2a2/Foundation/src/File_WIN32U.cpp#L110

```cpp
DWORD dw1 = GetFileAttributesW(L"C:");          // dw1 = 0x10
DWORD dw2 = GetFileAttributesW(L"C:\\");        // dw2 = 0x16
DWORD dw3 = GetFileAttributesW(L"\\\\?\\C:\\"); // dw3 = 0x16

// Poco::File error here
DWORD dw4 = GetFileAttributesW(L"\\\\?\\C:");   // dw4 = -1, GetLastError = ERROR_INVALID_PARAMETER
```

**Please add relevant environment information:**

 - Visual studio 2022
 - POCO v1.14.2

matejk:
@siren186, can you investigate, fix and prepare PR?

PR #5021, opened 2025-09-12
fix(File): Poco::File crashed when handle windows very long path(\\?\C:) #5011

siren186:
See https://github.com/pocoproject/poco/issues/5011

### DIFF
--- a/Foundation/src/File_WIN32U.cpp
+++ b/Foundation/src/File_WIN32U.cpp
@@ -57,25 +57,10 @@ FileImpl::FileImpl()
 
 FileImpl::FileImpl(const std::string& path): _path(path)
 {
-	// If the path only has disk letter, we must make sure it ends with backslash!
-	// Or it will be failed to call Windows API GetFileAttributesW().
-	// For example:
-	//   DWORD dw = GetFileAttributesW(L"\\\\?\\C:"); // dw == -1
 	std::string::size_type n = _path.size();
-	if (n > 1)
+	if (n > 1 && (_path[n - 1] == '\\' || _path[n - 1] == '/') && !((n == 3 && _path[1]==':')))
 	{
-		auto ch = _path.back();
-		if (ch == ':')
-		{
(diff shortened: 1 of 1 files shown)


Foundation/src/File_WIN32U.cpp
```diff
@@ -57,10 +57,25 @@ FileImpl::FileImpl()
...
-	if (n > 1 && (_path[n - 1] == '\\' || _path[n - 1] == '/') && !((n == 3 && _path[1]==':')))
+	if (n > 1)
 	{
-		_path.resize(n - 1);
+		auto ch = _path.back();
+		if (ch == ':')
+		{
+			_path.push_back('\\'); // Add backslash near to disk letter
+		}
+		else if (ch == '\\' || ch == '/')
```
matejk: Why is it necessary to remove the trailing (back)slash at all?

Foundation/src/File_WIN32U.cpp
```diff
@@ -81,10 +96,25 @@ void FileImpl::swapImpl(FileImpl& file)
...
```
matejk: Same question as above?

Foundation/src/File_WIN32U.cpp
```diff
@@ -57,10 +57,25 @@ FileImpl::FileImpl()
 
 FileImpl::FileImpl(const std::string& path): _path(path)
 {
+	// If the path only has disk letter, we must make sure it ends with backslash!
```
matejk: The logic to manipulate input path shall be moved to covertPath to have it in the single place only once.

Foundation/src/File_WIN32U.cpp
```diff
@@ -81,10 +96,25 @@ void FileImpl::swapImpl(FileImpl& file)
 void FileImpl::setPathImpl(const std::string& path)
 {
 	_path = path;
+	// If the path only has disk letter, we must make sure it ends with backslash!
```
matejk: Move the logic to convertPath to avoid code duplication.

Issue #5017, opened 2025-09-08
Poco::Data ODBC recordset is reading previous row value


solarobot:
**Describe the bug**
Poco::Data ODBC recordset is reading previous row value

**To Reproduce**
See the case, the relationdef value of row NO. 6 is null while the returned value is BB212A83-E455-4A0D-A9FD-A60826CE6C2E which is from row NO. 5,

<img width="1122" height="603" alt="Image" src="https://github.com/user-attachments/assets/92aa96c9-ec16-4396-9e39-dcc9bc3be793" />

**Expected behavior**
recordset.movenext() should return the correct row values..
		
```
stmt << "SELECT oid, RelationDef, Property, PropertyType, DateCreated, DateLastModified, IsDeleted "
			<< "FROM [dbo].[PropertyDef]";
		stmt.execute();
		RecordSet rsPropertyDef(stmt);

		more = rsPropertyDef.moveFirst();
		while (more) {
			PropertyDef pd;
			RecordSetPopulator<PropertyDef>::populateFromRecordSet(pd, rsPropertyDef);
			m_propDefs.emplace(pd.oid, pd);
			more = rsPropertyDef.moveNext();
		}
```

If I use an "order by" to make the null values first then non-null values, it returns vales correct.
```
stmt.reset();
stmt << "SELECT oid, RelationDef, Property, PropertyType, DateCreated, DateLastModified, IsDeleted "
	<< "FROM [dbo].[PropertyDef] order by RelationDef";
stmt.execute(); 
```

While this cannot be used as a workaround since there will be multiple nullable columns.

**Please add relevant environment information:**
 - Windows 10, SQL Server 2022
 - POCO 1.14.2

matejk:
@aleks-f , was this noticed already?

matejk:
@solarobot , can you fix it and prepare a PR?

solarobot:
> [@solarobot](https://github.com/solarobot) , can you fix it and prepare a PR?

I will investigate and tell my result.

solarobot:
uuid is prepared as char array
`prepareCharArray<char, DT_CHAR_ARRAY>(pos, SQL_C_BINARY, 16, 16);` <- here gives a size = 16, and length =16, it is over allocating memory

this results the actual size is put into _lenLengths[pos][0]
```
template <typename T, DataType DT>
void prepareCharArray(std::size_t pos, SQLSMALLINT valueType, std::size_t size, std::size_t length)
[...]

PR #5022, opened 2025-09-12
Data/ODBC: Fix UUID column isNull() detection in Preparator

solarobot:
#5017 
use prepareVariableLen instead so that the actual size can be determined by isNull(pos, row) in extractor.

```
inline void Preparator::prepare(std::size_t pos, const Poco::UUID&)
{
	//use prepareVariableLen instead so that the actual size can be determined by isNull(pos, row) in extractor.
	prepareVariableLen<char>(pos, SQL_C_BINARY, 16, DT_CHAR);
	//prepareCharArray<char, DT_CHAR_ARRAY>(pos, SQL_C_BINARY, 16, 16);
}
```

### DIFF
--- a/Data/ODBC/include/Poco/Data/ODBC/Preparator.h
+++ b/Data/ODBC/include/Poco/Data/ODBC/Preparator.h
@@ -554,7 +554,7 @@ private:
 
 			case MetaColumn::FDT_UUID:
 				if (pVal)
-					return prepareFixedSize<Poco::UUID>(pos, SQL_C_BINARY, 16);
+					return prepareFixedSize<Poco::UUID>(pos, SQL_C_BINARY, pVal->size());
 				else
 					return prepareFixedSize<Poco::UUID>(pos, SQL_C_BINARY);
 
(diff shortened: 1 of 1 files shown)


Data/ODBC/include/Poco/Data/ODBC/Preparator.h
```diff
@@ -1185,7 +1185,9 @@ inline void Preparator::prepare(std::size_t pos, const std::list<Poco::DateTime>
 
 inline void Preparator::prepare(std::size_t pos, const Poco::UUID&)
 {
-	prepareCharArray<char, DT_CHAR_ARRAY>(pos, SQL_C_BINARY, 16, 16);
+	//use prepareVariableLen instead so that the actual size can be determined by isNull(pos, row) in extractor.
+	prepareVariableLen<char>(pos, SQL_C_BINARY, 16, DT_CHAR);
+	//prepareCharArray<char, DT_CHAR_ARRAY>(pos, SQL_C_BINARY, 16, 16);
```
aleks-f: these comments are not needed

solarobot: > these comments are not needed

the commit has been updated.

Issue #5025, opened 2025-09-14
PostgreSQL reconnect() doesn't restore session state correctly


mkrivos:
**Description**

If database reconnect occurs when DB is in transaction, session state will never restore to good state.

Try this testcase:

void PostgreSQLTest::testTransactionWithReconnect()
{
    if (!_pSession) fail ("Test not available.");

    try
    {
        _pSession->begin();
        *_pSession << "CREATE TABLE Person (LastName VARCHAR(30), FirstName VARCHAR(30), Address VARCHAR(30), Age INTEGER)", now;
        _pSession->reconnect();
        _pSession->commit();
    }
    catch (Poco::Exception& e)
    {
        _pSession->rollback();
        std::cout << e.displayText() << std::endl;
    }

    try
    {
        _pSession->begin();
        *_pSession << "CREATE TABLE Person3 (LastName VARCHAR(30), FirstName VARCHAR(30), Address VARCHAR(30), Age INTEGER)", now;
        _pSession->commit();
    }
    catch (Poco::Exception& e)
    {
        _pSession->rollback();
        std::cout << e.displayText() << std::endl;
    }
}

**Expected behavior**
It is expected that DB session after reconnect is ready for operations.

**Logs**
!!!FAILURES!!!
Runs: 1   Failures: 0   Errors: 1

There was 1 error: 
 1: CppUnit::TestCaller<PostgreSQLTest>.testTransactionWithReconnect
    "Poco::Data::PostgreSQL::StatementException: PostgreSQL: [PostgreSQL]: DEALLOCATE statement failed: ERROR:  prepared statement "p1759b758p9160p11f0p9a85p7cf17e54aed3" does not exist
"
    in "<unknown>", line -1

WARNING:  there is no transaction in progress
WARNING:  there is no transaction in progress

Process finished with exit code 1

**Please add relevant environment information:**
 - OS Linux
 - Poco 1.14.2
 - Latest PostgreSQL

**Our fix**

void SessionHandle::connect(const std::string& aConnectionString)
{
	Poco::FastMutex::ScopedLock mutexLocker(_sessionMutex);

        // begin change
	_inTransaction = false;
	_preparedStatementsToBeDeallocated.clear();
        // end change

	if (isConnectedNoLock())
	{
		throw ConnectionFailedException("Already Connected");
	}
	else if (_pConnection)
	{
		// free bad connection
		PQfinish(_pConnection);
		_pConnection = 0;
	}

[...]

PR #5027, opened 2025-09-14
fix(Data/PostgreSQL): PostgreSQL reconnect() doesn't restore session …

mkrivos:
…state correctly (#5025)


mkrivos:
FYI: Bug is triggered when server is restarted under heavy load (and client is in the transaction).

aleks-f (approved):

PR #5029, opened 2025-09-17
Eliminate extra copies in JSON::Object

austin-beer:
I didn't create an associated issue for this. I just discovered this minor problem while reviewing the Poco code. The identical code in JSON::Array sets `_modified` to `false`, so I think it should do the same in JSON::Object.


austin-beer:
The builds failed due to a null pointer exception in one of the unit tests. I don't have time to look into why, but if someone else wants to take a look, please feel free to!

Issue #5070, opened 2025-12-05
CMake does not produce proper binary names


aleks-f:
Eg. debug testrunner binaries are produced without the 'd' postfix in the name

PR #5071, opened 2025-12-05
Fixes #5069, #5070, #5075, #5076: Missing DLL reporting, VS project removal, CI updates

### LINKED ISSUE #5069: Missing DLLs not reported on Windows
When a DLL is not found, API error gives the name of the dependent module, but not of the dependency. This makes it hard to determine what exactly is the missing DLL.

Name(s) of the missing DLLs in the error message.

**Describe alternatives you've considered**
There seems to be no way to get this information from the API error. The only doable way seems to be `MapAndLoad` the library, examine its dependencies and return the list of the missing ones.

**Additional context**
Been bitten by it too many times. Time to end the suffering.

aleks-f:


Foundation/src/SharedLibrary_HPUX.cpp
matejk, line 1: Why does HPUX have its own (duplicated) implementation? Isn't that Unix?

aleks-f: it is, but from the ancient times when there were many specifics in various vendors versions. I'm not sure what state it its posix-compliance today, last time I touched that OS was over 30 years ago. in any case, for us it is not important unless someone shows up willing to maintain it or pay for update/maintenance, I consider it a dead-end implementation

matejk: No other observations from my side.

PR #5089, opened 2025-12-15
fix(Net): Add POCO_HAS_UNIX_SOCKET guards to fix build without Unix sockets #5064

### LINKED ISSUE #5064: Net: Build fails when POCO_HAS_UNIX_SOCKET is not defined
With version 1.14.2
On Windows
When POCO_HAS_UNIX_SOCKET is usually not defined
Poco Net fails to build

In 'void HTTPSession::connect(const SocketAddress& address)'
line    'if (address.family() != SocketAddress::UNIX_LOCAL)'
Build fail because UNIX_LOCAL isn't defined and not protected by #if defined(POCO_HAS_UNIX_SOCKET)

In void SocketAddress::init(const std::string& hostAndPort)
line    'newLocal(hostAndPort);'
Build fail because newLocal methods are defined only if POCO_HAS_UNIX_SOCKET is defined, instead of guard the content of the methods.

matejk:
## Summary

When `POCO_HAS_UNIX_SOCKET` is not defined (e.g., older Windows SDKs lacking `afunix.h`, or when `POCO_NET_NO_UNIX_SOCKET` is explicitly defined), the build failed because:

- `HTTPSession::connect()` referenced `SocketAddress::UNIX_LOCAL`
- `SocketAddress::init()` called `newLocal()` which is only declared when `POCO_HAS_UNIX_SOCKET` is defined
- `WebSocketImpl` constructor referenced `SocketAddress::UNIX_LOCAL`

## Changes

1. **Net/src/HTTPSession.cpp** - Added `#if defined(POCO_HAS_UNIX_SOCKET)` guard around `UNIX_LOCAL` check
2. **Net/src/SocketAddress.cpp** - Added guard around `isUnixLocal()` and `newLocal()` call
3. **Net/src/WebSocketImpl.cpp** - Added guard around `UNIX_LOCAL` check
4. **Net/include/Poco/Net/SocketDefs.h** - Use `POCO_HAVE_CPP17_COMPILER` macro instead of direct `__cplusplus >= 201703L` check for consistency
5. **Data/ODBC/include/Poco/Data/ODBC/ODBC.h** - Same consistency improvement
6. **Foundation/include/Poco/Platform_POSIX.h** - Simplified thread-safe local static init check since POCO requires C++17 (which guarantees this since C++11)

Closes #5064

### DIFF
--- a/Data/ODBC/include/Poco/Data/ODBC/ODBC.h
+++ b/Data/ODBC/include/Poco/Data/ODBC/ODBC.h
@@ -54,7 +54,7 @@
 
 #include "Poco/Data/ODBC/Unicode.h"
 
-#if (__cplusplus >= 201703L)
+#if POCO_HAVE_CPP17_COMPILER
 	#if __has_include(<msodbcsql.h>)
 		#include <msodbcsql.h>
 		#define POCO_DATA_ODBC_HAVE_SQL_SERVER_EXT
--- a/Foundation/include/Poco/Platform_POSIX.h
+++ b/Foundation/include/Poco/Platform_POSIX.h
@@ -33,16 +33,6 @@
 
 
 //
-// Thread-safety of local static initialization
-//
-#if __cplusplus >= 201103L || __GNUC__ >= 4 || defined(__clang__)
-	#ifndef POCO_LOCAL_STATIC_INIT_IS_THREADSAFE
-		#define POCO_LOCAL_STATIC_INIT_IS_THREADSAFE 1
-	#endif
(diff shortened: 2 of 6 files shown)


matejk:
**Updated:** Changed item 6 - instead of simplifying the `POCO_LOCAL_STATIC_INIT_IS_THREADSAFE` macro check, the entire macro has been **removed** as it is dead code (defined but never used anywhere in the codebase). Thread-safe local static initialization is guaranteed since C++11.

PR #5117, opened 2025-12-17
fix(Foundation): Reject trailing garbage in DateTimeParser (#5030)

### LINKED ISSUE #5030: poco date:time parser does not work as expected
Hello all,

I'm using version 13.3.0,
and have found a few issues that seem to address some aspects of this ( 1258,1477,363,569 ) , but they all seem so be prior to 1.13.3....

I seem to be unable to parse a time value (HH:MM).
specifically, a string that is not in the requested format, does not throw SyntaxException for Poco::DateTimeParser::parse  or 'false' for Poco::DateTimeParser::tryParse  as is described in the documentation )

I've attached a file with some code that shows this.
[dateParse.cpp.txt](https://github.com/user-attachments/files/22422318/dateParse.cpp.txt)

can anyone tell me if i'm missed something ,
Or if this is a bug ?

matejk:
## Summary

Fixes #5030 - DateTimeParser now properly rejects input strings with unconsumed trailing characters.

## Problem

`DateTimeParser::parse()` and `tryParse()` were accepting invalid input strings that didn't match the format:

```cpp
// These incorrectly returned true / didn't throw:
tryParse("%H:%M", "12345", ...)  // parsed as 12:34, ignored trailing '5'
tryParse("????", "12345", ...)   // no format specifiers, ignored all input
```

## Solution

After consuming all format specifiers, the parser now:
1. Skips trailing whitespace in the input
2. Throws `SyntaxException` if any non-whitespace characters remain

The fix maintains backwards compatibility by keeping the lenient literal character matching (format literals that don't match input are skipped), which existing code relies on.

## Changes

- `DateTimeParser.cpp`: Added trailing garbage detection after parsing loop
- `DateTimeParserTest.cpp`: Added test cases from issue #5030

## Test Results

All existing tests pass, plus the new issue-specific tests:
- `"%H:%M"` with `"13:13"` → success ✓
- `"%H:%M"` with `"xxx"` → failure ✓  
- `"%H:%M"` with `"12345"` → failure (trailing garbage) ✓
- `"????"` with `"12345"` → failure (invalid format) ✓

### DIFF
--- a/Data/SQLite/testsuite/src/SQLiteTest.cpp
+++ b/Data/SQLite/testsuite/src/SQLiteTest.cpp
@@ -2601,7 +2601,7 @@ void SQLiteTest::testSQLChannel()
 		"ThreadId INTEGER,"
 		"Priority INTEGER,"
 		"Text VARCHAR,"
-		"DateTime DATE)", now;
+		"DateTime DATETIME)", now;
 
 	tmp << "DROP TABLE IF EXISTS T_POCO_LOG_ARCHIVE", now;
 	tmp << "CREATE TABLE T_POCO_LOG_ARCHIVE (Source VARCHAR,"
@@ -2611,7 +2611,7 @@ void SQLiteTest::testSQLChannel()
 		"ThreadId INTEGER,"
 		"Priority INTEGER,"
 		"Text VARCHAR,"
-		"DateTime DATE)", now;
+		"DateTime DATETIME)", now;
 
 	AutoPtr<SQLChannel> pChannel = new SQLChannel(Poco::Data::SQLite::Connector::KEY, "dummy.db", "TestSQLChannel");
 	Stopwatch sw; sw.start();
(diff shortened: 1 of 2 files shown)


matejk:
## Test Fix Details

The test failures (`testSQLChannel` and `testSQLLogger`) were caused by a schema mismatch, not by the DateTimeParser fix itself.

### Root Cause

The tests created tables with `DateTime DATE` column type, but SQLChannel stores full ISO8601 timestamps (e.g., `"2025-12-18T07:39:36Z"`) via the Binder.

When extracting from a `DATE` column:
- SQLite Extractor uses `SQLITE_DATE_FORMAT` = `"%Y-%m-%d"` (date-only)
- But the stored value is full ISO8601 format `"%Y-%m-%dT%H:%M:%S%z"`
- **Before this PR**: parser silently ignored trailing `"T07:39:36Z"`
- **After this PR**: parser correctly rejects it as garbage

### Fix

Changed column type from `DATE` to `DATETIME` in:
- `SQLiteTest::testSQLChannel()` (3 tables)
- `SQLiteTest::testSQLLogger()` (1 table)
- `DBLogger` sample (1 table)

This ensures the column type matches what's being stored. `DATETIME` columns use auto-detection parsing which correctly handles full ISO8601 format.

The DateTimeParser fix exposed a pre-existing schema bug in the tests.

PR #5130, opened 2025-12-20
fix: Resolve compiler warnings and fix potential bugs

matejk:
## Summary

Comprehensive compiler warning fixes across GCC, Clang (macOS), and MSVC builds.

| Metric | Count |
|--------|-------|
| Files Modified | 80 |
| Lines Changed | +3622 / -2730 |
| Bug Fixes | 5 critical + 2 high + 8 potential |
| Warnings Resolved | ~100+ |

### 1. RegularExpression.cpp - PCRE2 Unset Match Detection
Check for unset capture groups was ALWAYS FALSE (`ovec[0] < 0` on unsigned type).

### 2. ProcessRunner.h - Windows Invalid PID Value
`INVALID_PID = -1` is wrong on Windows where PIDs are unsigned and 0 indicates invalid.

### 3. ProcessRunner.cpp - PID Type Mismatch
PID file parsing used `int` instead of platform-specific `PID` type.

### 4. IPAddressImpl.cpp - Missing Base Class Initialization
Copy constructors didn't explicitly initialize `RefCountedObject` base class.

### 5. X509Certificate.cpp - Variable Shadowing
Inner loop iterator shadowed outer loop iterator.

### 6. NumericString.h `safeMultiply` - Overflow Detection
Overflow detection comparisons could give wrong results with mixed signed/unsigned types.

### 7. CppUnit TestCase.h - Test Assertions
`assertEquals` could give incorrect results comparing signed vs unsigned values.

## Other Changes

- **Third-party update**: Tessil/ordered-map v1.2.0 (fixes `-Wdeprecated-copy`)
- **ODBC null constants**: Added `constexpr` null handle constants to avoid `-Wzero-as-null-pointer-constant`
- **Var.h**: Use demangled type names in exception messages
- **API docs**: Added `POCO_DEPRECATED` and migration docs for Observer class
- **Zip module**: Fixed enum arithmetic for MSVC C++20 compatibility

## Compiler Coverage

| Compiler | Platform | Warnings Addressed |
|----------|----------|-------------------|
| GCC | Linux | Type-limits, zero-as-null, fallthrough, base-init, sign-compare |
| Clang | macOS | Sign-compare, deprecated-copy |
| MSVC | Windows | C4244, C4245, C4456, C5054, C4702, C4018 |

## Test plan

- [ ] CI passes on all platforms (Linux, macOS, Windows)
- [ ] No regressions in existing tests

### DIFF
--- a/MongoDB/include/Poco/MongoDB/Document.h
+++ b/MongoDB/include/Poco/MongoDB/Document.h
@@ -125,22 +125,21 @@ public:
 		{
 			throw NotFoundException(name);
 		}
-		else
+
+		if (ElementTraits<T>::TypeId == element->type())
 		{
-			if (ElementTraits<T>::TypeId == element->type())
+			auto* concrete = dynamic_cast<ConcreteElement<T>* >(element.get());
+			if (concrete != nullptr)
 			{
-				auto* concrete = dynamic_cast<ConcreteElement<T>* >(element.get());
-				if (concrete != nullptr)
-				{
-					return concrete->value();
-				}
+				return concrete->value();
 			}
-			throw BadCastException("Invalid type mismatch!");
 		}
+		throw BadCastException("Invalid type mismatch!");
+
 	}
 
 	template<typename T>
-	const T& get(const std::string& name, const T& def) const
+	T get(const std::string& name, const T& def) const
 		/// Returns the element with the given name and tries to convert
 		/// it to the template type. When the element is not found, or
 		/// has the wrong type, the def argument will be returned.
--- a/Net/include/Poco/Net/HTTPMessage.h
+++ b/Net/include/Poco/Net/HTTPMessage.h
@@ -82,7 +82,7 @@ public:
 		/// The value should be either IDENTITY_TRANSFER_CODING
 		/// or CHUNKED_TRANSFER_CODING.
 
-	const std::string& getTransferEncoding() const;
+	std::string getTransferEncoding() const;
 		/// Returns the transfer encoding used for this
 		/// message.
 		///
@@ -108,7 +108,7 @@ public:
 	void setContentType(const MediaType& mediaType);
 		/// Sets the content type for this message.
 
-	const std::string& getContentType() const;
+	std::string getContentType() const;
 		/// Returns the content type for this message.
 		///
 		/// If no Content-Type header is present,
--- a/Net/include/Poco/Net/NameValueCollection.h
+++ b/Net/include/Poco/Net/NameValueCollection.h
@@ -79,7 +79,7 @@ public:
 		///
 		/// Throws a NotFoundException if the name-value pair does not exist.
 
-	const std::string& get(const std::string& name, const std::string& defaultValue) const;
+	std::string get(const std::string& name, const std::string& defaultValue) const;
 		/// Returns the value of the first name-value pair with the given name.
 		/// If no value with the given name has been found, the defaultValue is returned.
 
--- a/Net/src/HTTPMessage.cpp
+++ b/Net/src/HTTPMessage.cpp
@@ -139,7 +139,7 @@ void HTTPMessage::setTransferEncoding(const std::string& transferEncoding)
 }
 
 
-const std::string& HTTPMessage::getTransferEncoding() const
+std::string HTTPMessage::getTransferEncoding() const
 {
 	return get(TRANSFER_ENCODING, IDENTITY_TRANSFER_ENCODING);
 }
@@ -175,7 +175,7 @@ void HTTPMessage::setContentType(const MediaType& mediaType)
 }
 
 
-const std::string& HTTPMessage::getContentType() const
+std::string HTTPMessage::getContentType() const
 {
 	return get(CONTENT_TYPE, UNKNOWN_CONTENT_TYPE);
 }
--- a/Net/src/NameValueCollection.cpp
+++ b/Net/src/NameValueCollection.cpp
@@ -106,7 +106,7 @@ const std::string& NameValueCollection::get(const std::string& name) const
 }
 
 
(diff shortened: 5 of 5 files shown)


Data/ODBC/include/Poco/Data/ODBC/ODBC.h
```diff
@@ -78,6 +79,18 @@
...
+//
+// ODBC headers define SQL_NULL_HANDLE, SQL_NULL_HENV, SQL_NULL_HDBC, SQL_NULL_HSTMT
+// as literal 0, which triggers -Wzero-as-null-pointer-constant warnings in C++.
+// We define these as nullptr to use in place of the ODBC macros.
+//
+constexpr std::nullptr_t ODBC_NULL_HANDLE = nullptr;
+constexpr std::nullptr_t ODBC_NULL_HENV   = nullptr;
+constexpr std::nullptr_t ODBC_NULL_HDBC   = nullptr;
+constexpr std::nullptr_t ODBC_NULL_HSTMT  = nullptr;
+constexpr std::nullptr_t ODBC_NULL_HDESC  = nullptr;
```
aleks-f: let's call these POCO_ODBC_*

Data/ODBC/src/ConnectionHandle.cpp
```diff
@@ -30,8 +30,8 @@ const std::string ConnectionHandle::CANT_SET_ATTR_SQLSTATE = "HY011";
 
 
 ConnectionHandle::ConnectionHandle(const std::string& connectString, SQLULEN loginTimeout, SQLULEN timeout):
-	_pEnvironment(SQL_NULL_HENV),
-	_hdbc(SQL_NULL_HDBC),
+	_pEnvironment(nullptr),
```
aleks-f: should be `ODBC_NULL_HENV` for consistency

Data/ODBC/src/ConnectionHandle.cpp
```diff
@@ -59,25 +59,25 @@ void ConnectionHandle::alloc()
 	if (Utility::isError(SQLAllocHandle(SQL_HANDLE_DBC, _pEnvironment->handle(), &_hdbc)))
 	{
 		delete _pEnvironment;
-		_pEnvironment = SQL_NULL_HENV;
-		_hdbc = SQL_NULL_HDBC;
+		_pEnvironment = nullptr;
```
aleks-f, line 58: same as above

Data/ODBC/src/ConnectionHandle.cpp
```diff
@@ -59,25 +59,25 @@ void ConnectionHandle::alloc()
...
 		SQLFreeHandle(SQL_HANDLE_DBC, _hdbc);
-		_hdbc = SQL_NULL_HDBC;
+		_hdbc = ODBC_NULL_HDBC;
 	}
 
 	if (_pEnvironment)
 	{
 		delete _pEnvironment;
-		_pEnvironment = SQL_NULL_HENV;
+		_pEnvironment = nullptr;
```
aleks-f, line 76: same as above

Foundation/include/Poco/PriorityStrategy.h
```diff
@@ -141,6 +141,14 @@ class PriorityStrategy<void, TDelegate>
 	using Iterator = typename Delegates::iterator;
 
 public:
+	PriorityStrategy() = default;
+
+	PriorityStrategy(const PriorityStrategy& s):
+		_delegates(s._delegates)
```
aleks-f, line 143: should there also be move constructor here?

Foundation/include/Poco/ordered_hash.h
```diff
@@ -1,7 +1,7 @@
 /**
  * MIT License
  *
- * Copyright (c) 2017 Tessil
+ * Copyright (c) 2017 Thibaut Goetghebuer-Planchon <tessil@gmx.com>
```
aleks-f, line 4: these belong in dependencies

Data/ODBC/src/ConnectionHandle.cpp
matejk: This one is actually redundant because the member is already default-initialized.

Data/ODBC/src/ConnectionHandle.cpp
matejk: `Poco::Data::ODBC::EnvironmentHandle` is actually Poco, not ODBC class, so it is more appropriate to use nullptr.

Data/ODBC/src/ConnectionHandle.cpp
matejk: `Poco::Data::ODBC::EnvironmentHandle` is actually Poco, not ODBC class, so it is more appropriate to use nullptr.

Foundation/include/Poco/ordered_hash.h
matejk: Not possible (easily) because these headers must be installed with `Poco::Foundation`: `Poco::OrderedSet` and `Poco::OrderedMap`are just aliases:

```
template<class Key,
		class Hash = std::hash<Key>,
		class KeyEqual = std::equal_to<Key>,
		class Allocator = std::allocator<Key>,
		class ValueTypeContainer = std::deque<Key, Allocator>>
using OrderedSet = tsl::ordered_set<Key, Hash, KeyEqual, Allocator, ValueTypeContainer>;
```

Foundation/include/Poco/PriorityStrategy.h
matejk: Improved by setting all ctors as default.

Data/ODBC/include/Poco/Data/ODBC/ODBC.h
matejk: Done.

PR #5136, opened 2025-12-22
Typehandler simplification and other modernizations in Poco::Data

frwilckens:
### Poco::Data::TypeHandler Improvements

Makes use of @matejk's new implementation of Poco::Tuple to simplify Poco::Data::TypeHandler and adds specializations for std::tuple (of any size) and std::optional. The code size is reduced from 249 KB to 15.8 KB. It uses C++ variadic templates and if constexpr.

The template specialization for Poco::Tuple is based on the TypeHandler for std::tuple. To make this work, I added explicit conversion operators to Poco::Tuple that make it interoperable with std::tuple. I think this is harmless because it cannot be used to undermine a class invariant. If you still consider this undesirable, the TypeHandler for Poco::Tuple could be implemented analogous to that for std::tuple, at the cost of some code duplication.

The TypeHandlers for std::tuple and std::optional allow to use Poco::Data without having to use the Poco-specific classes Poco::Tuple and Poco::Nullable. The code is battle-tested in production. A few test cases are added.

### More consistent use of virtual/override

Overridden virtual methods in Poco::Data sometimes have an "override" annotation, sometimes are prefixed with "virtual", sometimes have no annotation at all. I tried to make it more consistent (there may be other instances I didn't capture). I think this makes it easier to understand the complex class hierarchy in Poco::Data and increases safety.

### Use explicitly deleted constructors and assignment operators

In many cases, the code uses the old idiom of declaring special methods as private, but not defining them. Of course, explicitly deleting them makes intent clearer and allows to remove declarations from derived classes. Again, I don't claim complete coverage.

### DIFF
--- a/Data/include/Poco/Data/AbstractSessionImpl.h
+++ b/Data/include/Poco/Data/AbstractSessionImpl.h
@@ -51,13 +51,9 @@ public:
 		/// The getter method for a property.
 
 	AbstractSessionImpl(const std::string& connectionString,
-		std::size_t timeout = LOGIN_TIMEOUT_DEFAULT): SessionImpl(connectionString, timeout),
-			_storage(std::string("deque")),
-			_bulk(false),
-			_emptyStringIsNull(false),
-			_forceEmptyString(false),
-			_sqlParse(false),
-			_autoCommit(true)
+		std::size_t timeout = LOGIN_TIMEOUT_DEFAULT):
+			SessionImpl(connectionString, timeout),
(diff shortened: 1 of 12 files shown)


frwilckens:
@aleks-f Yes, I can do that. Hopefully later today.

matejk (approved):
These changes are a valuable contribution to the Poco. I have no objections. Clang-tidy has identified further improvements that I have added to this PR.

PR #5151, opened 2026-01-05
5150 reactor sock remove

### LINKED ISSUE #5150: SocketReactor socket removal
SocketReactor currently only allows removing handlers, one at a time. But removing a handler does nothing to prevent data already in the dispatch queue at removal time to be passed to the just removed handler. When the removal is happening during normal client object lifetime, this can be easily dealt with in the client handler function. However, removing handlers on destruction inevitably creates a potential for UB - calling a member function of a destroyed object.

Currently, the only way to deal with this situation is not to allow handle registration to survive until destruction and be prepared to deal with some data after handler removal. Since there is a way to avoid the problem, this issue is not categorized as a bug. But it is an easy pitfall.

The solution:

- [x] add `SocketReactor::remove(const Socket&)`
  - removes socket from the PollSet
  - removes all handlers for the socket
- [x] add `Socket SocketNotifier::socket() const`
  (see below why this is needed)
- [x] add checks in:
  - [x] `SocketReactor::dispatch(const Socket&, SocketNotification*)`
    `if (!_pollSet.has(socket)) return;`
  - [x] `SocketReactor::dispatch(SocketNotification*)`
    `if (!_pollSet.has(delegate->socket())) continue;`
- [x] accompanying (Async)NotificationCenter modifications and optimizations

All this narrows the race window, but it does not completely close it. The race potential can never be 100% eliminated for reactor client that unregisters its handlers in destructor, but eg. an atomic mutual exclusion flag (declared first, so it is destroyed last) between destructor and handlers further narrows the possibility for UB.

aleks-f:
Closes #5150

### DIFF
--- a/Foundation/src/NotificationCenter.cpp
+++ b/Foundation/src/NotificationCenter.cpp
@@ -14,17 +14,13 @@
 
 #include "Poco/NotificationCenter.h"
 #include "Poco/Notification.h"
-#include "Poco/Observer.h"
-#include "Poco/AutoPtr.h"
 #include "Poco/RWLock.h"
 
 
 namespace Poco {
 
 
-NotificationCenter::NotificationCenter()
-{
-}
+NotificationCenter::NotificationCenter() = default;
 
 
 NotificationCenter::~NotificationCenter()
@@ -43,7 +39,7 @@ NotificationCenter::~NotificationCenter()
 void NotificationCenter::addObserver(const AbstractObserver& observer)
 {
 	RWLock::ScopedLock lock(_mutex);
-	_observers.push_back(observer.clone());
+	_observers.emplace_back(observer.clone());
 	_observers.back()->start();
 }
 
--- a/Data/ODBC/CMakeLists.txt
+++ b/Data/ODBC/CMakeLists.txt
@@ -54,7 +54,7 @@ find_file(
 if (_msodbc_h)
 	get_filename_component(MSODBC_DIR ${_msodbc_h} DIRECTORY CACHE)
 	message(STATUS "ODBC: Found msodbcsql.h in: ${MSODBC_DIR}")
-	target_include_directories(DataODBC PUBLIC "${MSODBC_DIR}")
+	target_include_directories(DataODBC SYSTEM PRIVATE "${MSODBC_DIR}")
 
 	if (ENABLE_DATA_SQL_SERVER_BIG_STRINGS)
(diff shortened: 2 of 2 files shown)


Foundation/src/NotificationCenter.cpp
```diff
@@ -134,10 +129,10 @@ int NotificationCenter::backlog() const
 {
 	int cnt = 0;
 
-	ScopedLockWithUnlock<Mutex> lock(_mutex);
+	_mutex.readLock();
```
matejk, line 128: Use scoped lock.

Foundation/src/ProcessRunner.cpp
```diff
@@ -168,29 +171,62 @@ void ProcessRunner::stop()
...
+				{
+					if (_sw.elapsedSeconds() > _timeout)
+					{
+						try
+						{
+							pidFile.remove(false);
+						}
+						catch (...)
+						{
+							// give up trying to remove stale pid file
```
matejk, line 217: The comment text doesn't seem to match the code.

Net/src/SocketReactor.cpp
```diff
@@ -261,6 +261,34 @@ void SocketReactor::removeEventHandler(const Socket& socket, const Poco::Abstrac
...
+void SocketReactor::remove(const Socket& socket)
+{
+	const SocketImpl* pImpl = socket.impl();
+	if (pImpl == nullptr) return;
+
+	// Remove from pollset first - prevents new events
+	// epoll_ctl may fail if socket FD is already in bad state, but we must
+	// still clean up handlers to prevent further dispatch attempts
+	try { _pollSet.remove(socket); }
+	catch (...) { }
```
matejk, line 273: Under which circumstances can this fail?

matejk (approved):
A few minor observations.

Net/src/SocketReactor.cpp
aleks-f: https://github.com/pocoproject/poco/blob/bd09152affd77b86adb8cdc0c4a17754dd00d30b/Net/src/PollSet.cpp#L120-L121

Foundation/src/ProcessRunner.cpp
aleks-f: why not? time has elapsed, last attempt has failed. maybe it could be on different line, or we can remove it altogether

PR #5153, opened 2026-01-05
Modernize Poco::Util and clean up some code duplications

frwilckens:
This PR modernizes the code in Poco::Util by using more C++ 11 and later idioms such as

  - type aliases
  - deleted and defaulted special member functions
  - override annotations
  - auto variables
  - range-for loops
  - default member initializers
  - constructor delegation
  - initializer lists
  - lambdas
  
It also does some refactoring to reduce duplicated code:

  - In classes `Application` and `ServerApplication`, arguments of form `int argc, char** argv` or `int argc, wchar_t** argv` are now immediately transformed into a canonical `ArgVec` before doing anything substantial. No longer
  is there duplicated code in (for example) `int ServerApplication::run(int argc, char** argv)`,
  `int ServerApplication::run(const ArgVec& args)`, and `int ServerApplication::run(int argc, wchar_t** argv)`.
  
  - Since `ConfigurationView` and `LocalConfigurationView` have almost identical code, they now derive from an intermediate class `AbstractConfigurationView` that provides the common methods.

`SystemConfiguration` now uses an explicit map of functions instead of long if-else chains. Since all tags 
are used only once, we no longer need to define them in a list of `static const std::string` variables.

`AbstractConfiguration::keys`, `Application::getApplicationPath` and `Application::getApplicationDirectory`
got overloads that return the requested information instead of filling a provided data structure. The old 
versions still exist for downward compatibility. (I assume they exist because it was expensive before C++11 to return a large data structure from a function.)

Overall the PR reduces the code size by about 500 lines and makes it more maintainable and easier to read.

### DIFF
--- a/Util/include/Poco/Util/Application.h
+++ b/Util/include/Poco/Util/Application.h
@@ -149,7 +149,7 @@ public:
 	}
 
 #if defined(_WIN32)
-	void init(int argc, wchar_t* argv[])
+	void init(int argc, wchar_t** argv)
 		/// Processes the application's command line arguments
 		/// and sets the application's properties (e.g.,
 		/// "application.path", "application.name", etc.).
@@ -417,10 +417,10 @@ protected:
 	~Application() override;
 		/// Destroys the Application and deletes all registered subsystems.
 
-	static ArgVec toArgs(int argc, char* argv[]);
+	static ArgVec toArgs(int argc, char** argv);
 
 #if defined(_WIN32)
-	static ArgVec toArgs(int argc, wchar_t* argv[]);
+	static ArgVec toArgs(int argc, wchar_t** argv);
 #endif
 
 private:
--- a/Util/src/Application.cpp
+++ b/Util/src/Application.cpp
@@ -83,7 +83,7 @@ Application::Application():
 }
 
 
-Application::Application(int argc, char* argv[]):
+Application::Application(int argc, char** argv):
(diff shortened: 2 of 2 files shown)


frwilckens:
I also added a template to  Poco::Util::Units:

`
template <typename V, typename U>
Value<V, U> abs(const Value<V, U>& a)
{
    return Value<V, U>(std::abs(a.get()));
}
`
It is analogous to the `sqrt` template and the provided arithmetic operators. Without it, even `std::abs(val)` does not compile when `val` is a value with unit.

Util/include/Poco/Util/AbstractConfiguration.h
```diff
@@ -376,6 +380,11 @@ class Util_API AbstractConfiguration: public Poco::RefCountedObject
 		/// Sets the property with the given key to the given value.
 		/// An already existing value for the key is overwritten.
 
+
+	Keys keys(const std::string& key = std::string()) const;
+		// Returns the names of all subkeys under the given key.
```
matejk: Should be /// instead of //

frwilckens: Ah sure, I fixed that.

matejk:
@frwilckens , is the PR ready to be merged?

frwilckens:
Yes, it' ready.

Issue #5152, opened 2026-01-05
Add attributes to symbols and `noexcept` specifiers


mikomikotaishi:
**Is your feature request related to a problem? Please describe.**
Discarding retrieved values is often unintended during compilation, however the compiler by default does not report these. This can be a point or even source of bugs. Furthermore, without directly checking documentation one does not always have access to a list of deprecated symbols. If we use attributes (which have been available since C++11), we can report most of these issues during compilation.

It would be preferable if unused values had to be manually suppressed (i.e. using `(void)`) to prevent these potential sources of accidents.

Furthermore, many getter and setter methods, which simply reassign values or return values themselves, are (mostly) obvious not to throw exceptions. However, they are not marked `noexcept` in many classes (for example `Poco::Crypto::{CipherKey, CipherKeyImpl}`). We could make use of compiler optimisations by taking advantage of `noexcept` in such locations.

**Describe the solution you'd like**
I would like:
* Mark getter methods and functions that are otherwise intended to retrieve values (rather than perform mutation) to be marked `[[nodiscard]]`, as a diagnostic for unused values. For example, any method that is named "`is`", "`get`", etc.
* Mark classes and structs that represent data (i.e. records) `[[nodiscard]]`, to prevent accidentally producing temporary objects that do not get used
* Mark functions that do not return (if applicable) `[[noreturn]]`
* ~~Mark obsolescent or deprecated functions or classes `[[deprecated]]`~~ Already done by `POCO_DEPRECATED()`.
* Mark methods (such as getters and setters) which are guaranteed not to throw exceptions, `noexcept`
* Wherever else possible, make use of attributes like `[[likely]]`, `[[unlikely]]`, `[[assume()]]`, `[[no_unique_address]]`, and `[[fallthrough]]`, to potentially use compiler optimisations, if the situation calls for them.

**Describe alternatives you've considered**
No such alternatives to this feature exist in the language.

**Additional context**
For attributes like `[[nodiscard]]` and `[[deprecated]]`, an optional parameter may be supplied specifying a reason that the symbol should not be ignored, or why the symbol is deprecated. If possible, we should include these.

mikomikotaishi:
Ah, perhaps this issue should have been marked "enhancement" rather than "feature". I'm not able to change the tags, so please change the tag for me

mikomikotaishi:
[...]

PR #5156, opened 2026-01-08
enh(all): Add `[[nodiscard]]` and `[[maybe_unused]]` attributes (addresses #5152)

mikomikotaishi:
Addresses part of #5152. This PR adds `[[nodiscard]]` attributes to many methods and types that would otherwise be important not to discard.

Unlike `POCO_DEPRECATED()` which can be disabled, I am not going to put `[[nodiscard]]` behind a macro as using deprecated symbols is intended while discarding a value is certainly a programming error.


mikomikotaishi:
~~I have also moved `POCO_UNUSED` to `[[maybe_unused]]`, seeing as it is a standard attribute in C++17, but I think this is probably not needed any more.~~ `POCO_UNUSED` has been removed entirely as `[[maybe_unused]]` is a standard attribute and is portably representable.

mikomikotaishi:
Would it be possible to get this merged before the 1.15 release? I understand if not, but it is nearly done. I have covered all methods across the project.

mikomikotaishi:
@matejk Do you think you could review this PR?

Foundation/include/Poco/Platform.h
```diff
@@ -285,11 +285,7 @@
 #endif
 
 
-#ifdef __GNUC__
-#define POCO_UNUSED __attribute__((unused))
-#else
-#define POCO_UNUSED
-#endif // __GNUC__
+#define POCO_UNUSED [[maybe_unused]]
```
matejk: IMO the macro is not needed. [[maybe_unused]] shall be used directly instead of it.

mikomikotaishi: OK, that makes sense, I'll remove it entirely.

matejk:
> Would it be possible to get this merged before the 1.15 release? I understand if not, but it is nearly done. I have covered all methods across the project.

It is quite a big change. I'll need some time for review.

matejk:
There are lots of changes. Instead of adding comments to the source code I provide some general observations.

matejk:
[[nodiscard]] on Enum Types
```
  enum [[nodiscard]] EventType { ... }
  enum [[nodiscard]] Priority { ... }
  enum [[nodiscard]] SelectMode { ... }
  enum [[nodiscard]] Storage { ... }
  enum [[nodiscard]] Allocation : unsigned char { ... }
```
This is unusual and mostly useless. [[nodiscard]] on an enum means functions returning that enum type should have their return value checked. But enum values are typically used as parameters, not return values. The attribute on the enum declaration itself doesn't do anything for enum values.

matejk:
[[nodiscard]] on Internal/Helper Classes
```
  class [[nodiscard]] FunctorRunnable: public Runnable { ... }
  class [[nodiscard]] ValueHolder { ... }
  class [[nodiscard]] Block { ... }
  class [[nodiscard]] FDCompare { ... }
```
These are internal classes that users don't directly construct or receive from functions. Adding [[nodiscard]] adds noise without benefit.

matejk:
[[nodiscard]] on Dereference Operators
```
  [[nodiscard]] C* operator -> ()
  [[nodiscard]] C& operator * ()
```
  This is wrong. These operators are used for accessing members:
```
  ptr->doSomething();  // This would warn! The return of -> is "discarded"
```
  The -> operator result IS used implicitly. This will likely cause spurious warnings.

matejk:
[[nodiscard]] on Nullable and Optional Classes
```
  class [[nodiscard]] Nullable { ... }
  class [[nodiscard]] Optional { ... }
```
Reconsider this. While it ensures you use the object, these types are commonly constructed just to be assigned to variables or passed around. The standard library's std::optional is not marked nodiscard.

mikomikotaishi:
I decided to put nodiscard on enums and class internal types because I wanted to ensure that data like this is not discarded.

I have gone ahead and applied your requested changes, nonetheless.

matejk:
@mikomikotaishi , you don't have to force push. It is harder to track changes between the commits when force pushed.

mikomikotaishi:
Noted

andrewauclair:
> While it ensures you use the object, these types are commonly constructed just to be assigned to variables or passed around.

Not sure what you mean by this, but `[[nodiscard]]` on the enum or class declaration only applies to functions returning that type by value.

mikomikotaishi:
Is there anything else I should address in this PR, pending reviews from the other reviewers?

matejk:
Build with enabled compiler warnings shall be inspected whether introduction of attributes introduced new warnings and then decide how to fix those warnings:

* change the code at the warning point
* revert introduction of the attribute

Two jobs with enabled warnings are built on github: macos-clang-cmake-warnings and macos-msvc-cmake-warnings.

mikomikotaishi:
Apologies, accidentally force-pushed instead of creating a new commit. (I'll force-push when making amendments from now on and use proper commits when adding something new.)

Anyway, I think I've done what I can to remove the warnings we were getting.

mikomikotaishi:
The tests failing now is completely beyond me as to why this is happening, I don't understand why there are timeouts happening.

mikomikotaishi:
Never mind, I see the problem now, I accidentally overwrote the `handled` variable.

mikomikotaishi:
@matejk Is there anything to do next?

matejk:
I am quite busy with other projects currently. I'll take a look ASAP. Thank you so much for your contribution so far.

mikomikotaishi:
Hi @matejk, is there still any time to eventually review and merge this?

matejk:
Hi, sorry for not answering. We are preparing release 1.15.0. Changes in this PR will be included in the next release.

Thank you for your patience.

mikomikotaishi:
So would it be possible to roll out this in a minor release (like 1.15.1)?

mikomikotaishi:
Now that 1.15.0 is rolled out, could we look towards integrating this for 1.15.1 perhaps?

mikomikotaishi:
@matejk also I think the website needs to be updated for 1.15.0, it still says 1.14.2

mikomikotaishi:
I'd also like to ask here (before making an issue that might just be not worth investigating): is there any reason why we primarily use unscoped C-style `enum`s rather than `enum class`? If possible I think it's worthwhile migrating to `enum class`, and making it an option during compilation (perhaps remaining with `enum` would be a better default choice until later on when it could be deprecated).

matejk:
> I'd also like to ask here (before making an issue that might just be not worth investigating): is there any reason why we primarily use unscoped C-style `enum`s rather than `enum class`? If possible I think it's worthwhile migrating to `enum class`, and making it an option during compilation (perhaps remaining with `enum` would be a better default choice until later on when it could be deprecated).

The reason is very simple: lots of Poco code was written before C++11 standard. Probably some of the enums could be converted to enum class. Probably not all.

matejk:
> Now that 1.15.0 is rolled out, could we look towards integrating this for 1.15.1 perhaps?

We'll include into 1.16.0 with the next round of C++ modernisation.

andrewauclair:
> I'd also like to ask here (before making an issue that might just be not worth investigating): is there any reason why we primarily use unscoped C-style `enum`s rather than `enum class`? If possible I think it's worthwhile migrating to `enum class`, and making it an option during compilation (perhaps remaining with `enum` would be a better default choice until later on when it could be deprecated).

As a user of Poco and one who has done these types of changes, I'd suggest a middle ground approach. It's more forgiving on user code to use `enum class` in combo with `using enum`. That gives proper type checking without a crap ton of errors and code changes for the scoping part.

mikomikotaishi:
> We'll include into 1.16.0 with the next round of C++ modernisation.

Could this still be merged for the time being or will we have to wait until work on 1.16.0 begins before merging this into the library?

> That gives proper type checking without a crap ton of errors and code changes for the scoping part.

I agree, I think the primary goal of moving towards `enum class` is at having the benefits of type safety. Perhaps it could be a compile option on whether to enable `using enum` (by default it should be on for backwards compatibility, but I think it's better to move towards eventually deprecating it with `using`).

matejk:
@mikomikotaishi, why did you close the PR?

mikomikotaishi:
It must have been an accident. I didn't intend to, but I will reopen it.

andrewauclair:
Why so many force pushes? What is changing? It would be much easier to see the changes if you left them in small commits. Then it can be squashed later.

mikomikotaishi:
@matejk Could we still look at merging this PR for the next release?

matejk:
> @matejk Could we still look at merging this PR for the next release?

Yes, this is planned for release 1.16.0. Thank you for your work.

JSON/include/Poco/JSON/Parser.h
```diff
@@ -102,24 +104,30 @@ class JSON_API Parser: private ParserImpl
 		/// may result in a stack overflow when parsing a (malicious)
 		/// JSON document.
 
+	[[nodiscard]]
 	std::size_t getDepth() const;
 		/// Returns the allowed JSON depth.
 
+	[[nodiscard]]
```
matejk: `parse()` is documented as discardable when a Handler is set (see the class comment above, `parser.parse(json); // ostr.str() == json`). Please drop the attribute here and on the `istream` overload, and revert the `[[maybe_unused]] Var _ = parser.parse(...)` workarounds in `Util/src/JSONConfiguration.cpp` and `JSON/testsuite/src/JSONTest.cpp`.

Foundation/include/Poco/TextEncoding.h
```diff
@@ -161,19 +166,22 @@ class Foundation_API TextEncoding
 		/// Removes the encoding with the given name from the table
 		/// of text encodings.
 
+	[[nodiscard]]
```
matejk: `global(Ptr)` returns the previous encoding; install-and-forget is the normal use. Please drop the attribute and revert the `TextEncodingTest.cpp` changes.

NetSSL_Win/include/Poco/Net/SecureStreamSocketImpl.h
```diff
@@ -203,14 +211,17 @@ class NetSSL_Win_API SecureStreamSocketImpl: public StreamSocketImpl
...
 	bool sessionWasReused();
 		/// Returns true iff a reused session was negotiated during
 		/// the handshake.
 
 	// SocketImpl
 	virtual void setBlocking(bool flag) override;
+	[[nodiscard]]
 	virtual bool getBlocking() const override;
 	virtual void setRawOption(int level, int option, const void* value, poco_socklen_t length) override;
+	[[nodiscard]]
```
matejk: `[[nodiscard]]` on a `void` function; please remove.

dependencies/tessil/include/Poco/ordered_hash.h
```diff
@@ -112,6 +112,7 @@ struct is_vector<T,
 
 // Only available in C++17, we need to be compatible with C++11
 template <class T>
+[[nodiscard]]
```
matejk: Please revert all changes under `dependencies/`; this is vendored third-party code and is updated from upstream as a whole.

Foundation/include/Poco/Platform.h
```diff
@@ -289,13 +289,6 @@
 #endif
 
 
-#ifdef __GNUC__
-#define POCO_UNUSED __attribute__((unused))
```
matejk, line 293: Please keep `POCO_UNUSED`; it is public API used by downstream code. A comment that `[[maybe_unused]]` is preferred is fine.

Net/include/Poco/Net/SocketAcceptor.h
```diff
@@ -150,10 +150,11 @@ class SocketAcceptor
 	{
 		StreamSocket sock = _socket.acceptConnection();
 		_pReactor->wakeUp();
-		createServiceHandler(sock);
+		[[maybe_unused]] ServiceHandler* handler = createServiceHandler(sock);
```
matejk: The handler registers itself with the reactor; the original code discarded the pointer on purpose. If `createServiceHandler` keeps `[[nodiscard]]`, use `(void) createServiceHandler(sock);` (repo convention). Same for `ParallelSocketAcceptor.h`, `SocketConnector.h`, `Foundation/src/Random.cpp` (`next()` warm-up), `FileChannel.cpp`, `LocalDateTime.cpp`, `DirectoryIterator_*.cpp`, `Data/src/Statement.cpp`, `NetSSL_OpenSSL/src/SecureSocketImpl.cpp`, `Net/src/DialogSocket.cpp`, `JSON/src/ParserImpl.cpp`, `Data/MySQL/.../SessionImpl.h`, `Redis/.../Type.h`.

Data/include/Poco/Data/AbstractPreparator.h
```diff
@@ -67,7 +67,7 @@ class Data_API AbstractPreparator
 	virtual ~AbstractPreparator();
 		/// Destroys the AbstractPreparator.
 
-	virtual void prepare(std::size_t pos, const Poco::Int8&) = 0;
+	virtual void prepare(std::size_t pos, [[maybe_unused]] const Poco::Int8& val) = 0;
```
matejk: `[[maybe_unused]]` on parameters of a pure-virtual declaration has no effect. Please remove here and on the other declarations in this PR; annotate only definitions with an unused parameter.

Data/DataTest/src/SQLExecutor.cpp
```diff
@@ -376,17 +376,17 @@ void SQLExecutor::sessionPool(const std::string& connector, const std::string& c
 	{
 		sp.setFeature("autoBind"s, true);
 		fail("SessionPool must throw on setFeature after the first session was created.", __LINE__, __FILE__);
-	}	catch(const Poco::InvalidAccessException&) {}
+	}	catch([[maybe_unused]] const Poco::InvalidAccessException& e) {}
```
matejk: Please remove the exception name instead (`catch (const Poco::InvalidAccessException&) {}`); applies to all annotated catch clauses in the PR.

XML/include/Poco/DOM/AttrMap.h
```diff
@@ -33,14 +33,21 @@ class XML_API AttrMap: public NamedNodeMap
 	// returned by Element::attributes()
 {
 public:
+	[[nodiscard]]
 	Node* getNamedItem(const XMLString& name) const;
+	[[nodiscard]]
```
matejk: `setNamedItem`/`setNamedItemNS` are mutators whose return value is routinely discarded; `NamedNodeMap.h` is (correctly) left unmarked. Please drop here too.

matejk (commented):
Thanks for the continued work on this. The current head has no class/enum-level attributes and leaves the chaining and side-effect APIs (DOM mutators, socket send/receive, `Statement::execute`, `sendRequest`, `tryWait`, ...) alone, so the earlier review points are addressed. I went through the whole diff against a scan of the tree; general points here, specifics inline.

**Process**
- Please rebase on `main` (no merge commits) so the history is reviewable.
- CI failures on the last run are GitHub infrastructure errors ("Failed to resolve action download info"), not code.

**Scope / style**
- Put the attribute inline (`[[nodiscard]] T f() const;`), matching the ~160 existing uses in the tree (MongoDB, NumericString.h, VarHolder.h, HTTPClientSession.h, MailMessage.h, AbstractConfiguration.h). Own-line placement is only used by this PR.
- Do not modify `dependencies/` (vendored third-party code, updated from upstream as a whole).
- Keep the `POCO_UNUSED` macro; it is public API used by downstream code. Not using it anymore in POCO itself is fine.

**`[[maybe_unused]]`**
- For intentionally discarded results in library code prefer `(void) f();` (repo convention, e.g. `Net/src/SocketImpl.cpp:1270`, and this PR already uses it in 13 places) over `[[maybe_unused]] T x = f();`. Where a library call site has to be patched to silence a newly added `[[nodiscard]]`, that is usually a sign the function should not carry it.
- Unused catch parameters: remove the name (`catch (const E&)`) instead of annotating (~116 sites).
- Do not annotate parameters on declarations (no effect with `-Wno-unused-parameter`, and never on pure-virtual declarations); annotate only definitions with an unused parameter.

**Missing `[[nodiscard]]` (highest value)**
The selection rule seems to have been "const member or static/factory". The functions where `[[nodiscard]]` actually catches bugs are mostly not marked:
- the `try*` bool family: `Mutex/FastMutex/RWLock/NamedMutex tryLock/tryReadLock/tryWriteLock`, `Thread::tryJoin`, `MPSCQueue/SPSCQueue tryPush/tryPop`, `NumberParser::tryParse*`, `DateTimeParser/Path/UUID/ULID/IPAddress::tryParse`, `File::createFile/createDirectory`, `Pipe::readBytes/writeBytes`, `Process::tryWait`, `AtomicFlag::set`, `Manifest::insert`, `TaskManager::start`;
- `NotificationQueue::dequeueNotification/waitDequeueNotification` (+ Priority/Timed variants) returning an owning `Notification*`;
- class-level on non-copyable RAII guards: `ScopedLock`, `ScopedLockWithUnlock`, `ScopedUnlock`, `ScopedRWLock`, `NDCScope`, `Data::Transaction`.

These need ~30 call-site fixes (FileTest.cpp, fuzzing harnesses, Benchmark/, `Net/src/RemoteSyslogListener.cpp:322`), so it is fine to leave them for a follow-up PR; just noting that the current selection misses them.

Net/include/Poco/Net/HTTPChunkedStream.h
```diff
@@ -77,6 +78,7 @@ class Net_API HTTPChunkedInputStream: public HTTPChunkedIOS, public std::istream
 	HTTPChunkedInputStream(HTTPSession& session, MessageHeader* pTrailer = nullptr);
 	~HTTPChunkedInputStream();
 
+	[[nodiscard]]
```
matejk: `operator new` cannot be discarded through a new-expression; the attribute is noise here. Please remove (also `HTTPFixedLengthStream`, `HTTPHeaderStream`, `HTTPStream`).

Foundation/include/Poco/NumberParser.h
```diff
@@ -41,6 +41,7 @@ class Foundation_API NumberParser
 	static const unsigned short NUM_BASE_DEC = 10;
 	static const unsigned short NUM_BASE_HEX = 0x10;
 
+	[[nodiscard]]
```
matejk: `parse()` is marked but `tryParse()` right below is not. The `bool` result of the `try*` functions is where `[[nodiscard]]` catches real bugs (see the summary comment); either add them here or in the follow-up.

Foundation/include/Poco/NotificationQueue.h
```diff
@@ -114,9 +114,11 @@ class Foundation_API NotificationQueue
 	void wakeUpAll();
 		/// Wakes up all threads that wait for a notification.
 
+	[[nodiscard]]
```
matejk: In this header the functions that matter are `dequeueNotification()` / `waitDequeueNotification()` -- they return an owning `Notification*` and discarding leaks. Same for `PriorityNotificationQueue` and `TimedNotificationQueue`.

Data/include/Poco/Data/Transaction.h
```diff
@@ -95,13 +95,16 @@ class Data_API Transaction
 	void setIsolation(Poco::UInt32 ti);
 		/// Sets the transaction isolation level.
 
+	[[nodiscard]]
```
matejk: Consider `class [[nodiscard]] Transaction` instead: it is an RAII guard, and a discarded temporary (`Transaction(session);`) silently rolls back. All in-tree constructions are named locals, so no new warnings.

matejk:
Hi @mikomikotaishi. It took longer than I hoped to review the changes.

NetSSL_Win/include/Poco/Net/SecureStreamSocketImpl.h
mikomikotaishi: Fixed

Foundation/include/Poco/Platform.h
mikomikotaishi: Fixed

XML/include/Poco/DOM/AttrMap.h
mikomikotaishi: Fixed

Data/DataTest/src/SQLExecutor.cpp
mikomikotaishi: Fixed

Net/include/Poco/Net/HTTPChunkedStream.h
mikomikotaishi: Fixed

Data/include/Poco/Data/AbstractPreparator.h
mikomikotaishi: Fixed

dependencies/tessil/include/Poco/ordered_hash.h
mikomikotaishi: Fixed

Foundation/include/Poco/TextEncoding.h
mikomikotaishi: Fixed

JSON/include/Poco/JSON/Parser.h
mikomikotaishi: Fixed

Net/include/Poco/Net/SocketAcceptor.h
mikomikotaishi: Fixed

Data/include/Poco/Data/Transaction.h
mikomikotaishi: I'll leave it on the accessors, but I've added `[[nodiscard]]` to the class itself.

mikomikotaishi:
OK; I think this should hopefully cover most of the requested changes.

Another thing; could we consider moving `for (;;)` to `while (true)`? The former seems to be a vestige of (ancient) C, while the latter reads much more clearly in both intent and natural language, while both being exactly the same thing.

matejk:
I agree with for --> while change. Also take a look at failed jobs.

mikomikotaishi:
> I agree with for --> while change.

OK, I put it up at #5456.

matejk:
@mikomikotaishi , CodeQL reported some problems. Fix those then this PR can be merged.

mikomikotaishi:
Many of these are reporting unused variables, but the variables are themselves marked `[[maybe_unused]]`; are you sure you want me to make changes?

mikomikotaishi:
@matejk Could you clarify which findings you think should be fixed? I deliberately chose not to address the `args` being unused, or the `DateTime` findings, because they weren't related to things I touched nor did I think it was reasonable to try to address these. I did however, address the findings in `VarTest`, because it seemed like the only thing to be tested was triggering a throw, which indeed didn't need a temporary variable to do.

matejk:
I have rebased this branch onto current `main` (452b6190) and added one commit rewriting `DateTimeTest::testSetYearDay()` to use the existing API (`DateTime(year, 1, 1)` + `Timespan`, `dayOfYear()`, `isLeapYear()`), replacing the `#if 0` scaffolding and the unused `X`/`r2` locals. This resolves the three remaining CodeQL `Unused local variable` alerts; the `args` alerts in `Format.h`/`AutoPtr.h` are CodeQL false positives (the parameter packs are used via pack expansion) and will be dismissed separately.

Note: your fork's `main` was force-pushed. Before continuing work, run `git fetch` and `git reset --hard origin/main` in your clone.

PR #5160, opened 2026-01-10
fix(NotificationCenter): lock order inversion deadlock #5159

### LINKED ISSUE #5159: Lock-order-inversion deadlock in NotificationCenter
#### Summary

`NotificationCenter` methods call observer methods (`start()`, `disable()`) while holding the internal mutex, creating a lock-order-inversion with Observer's mutex.

#### Deadlock Scenario

**Thread A** (eg. `SocketRreactor::run()`):
1. `Observer::notify()` holds Observer mutex (M0)
2. Handler calls `removeEventHandler()` ([documented as safe](https://github.com/pocoproject/poco/blob/17864c39a8d9bf9c7ecb924f90d61f3e56a1f918/Net/include/Poco/Net/SocketReactor.h#L116-L118))
3. → `hasObserver()` waits for NotificationCenter mutex (M1)

**Thread B** (shutdown):
1. `removeObserver()` holds NotificationCenter mutex (M1)
2. → `disable()` waits for Observer mutex (M0)

**Result:** Deadlock (M0 → M1, M1 → M0)

#### Affected Methods

- `addObserver()` - calls `start()` under lock
- `removeObserver()` - calls `disable()` under lock
- `clear()` - calls `disable()` under lock

#### Fix

Release lock before calling observer methods:

```cpp
void NotificationCenter::addObserver(const AbstractObserver& observer)
{
	AbstractObserverPtr pObserver = observer.clone();
	{
		RWLock::ScopedWriteLock lock(_mutex);
		_observers.push_back(pObserver);
	}
	// Start outside the lock to avoid deadlock with Observer mutex
	pObserver->start();
}
```

aleks-f:
# Branch Summary: 5159-nc-deadlock

This branch fixes several thread-safety issues and deadlocks reported in issues #5159, #5161, and #5162.

Closes #5159, #5161, and #5162.

## Issues Addressed

- **#5159**: NotificationCenter lock order inversion deadlock
- **#5161**: PollSet and Pipe thread-safety issues (data races detected by TSan)
- **#5162**: General thread synchronization improvements

### IOLock (`Foundation/include/Poco/IOLock.h`, `Foundation/src/IOLock.cpp`)

A new lock-free synchronization primitive for coordinating I/O operations with close/shutdown:

- `IOLock` - tracks whether an I/O operation is in progress and whether close has been requested
- `ScopedIOLock` - RAII wrapper that enters on construction and leaves on destruction
- Uses `std::atomic<bool>` for `_closed` and `_inProgress` flags
- Provides `markClosed()`, `wait()`, `tryWait()`, and `close()` methods

### ScopedSocketReactor (`Net/include/Poco/Net/SocketReactor.h`)

A new RAII wrapper utility class that starts a SocketReactor in a background thread and automatically stops/joins on destruction:

```cpp
ScopedSocketReactor reactor;
reactor->addEventHandler(socket, observer);
// reactor runs in background thread
// automatically stopped and joined when reactor goes out of scope
```

### NotificationCenter (#5159)
- Fixed lock order inversion deadlock in observer notification

### Pipe (#5161)
- Added `IOLock` synchronization for thread-safe read/write operations
- Fixed close ordering: **write end must be closed before read end** to send EOF to blocked readers
- Fixed race condition: `wait()` for I/O completion now happens **before** closing file descriptors
- Affects both POSIX and Windows implementations

### PollSet (#5161)
- Added `IOLock` to coordinate `poll()` with destructor
- Prevents use-after-free when `poll()` is running during destruction
- Exit early on poll error when closed
- Affects epoll, poll, and select implementations

### New Tests
- `PipeTest` - comprehensive test suite including `testCloseRace` for concurrent close/read
- `testConcurrentHandlerRemoval` in SocketReactorTest - tests concurrent handler removal

### CI Enhancements
- Added sanitized builds (ASan, TSan, UBSan) to CI workflow

## Files Changed

| Component | Files |
|-----------|-------|
| IOLock | `IOLock.h`, `IOLock.cpp` (new) |
| Pipe | `PipeImpl_POSIX.h/cpp`, `PipeImpl_WIN32.h/cpp`, `Pipe.cpp` |
| PollSet | `PollSet.cpp` |
| SocketReactor | `SocketReactor.h`, `SocketReactor.cpp` |
| Tests | `PipeTest.cpp/h`, `SocketReactorTest.cpp/h`, `NotificationCenterTest.cpp/h` |
| CI | `.github/workflows/ci.yml` |

### Pipe Close Sequence (Before)
```cpp
void PipeImpl::closeRead() {
    if (_readfd != -1) close(fd);      // Race: close while read() in progress
}
```

### Pipe Close Sequence (After)
```cpp
void PipeImpl::closeRead() {
    _readLock.markClosed();
    _readLock.wait();              // Wait for any in-progress read to finish
    int fd = _readfd.exchange(-1);
    if (fd != -1) close(fd);       // Now safe to close
}
```

### Destructor Close Order (Before)
```cpp
~PipeImpl() {
    closeRead();   // Wrong: blocks readers waiting for data
    closeWrite();
}
```

### Destructor Close Order (After)
```cpp
~PipeImpl() {
    closeWrite();  // Correct: sends EOF to unblock readers
    closeRead();
}
```

### DIFF
--- a/Foundation/include/Poco/IOLock.h
+++ b/Foundation/include/Poco/IOLock.h
@@ -7,7 +7,7 @@
 //
 // Definition of the IOLock class.
 //
-// Copyright (c) 2004-2008, Applied Informatics Software Engineering GmbH.,
+// Copyright (c) 2026, Applied Informatics Software Engineering GmbH.,
 // Aleph ONE Software Engineering LLC,
 // and Contributors.
 //
@@ -84,10 +84,11 @@ public:
 
 	void wait();
 		/// Waits for any in-progress IO operation to complete.
+		/// Uses atomic wait if available, otherwise busy-waits.
 		/// Call after markClosed() and after closing the resource.
 
 	bool tryWait(int milliseconds);
-		/// Waits for any in-progress IO operation to complete,
+		/// Busy-waits for any in-progress IO operation to complete,
 		/// up to the specified timeout.
 		/// Returns true if no IO is in progress, false if timed out.
 
@@ -181,12 +182,6 @@ private:
 //
 // inlines
 //
-
-inline void IOLock::leave()
-{
-	_inProgress.store(false);
-}
-
 
 inline bool IOLock::isClosed() const
 {
--- a/Foundation/src/IOLock.cpp
+++ b/Foundation/src/IOLock.cpp
@@ -5,7 +5,7 @@
 // Package: Threading
 // Module:  IOLock
 //
-// Copyright (c) 2004-2008, Applied Informatics Software Engineering GmbH.,
+// Copyright (c) 2026, Applied Informatics Software Engineering GmbH.,
 // Aleph ONE Software Engineering LLC,
 // and Contributors.
 //
@@ -26,10 +26,22 @@ bool IOLock::enter()
 	_inProgress.store(true);
 	if (_closed.load())
 	{
-		_inProgress.store(false);
+		_inProgress.store(false, std::memory_order_release);
+#ifdef POCO_HAVE_ATOMIC_WAIT
+		_inProgress.notify_one();
+#endif
 		return false;
 	}
 	return true;
+}
+
+
+void IOLock::leave()
+{
+	_inProgress.store(false, std::memory_order_release);
+#ifdef POCO_HAVE_ATOMIC_WAIT
+	_inProgress.notify_one();
+#endif
 }
 
 
@@ -48,15 +60,22 @@ void IOLock::markClosed()
 
 void IOLock::wait()
 {
+#ifdef POCO_HAVE_ATOMIC_WAIT
+	auto old = _inProgress.load(std::memory_order_acquire);
+	// spurious wakeups possible, hence the loop
+	while (_inProgress.load(std::memory_order_acquire))
+		_inProgress.wait(old, std::memory_order_relaxed);
+#else
 	while (_inProgress.load())
 		std::this_thread::sleep_for(std::chrono::microseconds(10));
+#endif // POCO_HAVE_ATOMIC_WAIT
 }
 
 
 bool IOLock::tryWait(int milliseconds)
 {
 	auto deadline = std::chrono::steady_clock::now() + std::chrono::milliseconds(milliseconds);
-	while (_inProgress.load())
+	while (_inProgress.load(std::memory_order_acquire))
 	{
 		if (std::chrono::steady_clock::now() >= deadline)
 			return false;
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -226,4 +226,14 @@
 	#define POCO_ENABLE_FASTLOGGER
 #endif
 
+#if defined(__APPLE__) && __has_include(<Availability.h>)
+	#include <Availability.h>
+	#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 110000
+		 #define POCO_HAVE_ATOMIC_WAIT 1
+	#endif
+#elif defined(__cpp_lib_atomic_wait)
+	// Standard feature test macro (C++20)
+	#define POCO_HAVE_ATOMIC_WAIT 1
+#endif
+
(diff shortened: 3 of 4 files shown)


Foundation/include/Poco/IOLock.h
```diff
@@ -0,0 +1,200 @@
+//
+// IOLock.h
+//
+// Library: Foundation
+// Package: Threading
+// Module:  IOLock
+//
+// Definition of the IOLock class.
+//
+// Copyright (c) 2004-2008, Applied Informatics Software Engineering GmbH.,
```
matejk: Wrong  (C) year. Blast from the past! :)

Foundation/src/NotificationCenter.cpp
```diff
@@ -38,24 +38,34 @@ NotificationCenter::~NotificationCenter()
 
 void NotificationCenter::addObserver(const AbstractObserver& observer)
 {
-	RWLock::ScopedWriteLock lock(_mutex);
-	_observers.emplace_back(observer.clone());
-	_observers.back()->start();
+	AbstractObserverPtr pObserver = observer.clone();
```
matejk, line 41: Why does the observer have to be cloned before lock?

Foundation/include/Poco/IOLock.h
```diff
@@ -0,0 +1,200 @@
...
+
+
+#include "Poco/Foundation.h"
+#include <atomic>
+
+
+namespace Poco {
+
+
+class Foundation_API IOLock
```
matejk, line 29: Perhaps `IOResourcePin` or `IOResourceHold` to distinguish from mutex-like locking?

matejk (commented):
Some comments in the code. In addition: Would IOLock also deserve its own unit tests?

Foundation/src/IOLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+void IOLock::markClosed()
+{
+	_closed.store(true);
+}
+
+
+void IOLock::wait()
+{
+	while (_inProgress.load())
+		std::this_thread::sleep_for(std::chrono::microseconds(10));
```
matejk, line 70: Use std::condition_variable instead of busy loop?

Foundation/src/IOLock.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <chrono>
+#include <thread>
+
+
+namespace Poco {
+
+
+bool IOLock::enter()
+{
+	_inProgress.store(true);
```
matejk, line 26: Would it make sense to add detection if the second thread called enter()? debug assertion or exception?

aleks-f: IOLock explicitly does not support multi-threaded I/O. It can be done, but I don't think the complications it brings are worth it, the class does one thing only - syncing between destruction in one thread and I/O in another. Whoever is doing l/O from multiple threads, I wish him good luck.

EDIT: to be more specific, whoever is doing I/O on a single channel, or polls, from multiple threads

Foundation/src/IOLock.cpp
aleks-f: 2ec60b0c1
let's see how that does in ci ...

Foundation/src/NotificationCenter.cpp
aleks-f: because this is precisely where the deadlock potential resides:
```cpp
pObserver->start();
```

Foundation/src/IOLock.cpp
matejk: What I meant was to detect multiple threads using IOLock and prevent that (or log or throw and exception).

aleks-f: I guess one thing that could be done is to keep thread references and add some sanity checks, but I would do it for debug build only and I really don't have time to do it. As far as I'm concerned, we made a knife, if someone gets cut, thats' on them.

matejk:
@aleks-f , ready to merge?

Issue #5087, opened 2025-12-14
Undefined reference for pocoNetworkInitializer with MinGW


MayaPosch:
**Describe the bug**
While recompiling some existing projects that use Poco ([libnymphcast ](https://github.com/MayaPosch/libnymphcast)library and C binding) I got the error `undefined reference to `pocoNetworkInitializer'`. I did not change anything about the code or Makefiles. I am compiling this in MSYS2 (UCRT) with current GCC and Poco libraries as available ([1.14.2-1](https://packages.msys2.org/packages/mingw-w64-x86_64-poco)).

As a workaround I have added explicit network init in the C binding example and compile it with `-DPOCO_NO_AUTOMATIC_LIB_INIT`, in which case the project builds, links and runs fine.

**To Reproduce**
Running `make` on the main Makefile for the library (static and dynamic) or the C binding's example results in the linker failing with the above error.

**Expected behavior**
I would expect the code to link as it used to do a few Poco versions ago.

**Logs**
Feel free to suggest more information I can add.

**Screenshots**
Console output screenshots can be added as needed.

**Please add relevant environment information:**
 - OS: Windows 10, MSYS2 UCRT environment.
 - POCO Version: as above.

matejk:
Dear @MayaPosch , thank you very much for this bug report.

Unfortunately none of the maintainers is actively using MinGW. If you can provide a fix with a pull request, we would review and apply it.

dubstab96:
I have the same issue after i upgraded my MSYS2 (MinGW64) to the latest version of poco (mingw64/mingw-w64-x86_64-poco 1.14.2-1) available in pacman. MSYS2 does not allow revert to older versions (since they delete old packages) and this is a problem for my workflow.

hmartinez82:
Btw. I just noticed this while compiling the code with GCC in UCRT64:
```
D:/M_P/mingw-w64-poco/src/poco-1.14.2-all/Net/include/Poco/Net/Net.h:106:33: warning: attributes ignored on elaborated-type-specifier that is not a forward declaration [-Wattributes]
  106 | extern "C" const struct Net_API NetworkInitializer pocoNetworkInitializer;
```

hmartinez82:
@matejk  https://github.com/pocoproject/poco/pull/5167

PR #5167, opened 2026-01-19
net: Fix position of dll attribute in pocoNetworkInitializer to also support MinGW-GCC/Clang

hmartinez82:
Fixes https://github.com/pocoproject/poco/issues/5087


matejk:
@hmartinez82, please modify the PR for branch `main`. `devel` is not used anymore.

Issue #5168, opened 2026-01-19
Lock-Order-Inversion Deadlock in NObserver/Observer


aleks-f:
`NObserver::handle()` and `Observer::notify()` hold mutex while calling user callbacks. If the callback unregistered observers (e.g., during disconnect), it would call `NObserver::disable()` which tried to acquire the same mutex - causing a deadlock cycle.

Fix: Since _pObject is std::atomic<C*>, the mutex can be removed and `handle()`/`notify()` can read atomically to call handlers without holding any lock.

PR #5169, opened 2026-01-19
fix(NObserver): Lock-Order-Inversion Deadlock #5168

aleks-f:


aleks-f:
> The trade-off is that observers should not be modified (via assignment operators) concurrently with notification delivery, which is a reasonable assumption for typical observer pattern usage.

This is true, but does not apply to this change because it was always like that - destructor did not acquire mutex.

Foundation/include/Poco/AsyncObserver.h
```diff
@@ -83,7 +83,7 @@ class AsyncObserver: public NObserver<C, N>
 			setHandler(observer._handler);
 			setMatcher(observer._matcher);
 			_started = false;
-			_done =false;
+			_done = true;
```
matejk, line 86: Should variable _done be named _running or _active?

PR #5180, opened 2026-02-03
Fix ProcessRunner::start() to handle Windows executables without .exe extension

matejk:
## Summary
- Added public `File::getExecutablePath()` method that returns the executable path (appends `.exe` on Windows if no extension present)
- Fixed `ProcessRunner::start()` to use `getExecutablePath()` before checking file existence
- Added unit tests for `getExecutablePath()`

## Test plan
- [x] All FileTest tests pass (22 tests)
- [x] Verified `getExecutablePath()` returns correct paths on Unix (unchanged) and Windows (`.exe` appended)

### DIFF
--- a/Foundation/testsuite/src/FileTest.cpp
+++ b/Foundation/testsuite/src/FileTest.cpp
@@ -17,7 +17,6 @@
 #include "Poco/Exception.h"
 #include "Poco/Thread.h"
 #include <fstream>
-#include <iostream>
 #include <set>
 #if defined(POCO_OS_FAMILY_WINDOWS)
 #include <Windows.h>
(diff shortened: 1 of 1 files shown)


aleks-f:
@matejk appending `.exe` to a file that does not have it creates a file name that does not exist. `PATHEXT` environment variable already does the same thing as this commit, but it still won't work if the file does not actually have one of the extensions listed in `PATHEXT`. 

A file named `File` does not mean a file named `File.exe` exists, but the reverse works by virtue of the `PATHEXT`(if present) - you can execute `File.exe` by just issuing `File` command 

```
c:\Temp>dir Test
 Volume in drive C has no label.
 Volume Serial Number is CAC2-980E

 Directory of c:\Temp

02/02/2026  03:31 PM         5,148,160 Test
               1 File(s)      5,148,160 bytes
               0 Dir(s)  35,160,547,328 bytes free

c:\Temp>.\Test
'.\Test' is not recognized as an internal or external command,
operable program or batch file.

c:\Temp>.\Test.exe
'.\Test.exe' is not recognized as an internal or external command,
operable program or batch file.

c:\Temp>move Test Test.exe
        1 file(s) moved.

c:\Temp>echo %PATHEXT%
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC

c:\Temp>.\Test
Hello world
```

matejk:
You are right. I'll prepare proper implementation.

PR #5186, opened 2026-02-06
enh(File): proper implementation of File::getExecutablePath

matejk:
## Summary

- **`getExecutablePathImpl()`** is now a full resolver: on Unix it searches cwd + PATH via `findInPath()`, then verifies executability using euid/egid stat checks; on Windows it uses `SearchPathW` (two-call pattern for paths > MAX_PATH) with PATHEXT iteration. Paths containing a directory separator are resolved to absolute via `Path::makeAbsolute()`.
- **`canExecuteImpl()`** on Unix now uses euid/egid stat-based permission checks (consistent with `canReadImpl`/`canWriteImpl`) with an `S_ISREG` guard. Windows checks extension against PATHEXT. VxWorks treats any regular file as executable (no permission bits).
- **`absolutePath()`** no longer searches PATH — it only resolves relative paths against cwd
- **`existsAnywhere()`** reimplemented via shared `findInPath()` helper, fixed to not search PATH for absolute paths
- **`findInPath()`** extracted into `File.cpp` — always returns absolute paths via `Path::makeAbsolute()`, with cached PATH tokenization
- **Performance optimizations**:
  - PATH caching: tokenized once and cached with change detection (thread-safe with mutex)
  - PATHEXT caching (Windows): tokenized once at startup (thread-safe by design)
  - Combined stat checks: Unix/VxWorks now check existence and executability in a single stat() call
- `ProcessRunner::start()` updated to use `getExecutablePath()` / `canExecute()`
- VxWorks `File_VX.cpp` synced with Unix/Windows implementations
- Tests updated: added non-executable file test, directory test, relative path with separator test, non-existent relative path test
- Documentation: Added note about Windows UNC path limitations with SearchPathW

Follows up on #5180.

## Performance Improvements

- **Reduced syscalls**: Unix/VxWorks now use 1 stat() call instead of 2 for paths with directory separators
- **Cached PATH parsing**: Avoids repeated `Environment::get("PATH")` and string tokenization (thread-safe)
- **Cached PATHEXT parsing** (Windows): One-time initialization eliminates repeated parsing (thread-safe)
- **Thread-safe**: All caching implementations are thread-safe for concurrent calls

## Bug Fixes

- Fixed `existsAnywhere()` incorrectly searching PATH for absolute paths
- Eliminated TOCTOU race condition in `getExecutablePathImpl()` by removing redundant existence check

## Test plan

- [x] Build Foundation library and testsuite on macOS, Linux, Windows
- [x] Run `Foundation-testrunner` — FileTest and ProcessRunnerTest must pass
- [x] Verify `File("ls").getExecutablePath()` returns absolute path like `/usr/bin/ls` (Unix)
- [x] Verify `File("nonexistent").getExecutablePath()` returns empty string
- [x] Verify non-executable file returns empty from `getExecutablePath()`
- [x] Verify directory path returns empty from `getExecutablePath()`

### DIFF
--- a/Foundation/include/Poco/File.h
+++ b/Foundation/include/Poco/File.h
@@ -129,6 +129,12 @@ public:
 		///
 		/// Returns the fully resolved absolute path on success,
 		/// or an empty string if no executable was found.
+		///
+		/// Known limitation on Windows: If the PATH environment variable
+		/// contains UNC paths (e.g., \\server\share\bin), SearchPathW may
+		/// not find executables in those locations when searching for bare
+		/// filenames. This is a Windows API limitation, not a Poco limitation.
+		/// Workaround: Use the full UNC path including the executable name.
 
 	bool exists() const;
 		/// Returns true iff the file exists.
--- a/Foundation/src/File.cpp
+++ b/Foundation/src/File.cpp
@@ -17,6 +17,7 @@
 #include "Poco/DirectoryIterator.h"
 #include "Poco/Environment.h"
 #include "Poco/StringTokenizer.h"
+#include <mutex>
 
 
 namespace Poco {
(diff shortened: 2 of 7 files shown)


Foundation/include/Poco/File.h
```diff
@@ -106,32 +106,36 @@ class Foundation_API File: private FileImpl
...
+		/// On Windows, also tries appending each PATHEXT extension
+		/// (e.g. .EXE, .CMD, .BAT) using SearchPathW.
+		///
+		/// Returns the fully resolved absolute path on success,
+		/// or an empty string if no executable was found.
 
 	bool exists() const;
 		/// Returns true iff the file exists.
 
+	POCO_DEPRECATED("use getExecutablePath() instead")
```
aleks-f: I would not deprecate it. I think it has legitimate utility value.

matejk: Removed deprecation.

aleks-f (approved):

aleks-f:
@matejk see copilot suggestions, some make sense

matejk:
> @matejk see copilot suggestions, some make sense

They really do.

PR #5194, opened 2026-02-09
Fix release scripts and pocodoc

### LINKED ISSUE #5193: Release is broken
The `mkdoc` documentation generation pipeline (used by both `publish_release.yml` and `packages-qa.yml`) is completely non-functional. Running `mkdoc all` produces **zero documentation output**. Two independent bugs prevent any docs from being generated.

## Bug 1: `mkrelease` generates an immediately-failing Makefile (BLOCKER)

**File:** `release/script/mkrelease`, line 190
**Introduced in:** commit `7a23a039f` ("Reorganise external libs to separate subdirectory", Sep 2025)

The `mkrelease` script generates a Makefile that starts with:

```makefile
$(error Please build Poco with CMake)
```

This was added to direct end users toward CMake, but it also breaks internal tooling. The `mkdoc` script calls `mkrelease` to copy sources into a staging area, then runs `./configure && make` to build PocoDoc from source. The `$(error)` causes `make` to fail immediately, so **PocoDoc is never compiled** and the subsequent `PocoDoc` invocation fails with `PocoDoc: not found`.

### Reproduction

```bash
export POCO_BASE=$(pwd)
export PATH=$POCO_BASE/release/script:$PATH
export LD_LIBRARY_PATH=$POCO_BASE/stage/tools/lib/Linux/$(uname -m)
mkdoc all
```

Output includes:
```
Makefile:7: *** Please build Poco with CMake.  Stop.
...
PocoDoc: not found
```

### Suggested fix

Either:
- Remove or comment out the `$(error)` line in `mkrelease` (line 190), or
- Update `mkdoc` to build PocoDoc using CMake instead of the legacy Makefile build system

The rest of issues is discovered after commenting the blocker and running documentation generation.

## Bug 2: PocoDoc crashes (SIGABRT) on `FileStreamRWLock.h`

**Crash location:** while parsing `Foundation/include/Poco/FileStreamRWLock.h`

Even after working around Bug 1 (commenting out the `$(error)` and building PocoDoc), PocoDoc aborts during documentation generation. It processes 270 of 863 headers (31%) before crashing.

The crash occurs while parsing `FileStreamRWLock.h`. The likely cause is CppParser failing to handle one or more of:
- `std::atomic_bool _locked = false;` (in-class member initializer, line 79)
- `struct flock _flock;` (POSIX struct from `<fcntl.h>`, expanded by the preprocessor from the included `FileStreamRWLock_POSIX.h`)
- `= delete` / `= default` specifiers on constructors/destructors

Output:
```
Application: [Information] Parsing .../Foundation/include/Poco/FileStreamRWLock.h
Aborted
```

### Suggested fix

Either:
- Fix CppParser to handle the problematic construct, or
- Add `FileStreamRWLock.h` to the exclude list in `PocoDoc/cfg/mkdoc-poco.xml`

## Additional issues: Parse errors on deprecated headers

**Files:**
- `Foundation/include/Poco/DynamicAny.h`
- `Foundation/include/Poco/DynamicAnyHolder.h`
- `Foundation/include/Poco/DynamicStruct.h`

These deprecated compatibility headers use anonymous namespaces (`namespace { ... }`) to emit deprecation warnings. PocoDoc's CppParser cannot parse anonymous namespaces:

```
Application: [Error] Syntax error: Expected: namespace name: .../DynamicAny.h(23)
```

### Suggested fix

Add these three files to the `<exclude>` list in `PocoDoc/cfg/mkdoc-poco.xml`:
```xml
<exclude>
    ...existing excludes...
    ${PocoBuild}/Foundation/include/Poco/DynamicAny.h
    ${PocoBuild}/Foundation/include/Poco/DynamicAnyHolder.h
    ${PocoBuild}/Foundation/include/Poco/DynamicStruct.h
</exclude>
```

## Minor issues

- **Hardcoded x86_64 in workflow:** `.github/workflows/publish_release.yml` line 36 sets `LD_LIBRARY_PATH=.../Linux/x86_64`. Should use `$(uname -m)` for ARM64 runner compatibility. Same issue in `packages-qa.yml` line 25.
- **Missing `buildwin.cmd`/`buildwin.ps1`:** `mkrelease` tries to copy these Windows files (lines 119-120) but they don't exist on non-Windows checkouts, producing `cp: cannot stat` warnings.

## Impact

- `publish_release.yml` `mkdoc` job: **broken since Sep 2025** (commit `7a23a039f`)
- `packages-qa.yml` `mkdoc` job: **broken since Sep 2025** (same commit)
- No API reference documentation can be generated for releases

matejk:
Resolves #5193 — the `mkdoc` documentation pipeline was completely broken because `mkrelease` generated a Makefile starting with `$(error Please build Poco with CMake)`, preventing PocoDoc from being built.

This PR fixes the release scripts by migrating them from the legacy `./configure && make` build to CMake, and fixes multiple CppParser bugs that blocked documentation generation.

### 1. Migrate `mkrelease` from legacy Make to CMake (`release/script/mkrelease`)

The generated Makefile contained `$(error Please build Poco with CMake)` at line 190, making it immediately fail. The remaining ~240 lines of Makefile generation (component targets, dependency resolution, install rules) were dead code that could never execute.

- Replaced ~240 lines of legacy Makefile generation with a thin (~35 line) CMake-forwarding Makefile that provides `all`, `test`, `install`, `clean` targets delegating to `cmake`
- Replaced `build_vs160.cmd`/`build_vs170.cmd` (which called the nonexistent `buildwin` script) with a single `build_cmake.cmd`
- Removed copying of legacy build system artifacts (`configure`, `build/config`, `build/rules`, `build/script`, `buildwin.cmd`, `buildwin.ps1`) — these only served the defunct Make build; `buildwin.cmd`/`.ps1` didn't even exist in the repo anymore
- Kept `libversion` which CMake reads (`CMakeLists.txt:9`)
- Removed VS project file copies (`.sln`, `.vcxproj`, `.vcxproj.filters`, `.progen`) from CppUnit section
- Added `set -e` so scripts fail on error instead of silently continuing
- Added `|| true` to CppUnit copy commands that may legitimately find no matching files (with `set -e`)

### 2. Migrate `mkdoc` from `./configure && make` to CMake (`release/script/mkdoc`)

The `mkdoc` script was the only internal consumer of the legacy build system. It staged sources with `mkrelease`, then ran `./configure && make` to build PocoDoc — which failed due to the `$(error)` directive.

- Replaced `./configure && make` with CMake: `cmake -DPOCO_MINIMAL_BUILD=ON -DENABLE_POCODOC=ON` + `cmake --build --target PocoDoc`
- `POCO_MINIMAL_BUILD=ON` skips `find_package` for OpenSSL, MySQL, PostgreSQL, ODBC (not needed for PocoDoc)
- `ENABLE_POCODOC=ON` auto-enables all required dependencies (CppParser, Util, XML, Data, DataSQLite)
- `--target PocoDoc` builds only PocoDoc and its dependencies, not everything
- Removed Cygwin-specific handling (DLL rebasing, PATH manipulation) and Perforce references
- Removed `libversion` check (not needed by mkdoc; CMake handles it internally)
- Updated `PATH` and `LD_LIBRARY_PATH` to point to CMake output directories (`cmake-build/bin`, `cmake-build/lib`)

### 3. Fix CppParser crash on `= delete`/`= default` (`CppParser/src/Parser.cpp`)

Even after fixing the build, PocoDoc crashed with SIGABRT while parsing `FileStreamRWLock.h`.

**Root cause:** In `Parser::parseFunc()`, when the parser encountered `= delete` or `= default` after a function declaration, it dereferenced `pFunc` without a null check (lines 727, 729). `pFunc` is null when the function name contains `:` (scope-qualified names from preprocessed inline definitions like `ScopedFStreamReadRWLock::~ScopedFStreamReadRWLock() = default`). Every other `pFunc` method call in the same function properly guards with `if (pFunc)`.

**Secondary bug:** `makePureVirtual()` was called unconditionally after `= default`/`= delete`/`= 0`, but only the `= 0` case should mark a function as pure virtual.

**Fix:** Added `if (pFunc)` null guards for `makeDefault()` and `makeDelete()`, and moved `makePureVirtual()` into an `else` branch so it only applies to the `= 0` (integer literal) case.

### 4. Fix CppParser infinite loop and improve C++ parsing (`CppParser/src/Parser.cpp`, `CppParser/src/CppToken.cpp`)

After the crash fix, PocoDoc hung indefinitely when parsing modern C++ headers (e.g., `MPSCQueue.h`).

#### 4a. Fix infinite loop in `parseFunc()` post-qualifier loop

**Root cause:** `parseVarFunc()` misinterpreted `alignas(64)` as a function call, then `parseFunc()` entered its post-qualifier loop. When the next token didn't match any known qualifier (`const`, `noexcept`, `override`, `final`, `=`), the loop had no `break` or token advance — causing an infinite loop.

**Fix:**
- Added a `while` loop in `parseVarFunc()` to consume `alignas(...)` and `decltype(...)` as declaration prefixes before function parsing
- Added `else break` as a safety net in `parseFunc()` post-qualifier loop for unrecognized tokens
- Added `continue` after the `const` handler to avoid fallthrough to the break
- Added `noexcept(expr)` handling — consumes the parenthesized expression after `noexcept`
- Added `parseParenthesized()` helper for balanced parenthesis consumption

#### 4b. Handle `static_assert` in all dispatch switches

Added `KW_STATIC_ASSERT` cases to `parseFile()`, `parseNameSpace()`, and `parseClassMembers()` — these call `parseStaticAssert()` which skips until `;`.

#### 4c. Handle anonymous namespaces (`namespace { ... }`)

Deprecated headers (`DynamicAny.h`, `DynamicAnyHolder.h`, `DynamicStruct.h`) use anonymous namespaces. `parseNameSpace()` required an identifier after `namespace`, causing a syntax error on `namespace {`. Added an `else if` branch to detect `{` directly after `namespace` and skip the block with `parseBlock()`.

#### 4d. Handle inline friend function definitions

Headers like `ordered_hash.h` define inline friend functions:
```cpp
friend bool operator!=(const ordered_iterator& lhs, const ordered_iterator& rhs) { return !(lhs == rhs); }
```
`parseFriend()` skipped tokens until `;`, but inline definitions end with `}`. The parser overshot into the next declaration. Fixed by also stopping at `{`, then calling `parseBlock()` to skip the function body.

#### 4e. Handle C++11 raw string literals (`R"(...)"`)

`ObjectId.h` contains `R"(")"` which the tokenizer split into identifier `R`, string `("`, operator `)`, and an unterminated `"` — causing "Unterminated character literal" errors. Added `finishRawString()` to `IdentifierToken` that detects when the identifier is a raw string prefix (`R`, `u8R`, `uR`, `LR`) followed by `"`, and consumes the entire raw string literal (including arbitrary delimiters like `R"---(...)---"`) as a single token using a sliding window match.

#### 4f. Handle C++14 string literal suffixes (`"str"s`)

Headers like `LabeledMetricImpl.h` use `"str"s` and `"str"sv`. After the closing `"`, the suffix was tokenized as a separate identifier, which could confuse the parser. Added suffix consumption (`s`, `sv`, `_`-prefixed user-defined literals) in `StringLiteralToken::finish()`.

#### 4g. Replace string literal comparisons with named constants

Replaced inline `"override"` and `"final"` string comparisons in `parseFunc()` with static `const std::string` constants (`ID_OVERRIDE`, `ID_FINAL`) for clarity and to avoid repeated temporary string construction.

### 5. Update CI workflows (`.github/workflows/packages-qa.yml`, `publish_release.yml`)

- Added `cmake ninja-build` to package install steps (needed to build PocoDoc)
- Removed hardcoded `LD_LIBRARY_PATH=$POCO_BASE/stage/tools/lib/Linux/x86_64` — the `mkdoc` script now sets this internally, and the old path was architecture-specific (would break on ARM64 runners)
- Added `set -e` to `mkrel` and `mkdocumentation` scripts

## Test plan

- [ ] `mkrelease -o /tmp/poco-test 1.15.0` produces a release archive with the CMake-forwarding Makefile, `build_cmake.cmd`, and no legacy build artifacts
- [ ] `mkdoc all` builds PocoDoc via CMake and generates documentation without crashing or hanging
- [ ] PocoDoc parses all headers with 0 CppParser syntax errors (verified locally on macOS)
- [ ] `packages-qa.yml` CI jobs pass (mkdoc, mkrelease, cmake build from archive)
- [ ] Release archives build correctly with CMake (validated by existing `linux-gcc-cmake-mkrelease*` and `windows-2022-msvc-cmake-mkrelease*` CI jobs)

### DIFF
--- a/CppParser/src/CppToken.cpp
+++ b/CppParser/src/CppToken.cpp
@@ -370,11 +370,11 @@ void IdentifierToken::finish(std::istream& istr)
 		next = istr.peek();
 	}
 
-	// C++11 raw string literals: R"delim(...)delim", also u8R"...", uR"...", LR"..."
+	// C++11 raw string literals: R"delim(...)delim", also u8R"...", uR"...", UR"...", LR"..."
 	// Without this handling, the tokenizer would split R"(")" into separate tokens:
 	//   identifier 'R', string literal '("', operator ')', and an unterminated '"',
 	// causing "Unterminated character literal" errors downstream.
-	if ((_value == "R" || _value == "u8R" || _value == "uR" || _value == "LR")
+	if ((_value == "R" || _value == "u8R" || _value == "uR" || _value == "UR" || _value == "LR")
 		&& istr.peek() == '"')
 	{
 		finishRawString(istr);
--- a/CppParser/testsuite/src/CppParserTest.cpp
+++ b/CppParser/testsuite/src/CppParserTest.cpp
@@ -14,6 +14,9 @@
 #include "Poco/CppParser/Utility.h"
 #include "Poco/CppParser/Symbol.h"
 #include "Poco/CppParser/CppToken.h"
+#include "Poco/CppParser/Parser.h"
+#include "Poco/CppParser/Struct.h"
+#include "Poco/CppParser/Function.h"
 #include <sstream>
 #include <iostream>
 
@@ -22,18 +25,21 @@ using namespace Poco::CppParser;
 
 
 std::string linker("cl");
-std::string options("/I \"C:\\Program Files\\Microsoft Visual Studio 8\\VC\\INCLUDE\", "
-				"/I \"C:\\Program Files\\Microsoft Visual Studio 8\\VC\\PlatformSDK\\include\", "
-				"/I \"p:\\poco\\Foundation\\include\", "
-				"/I \"p:\\poco\\XML\\include\", "
-				"/I \"p:\\poco\\Util\\include\", "
-				"/I \"p:\\poco\\Net\\include\", "
-				"/D \"WIN32\", "
-				"/D \"_DEBUG\", "
-				"/D \"_WINDOWS\", "
-				"/D \"_MBCS\", "
-				"/C, /P, /TP");
-std::string path("C:\\Program Files\\Microsoft Visual Studio 8\\Common7\\IDE;C:\\Program Files\\Microsoft Visual Studio 8\\VC\\BIN;C:\\Program Files\\Microsoft Visual Studio 8\\Common7\\Tools;C:\\Program Files\\Microsoft Visual Studio 8\\Common7\\Tools\\bin");
+std::string options(R"(/I "C:\Program Files\Microsoft Visual Studio 8\VC\INCLUDE", )"
+	R"(/I "C:\Program Files\Microsoft Visual Studio 8\VC\PlatformSDK\include", )"
+	R"(/I "p:\poco\Foundation\include", )"
+	R"(/I "p:\poco\XML\include", )"
+	R"(/I "p:\poco\Util\include", )"
+	R"(/I "p:\poco\Net\include", )"
+	R"(/D "WIN32", )"
+	R"(/D "_DEBUG", )"
+	R"(/D "_WINDOWS", )"
+	R"(/D "_MBCS", )"
+	"/C, /P, /TP");
+std::string path(R"(C:\Program Files\Microsoft Visual Studio 8\Common7\IDE;)"
+	R"(C:\Program Files\Microsoft Visual Studio 8\VC\BIN;)"
+	R"(C:\Program Files\Microsoft Visual Studio 8\Common7\Tools;)"
+	R"(C:\Program Files\Microsoft Visual Studio 8\Common7\Tools\bin)");
 
 
 CppParserTest::CppParserTest(const std::string& name): CppUnit::TestCase(name)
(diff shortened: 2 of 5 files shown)


matejk:
@aleks-f , @obiltschnig : scripts and CppParser fixed. PR ready for review and merge.

matejk:
## Addressed review comments

- **UR raw string prefix**: Added missing `UR` to the raw string prefix check in `IdentifierToken::finish()` (`CppToken.cpp`)
- **LD_LIBRARY_PATH guard**: `mkdoc` now guards against appending to an empty `LD_LIBRARY_PATH` (avoids stray colon)
- **build_cmake.cmd arg forwarding**: Uses `shift` + `%*` instead of `%2 %3 ... %9` for unlimited extra CMake options

## Unit tests added

8 new test methods in `CppParserTest` covering all recent CppParser fixes:

| Test | What it verifies |
|------|-----------------|
| `testTokenizerRawStrings` | Raw string literal tokenization with all prefixes (R, u8R, uR, UR, LR), custom delimiters, special characters, empty content |
| `testTokenizerStringSuffixes` | C++14/17 string literal suffixes (`s`, `sv`, user-defined `_mylit`) |
| `testParseStaticAssert` | `static_assert` is skipped without adding symbols |
| `testParseAlignas` | `alignas(N)` on member variables doesn't cause infinite loop |
| `testParseNoexceptExpr` | `noexcept(expr)` is parsed and `FN_NOEXCEPT` flag is set |
| `testParseDeleteDefault` | `= default` / `= delete` are recognized on constructors |
| `testParseAnonymousNamespace` | Anonymous namespaces are skipped |
| `testParseFriendInline` | Inline friend function bodies are consumed without breaking subsequent members |

All 18 tests pass (10 pre-existing + 8 new).

PR #5202, opened 2026-02-11
fix(cmake+msvc): Restore dllimport on extern template declarations (#5200)

### LINKED ISSUE #5200: 1.15.0: linker errors with MSVC link-time-code-generation
**Describe the bug**
When building our application, I am getting this linker error after upgrading to 1.15.0:
```
13:10:13  zstd_streambuf.obj : error LNK2005: "public: virtual int __cdecl Poco::BasicBufferedStreamBuf<char,struct std::char_traits<char>,class Poco::BufferAllocator<char> >::overflow(int)" (?overflow@?$BasicBufferedStreamBuf@DU?$char_traits@D@std@@V?$BufferAllocator@D@Poco@@@Poco@@UEAAHH@Z) already defined in libPocoFoundation.dll.if.lib(libPocoFoundation.dll)
13:10:13  zstd_streambuf.obj : error LNK2005: "public: virtual int __cdecl Poco::BasicBufferedStreamBuf<char,struct std::char_traits<char>,class Poco::BufferAllocator<char> >::underflow(void)" (?underflow@?$BasicBufferedStreamBuf@DU?$char_traits@D@std@@V?$BufferAllocator@D@Poco@@@Poco@@UEAAHXZ) already defined in libPocoFoundation.dll.if.lib(libPocoFoundation.dll)
13:10:13  bazel-out\x64_windows-opt\bin\projects\libs\shared\libShared.dll : fatal error LNK1169: one or more multiply defined symbols found
```

Weirdly enough, this only happens for release builds and only if LTCG is enabled.
We are using MSVC 14.44.35207 (VS2022.14).

I believe the root cause is that daf19e5223e2077660024ae5ca6727c6c0e9be3c removed the `dllimport` annotation on the `extern template class` declarations.
Note that MSVC's `dllexport` is not valid on such declarations, so omitting the `Foundation_API` macro is correct for MSVC when building Poco itself. (the `dllexport` annotation instead belongs on the template instance definition in the .cpp file)
But when the header is included into an application where `Foundation_API` expands to `dllimport` instead, this annotation is still necessary.

I believe the correct code for such an `extern template` should be:
```diff
-#if defined(POCO_OS_FAMILY_WINDOWS)
+#if defined(POCO_OS_FAMILY_WINDOWS) && defined(Foundation_EXPORTS)
 extern template class BasicBufferedStreamBuf<char, std::char_traits<char>>;
 #else
 extern template class Foundation_API BasicBufferedStreamBuf<char, std::char_traits<char>>;
 #endif
```
Applying this patch locally allows our build to succeed.

matejk:
## Summary

- Fix MSVC LNK2005 linker errors when consumers build with LTCG (Link-Time Code Generation) enabled, reported in #5200
- Change `#if defined(POCO_OS_FAMILY_WINDOWS)` to `#if defined(POCO_OS_FAMILY_WINDOWS) && defined(<Component>_EXPORTS)` in all `extern template` declarations (headers) and explicit template instantiations (`.cpp` files) across Foundation, JSON, and Data/ODBC
- Enable LTCG (`CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`) on the `windows-2025-msvc-cmake` CI job to catch this class of bug going forward

## Root cause

Commit daf19e5 stripped the API macro (e.g. `Foundation_API`) from `extern template class` declarations in headers on all Windows builds. This is correct when **building** the DLL (`dllexport` is invalid on `extern template` in MSVC), but wrong when **consuming** the DLL — the `dllimport` annotation is needed so the compiler knows the instantiation lives in the DLL. Without it, consumer TUs silently generate local instantiations, which cause LNK2005 duplicate symbol errors under LTCG.

The fix narrows the condition to only strip the macro when `<Component>_EXPORTS` is defined (i.e. when building the DLL itself).

## Test plan

- [x] macOS build passes (verified locally)
- [x] Windows shared-library build with LTCG passes (CI `windows-2025-msvc-cmake` job, now with `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`)
- [x] All existing test suites pass

Closes #5200

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -645,7 +645,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: >-
-          cmake -S. -Bcmake-build -DENABLE_NETSSL_WIN=ON -DENABLE_NETSSL=OFF
+          cmake -S. -Bcmake-build -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
+          -DENABLE_NETSSL_WIN=ON -DENABLE_NETSSL=OFF
           -DENABLE_CRYPTO=OFF -DENABLE_JWT=OFF -DENABLE_DATA=ON -DENABLE_DATA_ODBC=ON
           -DENABLE_DATA_MYSQL=OFF -DENABLE_DATA_POSTGRESQL=OFF -DENABLE_TESTS=ON
       - run: cmake --build cmake-build --config Release --parallel 4
--- a/Data/ODBC/include/Poco/Data/ODBC/Diagnostics.h
+++ b/Data/ODBC/include/Poco/Data/ODBC/Diagnostics.h
@@ -245,7 +245,7 @@ private:
 // explicit instantiation definition
 #ifndef POCO_DOC
 
-#if defined(POCO_OS_FAMILY_WINDOWS)
+#if defined(POCO_OS_FAMILY_WINDOWS) && defined(ODBC_EXPORTS)
 extern template class Diagnostics<SQLHENV, SQL_HANDLE_ENV>;
 extern template class Diagnostics<SQLHDBC, SQL_HANDLE_DBC>;
 extern template class Diagnostics<SQLHSTMT, SQL_HANDLE_STMT>;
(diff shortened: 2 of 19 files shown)


dgrunwald:
I confirm this PR solves the linker error I was seeing in #5200.

Issue #5201, opened 2026-02-11
ProcessRunner: add PROCESS_KILL_TREE option


aleks-f:
## Summary

On Windows, `TerminateProcess()` only kills the target process — it does not terminate child processes spawned by it. When `ProcessRunner::stop()` force-kills a process after timeout, any grandchild processes are orphaned and continue running. This can hold DLLs, ports, and file handles open indefinitely.

## Reproduction

1. Use `ProcessRunner` to launch a process that itself spawns child processes (e.g., a process supervisor that launches worker processes)
2. Call `ProcessRunner::stop()` — it sends termination signal, waits, then calls `TerminateProcess()` after timeout
3. The main process is killed, but its children survive as orphans
4. The orphaned processes hold resources (TCP ports, DLL locks, file handles) preventing subsequent launches or cleanup

## Root cause

`ProcessRunner::stop()` ultimately calls `TerminateProcess()` which only affects the single process identified by its handle. Windows has no built-in mechanism to propagate termination to child processes (unlike Unix process groups with `kill(-pgid, sig)`).

## Proposed fix

Add a `PROCESS_KILL_TREE` option flag that, when set, creates a Windows Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` and assigns the child process to it. When the Job Object handle is closed (either explicitly in `stop()` or implicitly when `ProcessRunner` is destroyed), the kernel automatically terminates all processes in the job — the child and all its descendants.

### API

```cpp
// New flag in Process.h (or ProcessRunner.h), next available bit:
static const int PROCESS_KILL_TREE = 0x10;

// Usage:
ProcessRunner pr("agent.exe", args,
    Process::PROCESS_CLOSE_STDOUT | Process::PROCESS_KILL_TREE);
```

### Implementation (Windows)

In `ProcessRunner::start()`, after `CreateProcess`:

```cpp
#if defined(POCO_OS_FAMILY_WINDOWS)
if (_options & PROCESS_KILL_TREE)
{
    _hJob = CreateJobObjectW(nullptr, nullptr);
    if (_hJob)
    {
        JOBOBJECT_EXTENDED_LIMIT_INFORMATION jeli = {};
        jeli.BasicLimitInformation.LimitFlags = JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
        SetInformationJobObject(_hJob, JobObjectExtendedLimitInformation, &jeli, sizeof(jeli));
[...]

PR #5203, opened 2026-02-11
5199 process termination

aleks-f:


Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +200,78 @@ void ProcessRunner::stop()
...
+#endif
+
+#if !defined(POCO_OS_FAMILY_WINDOWS)
+			if (_options & PROCESS_KILL_TREE)
 			{
-				if (_sw.elapsedSeconds() > _timeout)
+				// Wait for the entire process group to exit.
+				// setpgid(0, 0) was called in the child, so -pid
+				// addresses the whole group.
+				while (::kill(-pid, 0) == 0)
```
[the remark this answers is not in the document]
aleks-f: 9fb910319

Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +200,78 @@ void ProcessRunner::stop()
...
 				{
```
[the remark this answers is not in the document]
aleks-f: a2e0b7e57

Foundation/src/ProcessRunner.cpp
```diff
@@ -126,6 +132,26 @@ void ProcessRunner::run()
...
+				JOBOBJECT_EXTENDED_LIMIT_INFORMATION jeli = {};
+				jeli.BasicLimitInformation.LimitFlags = JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE;
+				SetInformationJobObject(static_cast<HANDLE>(_hJob),
+					JobObjectExtendedLimitInformation, &jeli, sizeof(jeli));
+				HANDLE hProc = OpenProcess(PROCESS_SET_QUOTA | PROCESS_TERMINATE,
+					FALSE, static_cast<DWORD>(pPH->id()));
+				if (hProc)
+				{
+					AssignProcessToJobObject(static_cast<HANDLE>(_hJob), hProc);
+					CloseHandle(hProc);
```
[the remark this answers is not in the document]
aleks-f: c6dea117d

Util/src/ServerApplication.cpp
```diff
@@ -270,8 +270,22 @@ bool ServerApplication::isService()
...
+	// but stdin is still inherited from the parent.
+	HANDLE handles[] = {
+		GetStdHandle(STD_OUTPUT_HANDLE),
+		GetStdHandle(STD_INPUT_HANDLE),
+		GetStdHandle(STD_ERROR_HANDLE)
+	};
+	for (HANDLE h : handles)
+	{
+		if (h != INVALID_HANDLE_VALUE && h != NULL)
+			return true;
```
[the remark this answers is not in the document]
aleks-f: 858ca0cc8

Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +200,78 @@ void ProcessRunner::stop()
...
 			}
 			_t.join();
 		}
 
+#if defined(POCO_OS_FAMILY_WINDOWS)
+		if (_hJob)
+		{
+			CloseHandle(static_cast<HANDLE>(_hJob));
+			_hJob = nullptr;
+		}
```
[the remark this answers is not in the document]
aleks-f: c6dea117da59c663bb56dc7b54332cb8ed16f4a2

matejk:
Related ticket #5201 is actually a new feature that adds functionality and extends the API.

Shall we add new feature to a bugfix release?

aleks-f:
> Related ticket #5201 is actually a new feature that adds functionality and extends the API.
> 
> Shall we add new feature to a bugfix release?

It doesn't break anything, it is just one more bit in options, off by default. strictly speaking, we shouldn't, but we need it and I don't want to wait for next release

matejk:
The name of the PR does not reflect the actual changes: it fixes multiple issies.

matejk (commented):
## PR Review: Process Termination Fixes

CI is green (all 88 checks pass, including ASAN/TSAN/UBSAN). The clang-cl fix in TestResult.cpp and the QNX `SPAWN_SETGROUP`/`SPAWN_NEWPGROUP` path look correct and clean.

### Critical

1. **`_hJob` handle leaked on normal process exit** — The Job Object is created in `run()` but only cleaned up in `stop()` inside the `_pPH` guard. If the child exits on its own, `run()` completes and nulls `_pPH`, so `stop()` skips the cleanup block. Handle leaks on every natural-exit cycle.

2. **Stale `errno` in `kill(-pid, 0)` polling loop** — `Thread::sleep(10)` between iterations can modify `errno`. If `kill()` returns -1/ESRCH (group gone) but `errno` was clobbered to EPERM by sleep internals, the `|| errno == EPERM` condition passes and the loop spins until timeout.

3. **Unchecked `setpgid(0, 0)` return** — If it fails, the child continues in the parent's process group. Later `kill(-pid, SIGINT)` targets a nonexistent group (ESRCH), the polling loop exits immediately, and `stop()` thinks the tree is dead.

### Important

4. **`start()` catch block doesn't close `_hJob`** — Handle leak on startup failure.

5. **Windows `stop()` doesn't wait for child tree** — Only waits for the main PID; closes the Job Object but returns immediately. Asymmetric with Unix behavior.

6. **`kill(-pid, SIGINT)` return unchecked** — EPERM on the initial signal gets masked as a timeout.

7. **fork/setpgid race** — Parent could call `kill(-pid, ...)` before child reaches `setpgid(0, 0)`. Standard fix: also call `setpgid(pid, pid)` in parent after fork.

### Minor

8. **No tests for `PROCESS_KILL_TREE`** — The existing ProcessRunnerTest doesn't exercise the new option.

9. **Silent degradation on Windows** — If Job Object creation fails, PROCESS_KILL_TREE is silently ignored with no log/warning.

10. **`hasConsole()` may over-detect** — Piped handles (e.g. from a service manager) would cause a false positive. Consider `GetFileType(h) == FILE_TYPE_CHAR` to confirm it's a console device.

See inline comments for details.

Foundation/src/ProcessRunner.cpp
```diff
@@ -126,6 +133,42 @@ void ProcessRunner::run()
...
+				}
+				else
+				{
+					CloseHandle(static_cast<HANDLE>(_hJob));
+					_hJob = nullptr;
+				}
+			}
+		}
+#endif
 		// Set _pPH after _pid to ensure pid() returns valid value when running() is true
```
matejk, line 184: **`_hJob` handle leak on normal process exit.**

This Job Object is created here in `run()`, but the only cleanup is in `stop()` at lines 291-296 — which is inside the `if (_pPH.exchange(nullptr) && ...)` guard.

If the child process exits normally, `run()` completes via `pPH->wait()` and sets `_pPH = nullptr` (line 203). A subsequent `stop()` finds `_pPH` already null and skips the inner block entirely, including the `_hJob` cleanup. The handle leaks.

On repeated start/stop cycles where the child exits on its own, this leaks a kernel handle per cycle. The leaked Job Object with `KILL_ON_JOB_CLOSE` will also unexpectedly kill assigned processes when the parent eventually exits.

Fix: add `_hJob` cleanup at the end of `run()` (after the try/catch, before `delete pPH`):

```cpp
	if (_hJob)
	{
		CloseHandle(static_cast<HANDLE>(_hJob));
		_hJob = nullptr;
	}
#endif
	_pid = INVALID_PID;
	_pPH = nullptr;
```

Alternatively, move the `_hJob` cleanup in `stop()` outside the `_pPH` guard so it runs unconditionally when `_started` is true.

Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +217,85 @@ void ProcessRunner::stop()
...
+			{
+				// Wait for the entire process group to exit.
+				// setpgid(0, 0) was called in the child, so -pid
+				// addresses the whole group.
+				// kill(-pid, 0) returns 0 if any process in the group exists
+				// and we have permission, or -1 with errno:
+				//   ESRCH  - no such process group (done)
+				//   EPERM  - group exists but no permission (still alive)
+				while (::kill(-pid, 0) == 0 || errno == EPERM)
+				{
```
matejk, line 283: **Stale `errno` bug.**

`errno` is only guaranteed to be meaningful immediately after a syscall that failed. `Thread::sleep(10)` at line 263 (position 107) may call `nanosleep`/`usleep` internally, which can set `errno`.

Scenario:
1. `kill(-pid, 0)` returns 0 (group alive) → loop continues, sleeps.
2. Group exits during sleep.
3. `kill(-pid, 0)` returns -1 with `errno = ESRCH` (group gone) → loop should exit.
4. But if `Thread::sleep` set `errno` to `EPERM` (or left a stale EPERM from a prior iteration), the `|| errno == EPERM` condition passes → loop keeps spinning until timeout.

The same pattern appears in the inner SIGKILL loop at position 97.

Fix: capture result and errno atomically after each `kill` call:

```cpp
for (;;)
{
	int rc = ::kill(-pid, 0);
	int err = errno;
	if (rc == -1 && err != EPERM)
		break; // ESRCH or other: group is gone
	if (_sw.elapsedSeconds() > _timeout)
	{
		// ... escalate to SIGKILL ...
	}
	Thread::sleep(10);
}
```

Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +217,85 @@ void ProcessRunner::stop()
...
+			// ServerApplication::_terminate member), the kernel event object is destroyed
+			// when our handle closes, and the termination signal is lost.
+			// Keep the NamedEvent alive here so the kernel object persists until the child
+			// opens its own handle and receives the signal.
+			NamedEvent terminateEvent(Process::terminationEventName(pid));
+			terminateEvent.set();
+#else
+			if (_options & PROCESS_KILL_TREE)
+				::kill(-pid, SIGINT); // signal entire process group
+			else
```
matejk, line 266: **Return value unchecked.**

If `kill(-pid, SIGINT)` fails with `EPERM` (no permission to signal the group), the code proceeds to the polling loop, which will spin for `_timeout` seconds before escalating to SIGKILL, which will also fail with EPERM, and then throw `TimeoutException("Unable to terminate process tree")`. The user sees a misleading timeout error instead of a permission error.

Consider checking the return value:
```cpp
if (::kill(-pid, SIGINT) != 0 && errno != ESRCH)
{
	throw Poco::SystemException(
		Poco::format("Cannot signal process group %d", (int)pid));
}
```

Foundation/src/Process_UNIX.cpp
```diff
@@ -256,6 +261,12 @@ ProcessHandleImpl* ProcessImpl::launchByForkExecImpl(const std::string& command,
 				close(j);
 			}
 
+			// Create a new process group so the entire tree can be signaled.
+			if (options & PROCESS_KILL_TREE)
+			{
+				setpgid(0, 0);
+			}
```
matejk, line 269: **Unchecked `setpgid(0, 0)` return.**

If `setpgid` fails (e.g. EPERM), the child continues into `execvp` in the parent's process group. Later, `ProcessRunner::stop()` calls `kill(-pid, SIGINT)`, where `-pid` addresses a nonexistent process group. `kill(-pid, 0)` in the polling loop immediately returns ESRCH, causing `stop()` to think the tree is already dead. The child and all its descendants become unmanageable orphans.

Since you can't throw after `fork()`, the standard pattern is to `_exit()` with a distinctive code:

```cpp
if (options & PROCESS_KILL_TREE)
{
	if (setpgid(0, 0) != 0)
		_exit(73); // distinctive code indicating setpgid failure
}
```

Also consider calling `setpgid(pid, pid)` in the **parent** after `fork()` returns (around line 274) to close the race window where `stop()` could be called before the child reaches `setpgid`:

```cpp
if (options & PROCESS_KILL_TREE)
	setpgid(pid, pid); // idempotent with child's setpgid(0,0)
```

Util/src/ServerApplication.cpp
```diff
@@ -270,8 +270,22 @@ bool ServerApplication::isService()
...
+	// but stdin is still inherited from the parent.
+	HANDLE handles[] = {
+		GetStdHandle(STD_OUTPUT_HANDLE),
+		GetStdHandle(STD_INPUT_HANDLE),
+		GetStdHandle(STD_ERROR_HANDLE)
+	};
+	for (HANDLE h : handles)
+	{
+		if (h != INVALID_HANDLE_VALUE && h != nullptr)
+			return true;
```
matejk: **Potential false positive with piped handles.**

If a service is launched by the SCM with redirected standard handles (e.g. a logging pipe on stdin), `hasConsole()` will return `true` even though there is no console. The `GetConsoleWindow()` check at line 273 is the authoritative console test; the fallback here could be tightened by verifying the handle is actually a console device:

```cpp
if (h != INVALID_HANDLE_VALUE && h != nullptr)
{
	if (GetFileType(h) == FILE_TYPE_CHAR)
		return true;
}
```

This avoids over-detecting when handles are pipes or files rather than console devices.

aleks-f: 3dc180d29

Foundation/src/Process_UNIX.cpp
aleks-f: 27e05eee1

Foundation/src/ProcessRunner.cpp
aleks-f: b28e44813

Foundation/src/ProcessRunner.cpp
aleks-f: 37b130002

Foundation/src/ProcessRunner.cpp
aleks-f: 2ec96b619

Foundation/src/ProcessRunner.cpp
```diff
@@ -126,6 +133,37 @@ void ProcessRunner::run()
 		_pid = pPH->id();
 		errPID = Error::last();
 
+#if defined(POCO_OS_FAMILY_WINDOWS)
+		if (_options & PROCESS_KILL_TREE)
+		{
```
matejk, line 138: **Job Object setup fails silently — KILL_TREE degrades to no-op.**

The entire chain `CreateJobObjectW` → `SetInformationJobObject` → `OpenProcess` → `AssignProcessToJobObject` (lines 22-47 in the diff) can fail at any stage. Every failure calls `_hJob.reset()` and continues silently — no log, no warning, no exception.

The most common real-world failure: `AssignProcessToJobObject` returns `ERROR_ACCESS_DENIED` when the child is already in a non-breakaway job (typical in CI environments, containers, and Windows service managers).

The user explicitly opts in to `PROCESS_KILL_TREE` but gets no feedback that the feature is inoperative. Orphaned process trees will accumulate silently.

**Suggestion:** At minimum, capture `GetLastError()` at each failure point and populate `_error` or log. Consider whether a failed setup should throw (hard error) or at least emit a warning.

Foundation/src/ProcessRunner.cpp
```diff
@@ -174,28 +215,94 @@ void ProcessRunner::stop()
 			if (pid <= 0)
 				throw Poco::IllegalStateException("Invalid PID, can't terminate process");
 
-			Process::requestTermination(pid);
-			while (Process::isRunning(pid))
+#if defined(POCO_OS_FAMILY_WINDOWS)
+			// On Windows, Process::requestTermination() creates a temporary NamedEvent,
```
matejk, line 245: **Windows `stop()` sends NamedEvent unconditionally when KILL_TREE is active.**

On Unix, the code branches on `PROCESS_KILL_TREE` (line 80) to decide whether to signal the group or the individual process. On Windows, the NamedEvent graceful signal is always sent regardless of the flag.

Two concerns:
1. The NamedEvent triggers graceful shutdown in the leader, which may start resource cleanup. The `_hJob` close later (line 163) forcefully terminates the tree via `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, potentially interrupting that cleanup.
2. The NamedEvent mechanism only works for Poco `ServerApplication`-based processes. Non-Poco children in the tree get no graceful termination signal at all — they're just killed when the Job Object handle closes.

**Suggestion:** Either document this asymmetry explicitly, or conditionally skip the NamedEvent when KILL_TREE is set and let the Job Object handle everything.

Foundation/src/Process_UNIX.cpp
```diff
@@ -256,10 +261,24 @@ ProcessHandleImpl* ProcessImpl::launchByForkExecImpl(const std::string& command,
 				close(j);
 			}
 
+			// Create a new process group so the entire tree can be signaled.
+			if (options & PROCESS_KILL_TREE)
+			{
+				if (setpgid(0, 0) != 0)
+					_exit(73);
+			}
```
matejk, line 269: **`_exit(73)` magic number is opaque to the parent.**

If `setpgid(0,0)` fails, the child exits with code 73. The parent observes this via `pPH->wait()` returning 73, and the error message becomes something like `"return=73 (0:Success)"` — indistinguishable from the application itself returning 73.

The actual `errno` from `setpgid` (e.g. `EPERM`) is lost forever — it's in the child's address space and never communicated to the parent.

**Suggestions:**
- Define a named constant: `static constexpr int EXIT_SETPGID_FAILED = 73;`
- In `ProcessRunner::run()`, when `PROCESS_KILL_TREE` is active and `_rc == 73`, set a specific error message:
  ```cpp
  if ((_options & PROCESS_KILL_TREE) && _rc == 73)
      setError("child failed to create process group (setpgid failed)");
  ```
- Consider using a pipe between parent and child to communicate the specific errno on failure (standard pattern for robust fork/exec error reporting).

Foundation/include/Poco/ProcessOptions.h
```diff
@@ -35,6 +35,17 @@ enum ProcessOptions
...
+
+	PROCESS_KILL_TREE = 8,
+		/// Ensures that all child processes (and their descendants) are
+		/// terminated when the ProcessRunner is stopped or destroyed.
+		///
+		/// On Windows, creates a Job Object with JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE.
+		///
+		/// On Unix/Linux, the child is placed in a new process group via
+		/// setpgid(0, 0). On stop, the termination request is sent to the
+		/// main process; if it does not exit before the timeout expires,
```
matejk: **Documentation/code mismatch.**

This says: *"On stop, the termination request is sent to the main process; if it does not exit before the timeout expires, the entire process group is signaled (forcibly) with SIGKILL."*

But the code at `ProcessRunner.cpp` line 231 sends `kill(-pid, SIGINT)` — which signals the **entire process group** immediately, not just the main process. The initial signal is group-wide SIGINT, then SIGKILL on timeout.

**Suggested fix:** Update the comment to match the actual behavior:
```
/// On Unix/Linux, the child is placed in a new process group via
/// setpgid(0, 0). On stop, SIGINT is sent to the entire process
/// group; if the group does not exit before the timeout expires,
/// SIGKILL is sent to the group.
```

Foundation/include/Poco/ProcessRunner.h
```diff
@@ -178,6 +178,29 @@ class Foundation_API ProcessRunner: public Poco::Runnable
 	Stopwatch _sw;
 	std::string _error;
 	mutable Poco::FastMutex _mutex;
+#if defined(POCO_OS_FAMILY_WINDOWS)
+	struct HandleGuard
+		/// RAII wrapper for a Windows HANDLE.
```
matejk, line 183: **HandleGuard design concerns.**

1. **Public `handle` member** — The RAII invariant (close-on-destroy) can be bypassed by writing `_hJob.handle = x` directly, leaking the old handle. The code already accesses `_hJob.handle` directly in `SetInformationJobObject` and `AssignProcessToJobObject` calls. Compare with `ProcessHandleImpl` in `Process_WIN32U.h` which uses a private `_hProcess` + `process()` accessor. A `get()` const accessor would be more consistent.

2. **No self-assignment protection in `reset()`** — `_hJob.reset(_hJob.handle)` would `CloseHandle` the handle, then assign `nullptr`. The standard pattern is: save old, assign new, close old.

3. **No move semantics** — Not needed today, but the generic name "HandleGuard" suggests potential reuse. Missing move ctor/assign prevents safe ownership transfer.

aleks-f: 6a6545997

Foundation/include/Poco/ProcessOptions.h
aleks-f: 2d590245e

Foundation/src/Process_UNIX.cpp
aleks-f: 2d590245e

Foundation/src/ProcessRunner.cpp
aleks-f: 309751398

Foundation/src/ProcessRunner.cpp
aleks-f: be9b9f1f3

PR #5209, opened 2026-02-15
fix: Add MINGW compiler macro for MSYS2 Clang64 environment

feihongmeilian:
In the MSYS2 clang64 compilation environment, although the compiler is Clang, it is based on the MinGW runtime underneath. In this case, the Windows SEH syntax with __try/__except in the code is incompatible, which leads to compilation failures.

By detecting the MINGW32/MINGW64 macros under the Clang compiler branch and defining POCO_COMPILER_MINGW, subsequent code can skip the __try/__except related logic to ensure compilation compatibility.

### DIFF
--- a/Foundation/include/Poco/NotificationQueue.h
+++ b/Foundation/include/Poco/NotificationQueue.h
@@ -149,6 +149,7 @@ private:
 
 	NfQueue           _nfQueue;
 	WaitQueue         _waitQueue;
+	bool              _wokeUp = false;
 	mutable FastMutex _mutex;
 };
 
--- a/Foundation/src/File.cpp
+++ b/Foundation/src/File.cpp
@@ -26,19 +26,6 @@ std::string findInPath(const std::string& name);
 }
 
 
-#if defined(POCO_OS_FAMILY_WINDOWS)
-#include "File_WIN32U.cpp"
-#elif defined(POCO_VXWORKS)
-#include "File_VX.cpp"
-#elif defined(POCO_OS_FAMILY_UNIX)
-#include "File_UNIX.cpp"
-#endif
-#include "Poco/Thread.h"
-
-
-namespace Poco {
-
-
 namespace {
 
 /// Return cached PATH directories as a vector of strings.
@@ -46,6 +33,10 @@ namespace {
 /// Thread-safe: uses mutex to protect cache updates.
 std::vector<std::string> getPathDirectories()
 {
+	using Poco::Environment;
+	using Poco::Path;
+	using Poco::StringTokenizer;
+
 	static std::mutex mutex;
 	static std::string cachedPath;
 	static std::vector<std::string> cachedDirs;
@@ -76,6 +67,19 @@ std::vector<std::string> getPathDirectories()
 }
 
 } // anonymous namespace
+
+
+#if defined(POCO_OS_FAMILY_WINDOWS)
+#include "File_WIN32U.cpp"
+#elif defined(POCO_VXWORKS)
+#include "File_VX.cpp"
+#elif defined(POCO_OS_FAMILY_UNIX)
+#include "File_UNIX.cpp"
+#endif
+#include "Poco/Thread.h"
+
+
+namespace Poco {
 
 
 std::string findInPath(const std::string& name)
--- a/Foundation/src/File_UNIX.cpp
+++ b/Foundation/src/File_UNIX.cpp
@@ -17,6 +17,8 @@
 #include "Poco/Exception.h"
 #include "Poco/Error.h"
 #include "Poco/Path.h"
+#include "Poco/Environment.h"
+#include "Poco/StringTokenizer.h"
 #include <algorithm>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -90,27 +92,35 @@ std::string FileImpl::getExecutablePathImpl() const
 		Path p(_path);
 		p.makeAbsolute();
 		std::string absPath = p.toString();
-
-		// Optimized: check existence and executability in one stat() call
-		struct stat st;
-		if (::stat(absPath.c_str(), &st) != 0 || !S_ISREG(st.st_mode))
-			return {};
-
-		bool executable;
-		if (st.st_uid == ::geteuid() || ::geteuid() == 0)
-			executable = (st.st_mode & S_IXUSR) != 0;
-		else if (st.st_gid == ::getegid())
-			executable = (st.st_mode & S_IXGRP) != 0;
-		else
-			executable = (st.st_mode & S_IXOTH) != 0;
-
-		return executable ? absPath : std::string();
+		return canExecuteImpl(absPath) ? absPath : std::string();
 	}
 
-	std::string found = findInPath(_path);
-	if (!found.empty() && !canExecuteImpl(found))
-		return {};
-	return found;
+	// Bare name — search CWD then PATH, requiring regular executable file.
+	// Cannot use findInPath() here because it returns the first existing
+	// entry (including directories) with no way to continue searching.
+	{
+		Path curPath(Path::current());
+		curPath.append(_path);
+		curPath.makeAbsolute();
+		std::string absPath = curPath.toString();
+		if (canExecuteImpl(absPath))
+			return absPath;
+	}
+
+	for (const auto& dir : getPathDirectories())
+	{
+		try
+		{
+			Path candidate(dir);
+			candidate.append(_path);
+			candidate.makeAbsolute();
+			std::string absPath = candidate.toString();
+			if (canExecuteImpl(absPath))
+				return absPath;
+		}
+		catch (const Poco::PathSyntaxException&) {}
+	}
+	return {};
 }
 
 
@@ -169,7 +179,9 @@ bool FileImpl::canExecuteImpl(const std::string& absolutePath) const
 	if (::stat(absolutePath.c_str(), &st) != 0 || !S_ISREG(st.st_mode))
 		return false;
 
-	if (st.st_uid == ::geteuid() || ::geteuid() == 0)
+	if (::geteuid() == 0)
+		return (st.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH)) != 0;
+	else if (st.st_uid == ::geteuid())
 		return (st.st_mode & S_IXUSR) != 0;
 	else if (st.st_gid == ::getegid())
 		return (st.st_mode & S_IXGRP) != 0;
--- a/Foundation/src/NotificationQueue.cpp
+++ b/Foundation/src/NotificationQueue.cpp
@@ -40,6 +40,7 @@ void NotificationQueue::enqueueNotification(Notification::Ptr pNotification)
 {
 	poco_check_ptr (pNotification);
 	FastMutex::ScopedLock lock(_mutex);
+	_wokeUp = false;
 	if (_waitQueue.empty())
 	{
 		_nfQueue.push_back(std::move(pNotification));
@@ -58,6 +59,7 @@ void NotificationQueue::enqueueUrgentNotification(Notification::Ptr pNotificatio
 {
 	poco_check_ptr (pNotification);
 	FastMutex::ScopedLock lock(_mutex);
+	_wokeUp = false;
 	if (_waitQueue.empty())
 	{
 		_nfQueue.push_front(std::move(pNotification));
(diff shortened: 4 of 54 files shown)


feihongmeilian:
Note: In my opinion, a better approach would be to separate platform detection (__MINGW32__/__MINGW64__) from compiler detection, and define POCO_COMPILER_MINGW in a standalone check instead of duplicating the logic inside each compiler branch.

This would make the code cleaner and easier to maintain. However, since I don't have access to enough platforms and environments for full validation, I'm applying this minimal fix for now to resolve the compilation issue on MSYS2 Clang64.

The separated detection logic can be refactored in a future change after more comprehensive testing.

Trace/include/Poco/Trace/cpptrace/cpptrace.hpp
```diff
@@ -11,7 +11,7 @@
 #include <utility>
 #include <vector>
 
-#ifdef _WIN32
+#ifdef _WIN32 && defined(_MSC_VER)
```
matejk: **Bug:** `#ifdef` only accepts a single identifier — it does **not** support expressions. The `&& defined(_MSC_VER)` part will either be silently ignored or cause a preprocessor error depending on the compiler/mode.

```c
// Current (broken):
#ifdef _WIN32 && defined(_MSC_VER)

// Should be:
#if defined(_WIN32) && defined(_MSC_VER)
```

That said, see my top-level review comment — I believe this change (and all the other `__declspec` guard changes) should actually be reverted entirely.

Trace/include/Poco/Trace/ctrace/ctrace.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 #include <stdio.h>
 
-#ifdef _WIN32
+#ifdef _WIN32 && defined(_MSC_VER)
```
matejk: **Bug:** Same `#ifdef` vs `#if` syntax error as in `cpptrace.hpp`. `#ifdef` does not support compound expressions.

```c
// Current (broken):
#ifdef _WIN32 && defined(_MSC_VER)

// Would need to be:
#if defined(_WIN32) && defined(_MSC_VER)
```

Data/SQLParser/src/sqlparser_win.h
```diff
@@ -15,7 +15,7 @@
 		#define SQLParser_EXPORTS
 	#endif
 #else
-	#if defined(_DLL) || defined(_USRDLL)
+	#if defined(_DLL) || defined(_USRDLL) && defined(_MSC_VER)
```
matejk: **Bug: operator precedence.** In the C preprocessor `&&` binds tighter than `||`, so this evaluates as:

```c
#if defined(_DLL) || (defined(_USRDLL) && defined(_MSC_VER))
```

The `defined(_DLL)` path is completely unguarded by `_MSC_VER`, which is almost certainly not the intent. If this change were to be kept, it should be:

```c
#if (defined(_DLL) || defined(_USRDLL)) && defined(_MSC_VER)
```

Foundation/include/Poco/ClassLibrary.h
```diff
@@ -23,7 +23,7 @@
 #include <typeinfo>
 
 
-#if defined(_WIN32)
+#if defined(_WIN32) && defined(_MSC_VER)
```
matejk: **Regression:** This breaks MinGW plugin loading. With this change, MinGW builds will fall through to the `__GNUC__` visibility branch, which uses ELF-style `__attribute__((visibility("default")))`. That attribute has no effect on Windows — plugins built with MinGW will fail to export the required `ClassLoader` entry points.

`__declspec(dllexport)` is correct and supported by all Windows compilers (MSVC, GCC-MinGW, Clang-MinGW). The original `#if defined(_WIN32)` guard was right here.

CppParser/include/Poco/CppParser/CppParser.h
```diff
@@ -31,7 +31,7 @@
 // CppParser_API functions as being imported from a DLL, whereas this DLL sees symbols
 // defined with this macro as being exported.
 //
-#if (defined(_WIN32) || defined(__CYGWIN__)) && defined(POCO_DLL)
+#if defined(_WIN32) && defined(_MSC_VER) && defined(POCO_DLL)
```
matejk: **Regression:** The original guard included `defined(__CYGWIN__)`. This change silently drops Cygwin DLL export/import support. The same issue applies to `NetSSL_OpenSSL/include/Poco/Net/NetSSL.h` and `NetSSL_Win/include/Poco/Net/NetSSL.h`.

Foundation/include/Poco/Platform.h
```diff
@@ -250,6 +250,9 @@
 	#if __has_include(<cxxabi.h>)
 		#define POCO_HAVE_CXXABI_H
 	#endif
+	#if defined (__MINGW32__) || defined (__MINGW64__)
```
matejk, line 253: 👍 This addition is correct and valuable. Currently, MSYS2 Clang64 gets `POCO_COMPILER_CLANG` but not `POCO_COMPILER_MINGW`, which means MinGW-specific workarounds (e.g. avoiding `__try/__except` SEH) do not trigger. This is the right fix for the core issue described in this PR.

Trace/include/Poco/Trace/cpptrace/cpptrace.hpp
```diff
@@ -11,7 +11,7 @@
...
 #define CPPTRACE_EXPORT_ATTR __declspec(dllexport)
```
matejk, line 15: **Maintenance concern:** This is a vendored upstream header (cpptrace). Changes here will be overwritten on the next vendor update. If a fix is needed, it would be better handled via a patch file or build-system defines that avoid modifying the vendored source directly.

feihongmeilian:
I added the check for _MSC_VER because when compiling POCO, the JSON library depends on the Foundation library. Without the _MSC_VER check, the code would use the __declspec syntax. The following code exists in Foundation\include\Poco\Dynamic\Struct.h:
```

extern template class Struct<std::string>;
extern template class Struct<int>;

extern template class Struct<std::string, Poco::OrderedMap<std::string, Var>, Poco::OrderedSet<std::string>>;
extern template class Struct<int, OrderedMap<int, Var>, OrderedSet<int>>;

#else

extern template class Foundation_API Struct<std::string>;
extern template class Foundation_API Struct<int>;

extern template class Foundation_API Struct<std::string, Poco::OrderedMap<std::string, Var>, Poco::OrderedSet<std::string>>;
extern template class Foundation_API Struct<int, OrderedMap<int, Var>, OrderedSet<int>>;

#endif

```
It is mandatory to add Foundation_API before the template classes within the POCO_OS_FAMILY_WINDOWS block as well; otherwise, the function definitions cannot be found. However, MSVC (Microsoft Visual C++) behaves correctly in this scenario: it automatically adds Foundation_API to template classes, so the missing Foundation_API in the Windows branch does not cause issues for pure MSVC builds.
That said, after modifying this part for non-MSVC Windows compilers (e.g., mingw64/clang), many other places still need to be adjusted, and some template classes require additional export syntax. When I added the _MSC_VER check, the code uses the __attribute__ syntax instead, which resolves this issue universally. The POCO dynamic library can be fully compiled under clang and mingw64 in msys2, so I confirm this approach works.
That said, the following modification might be more robust and compatible with the clang-cl compiler:
```
#if defined(_WIN32) && defined(POCO_DLL)
    #if defined(_MSC_VER) || (defined(__clang__) && defined(_MSC_VER))
        #ifdef Foundation_EXPORTS
            #define Foundation_API __declspec(dllexport)
        #else
            #define Foundation_API __declspec(dllimport)
        #endif
    #else
        #ifdef Foundation_EXPORTS
            #define Foundation_API __attribute__((visibility("default")))
        #else
            #define Foundation_API
        #endif
    #endif
#endif

#if !defined(Foundation_API)
	#if !defined(POCO_NO_GCC_API_ATTRIBUTE) && defined (__GNUC__) && (__GNUC__ >= 4)
		#define Foundation_API __attribute__ ((visibility ("default")))
	#else
		#define Foundation_API
	#endif
#endif
```

feihongmeilian:
I am not very familiar with submitting changes to GitHub. For each submission, I end up forking the repository again from scratch. Should I close this current pull request (PR), re-fork the repository, and then create three separate branches in the forked repository to submit my changes individually?
I originally intended to make three separate submissions. However, since I did not know how to fork the repository three times (or manage multiple forks properly), and my first commit was made directly on the main branch, all of my changes ended up being included in a single pull request.

matejk:
You can add more commits to the branch you created for the PR and push them.

matejk:
@feihongmeilian Thanks for the detailed explanation about the `extern template` issue — that's a real problem and I appreciate you digging into it.

However, I believe this was already resolved by PR #5202 (merged Feb 12, before this PR was created). That PR changed all the `extern template` guards from:

```cpp
extern template class Struct<std::string>;           // no _API macro
#else
extern template class Foundation_API Struct<std::string>;
#endif
```

to:

```cpp
#if defined(POCO_OS_FAMILY_WINDOWS) && defined(Foundation_EXPORTS)
extern template class Struct<std::string>;           // only when building the DLL
#else
extern template class Foundation_API Struct<std::string>; // consumers + non-Windows
#endif
```

This means MinGW consumers now always take the `#else` branch and get the properly `__declspec(dllimport)`-decorated `extern template` declarations, which should resolve the linker errors you were seeing. Since your branch is based on `main`, you should already have this fix.

Given that, the `__declspec` → `__attribute__((visibility))` workaround in this PR should no longer be necessary. I'd suggest:

1. **Keep** the `POCO_COMPILER_MINGW` addition in `Platform.h` — this is correct and valuable
2. **Keep** the ODBCVER bump (ideally as a separate PR, but not a blocker)
3. **Drop** all the `_MSC_VER` guard changes to the `_API` macros — with #5202 merged, `__declspec(dllexport/dllimport)` should work correctly on MinGW

Could you rebase on current `main` and verify that the MinGW/Clang64 build works without the `_API` macro changes?

feihongmeilian:
I've reverted the commit regarding symbol exports and split the ODBCVER modification into a new PR. However, after updating my local branch to main, I'm still encountering compilation failures with Clang in MSYS2. I may need to explore alternative approaches to fix this issue going forward.

feihongmeilian:
@matejk I have concerns about the appropriateness of the following implementation:
```
#if defined(POCO_OS_FAMILY_WINDOWS) && defined(Foundation_EXPORTS)
extern template class Struct<std::string>;           // only when building the DLL
#else
extern template class Foundation_API Struct<std::string>; // consumers + non-Windows
#endif
```

When POCO_OS_FAMILY_WINDOWS is defined and Foundation_EXPORTS is set, the code path extern template class Struct<std::string>; will be taken. Could this result in missing symbol exports from the DLL when the function is used?
Furthermore, I think conditional checks for function exports are less optimal than directly setting Foundation_API to an empty value in specific scenarios. Splitting the judgment logic between export and import stages will lead to fragmented logic and increase the risk of errors.
To simplify the logic and avoid potential symbol export issues, I suggest removing the conditional branches and directly using the unified declaration:
```
extern template class Foundation_API Struct<std::string>;
```

This approach centralizes the export/import control in the Foundation_API macro itself (e.g., setting it to __declspec(dllexport), __declspec(dllimport) or an empty value based on compilation context), eliminating scattered conditional judgments and ensuring consistent symbol handling across all platforms and build scenarios.

matejk:
@feihongmeilian The current pattern is correct. The `extern template` declaration in the header only **suppresses implicit instantiation** — it does not control symbol export. The actual export happens in the `.cpp` file.

When **building the DLL** (`Foundation_EXPORTS` is defined):

- `Struct.h:291` — `extern template class Struct<std::string>;` — tells the compiler "don't instantiate this here," no export annotation needed
- `VarHolder.cpp:27` — `template class Foundation_API Struct<std::string>;` — this is where the single explicit instantiation happens, and `Foundation_API` expands to `__declspec(dllexport)`, so the symbols get exported

When **consuming the DLL** (`Foundation_EXPORTS` is not defined):

- `Struct.h:299` — `extern template class Foundation_API Struct<std::string>;` — here `Foundation_API` expands to `__declspec(dllimport)`, so the consumer knows to import these symbols from the DLL

Your suggestion to unconditionally use `extern template class Foundation_API Struct<...>;` would mean that when building the DLL we'd get `extern template class __declspec(dllexport) Struct<...>;`. MSVC has known issues with this combination — it may ignore the `extern` and instantiate the template in every translation unit, causing duplicate symbols. The two-branch approach exists specifically to avoid this.

Regarding Clang on Windows: both Clang-CL and Clang/MinGW support `__declspec(dllexport/dllimport)` — this is a PE/COFF format requirement, not MSVC-specific. `__attribute__((visibility("default")))` is an ELF concept and has no effect on Windows DLL exports.

If you're still seeing build failures with Clang64 after rebasing on current `main`, could you share the specific error messages? That would help identify whether the issue is in the `extern template` handling or somewhere else.

PR #5224, opened 2026-02-24
fix(File): is*() and can*() return false for non-existent files

### LINKED ISSUE #5212: File throws on query
We have `File` throwing on querying in `is*()` and `can*()` when an entity verifiably does not exist. All these functions should first check `File::exists()` and just return false if it does not exist.

matejk:
## Summary
- Boolean query methods (`canRead`, `canWrite`, `isFile`, `isDirectory`, `isLink`, `isDevice`, `isHidden`) now return `false` when the file does not exist, instead of throwing `FileNotFoundException`
- The error code from the underlying system call (`stat`/`GetFileAttributesW`) is checked directly, avoiding the TOCTOU race that a separate `exists()` check would introduce
- Other errors (I/O, permissions) still throw as before
- Fixed Windows `isFileImpl()` which previously delegated to `!isDirectoryImpl() && !isDeviceImpl()` — would have incorrectly returned `true` for non-existent files after the change

Closes #5212

## Test plan
- [x] Updated `testFileAttributes1` to assert `false` returns instead of expecting exceptions
- [x] Added assertions for `isLink()`, `isDevice()`, `isHidden()` on non-existent files
- [x] All 31 FileTest tests pass

### DIFF
--- a/Foundation/src/File_UNIX.cpp
+++ b/Foundation/src/File_UNIX.cpp
@@ -147,7 +147,10 @@ bool FileImpl::canReadImpl() const
 		else
 			return (st.st_mode & S_IROTH) != 0 || ::geteuid() == 0;
 	}
-	else handleLastErrorImpl(_path);
+	else if (const auto err = errno; err == ENOENT)
+		return false;
+	else
+		handleLastErrorImpl(err, _path);
 	return false;
 }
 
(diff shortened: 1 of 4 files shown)


matejk:
The `const auto err = errno` init-statement is intentional — it captures errno once and forwards it to `handleLastErrorImpl(err, _path)`, ensuring the same value is used for both the ENOENT check and the error handler. Reading errno twice would be fragile if any code is ever inserted between the two reads.

1. Guarantees the same errno value is used in both the check and the error handler
2. Protects against future code changes that might insert something between the check and the handler (e.g. logging)
3. Calls handleLastErrorImpl(int, string) directly rather than going through the handleLastErrorImpl(string) wrapper that re-reads errno

PR #5227, opened 2026-02-25
fix(Net): reactor-based http server not response when requests with body 

sky92zwq:
Some problems fixed [PR](https://github.com/pocoproject/poco/pull/4946) :

1. not response when requests with body
2. the reactor server can't quit with `CTRL + C` correctly.
3. fix redefination of  `HTTPReactorServer.h`

### DIFF
--- a/Net/include/Poco/Net/TCPReactorAcceptor.h
+++ b/Net/include/Poco/Net/TCPReactorAcceptor.h
@@ -41,7 +41,7 @@ private:
 	std::shared_ptr<ThreadPool>                 _threadPool;
 	RecvMessageCallback                         _recvMessageCallback;
 	TCPServerParams::Ptr                        _pParams;
-	std::atomic<bool>                           _stopped;
+	std::atomic<bool>                           _stopped{false};
 };
 
 }} // namespace Poco::Net
--- a/Net/src/HTTPReactorServer.cpp
+++ b/Net/src/HTTPReactorServer.cpp
@@ -4,16 +4,19 @@
 #include "Poco/Net/HTTPSession.h"
 #include <cstring>
 
-namespace Poco { namespace Net {
+namespace Poco {
+namespace Net {
 
 HTTPReactorServer::HTTPReactorServer(int port, HTTPServerParams::Ptr pParams, HTTPRequestHandlerFactory::Ptr pFactory)
-: _tcpReactorServer(port, pParams)
+	: _tcpReactorServer(port, pParams)
 {
 	_pParams = pParams;
 	_pFactory = pFactory;
-	_tcpReactorServer.setRecvMessageCallback([this](const TcpReactorConnectionPtr& conn) {
-		this->onMessage(conn);
-	});
+	_tcpReactorServer.setRecvMessageCallback(
+		[this](const TcpReactorConnectionPtr& conn)
+		{
+			this->onMessage(conn);
+		});
 }
 
 HTTPReactorServer::~HTTPReactorServer()
@@ -41,7 +44,7 @@ void HTTPReactorServer::onMessage(const TcpReactorConnectionPtr& conn)
 		}
 
 		HTTPServerResponseImpl response(session);
-		HTTPServerRequestImpl  request(response, session, _pParams);
+		HTTPServerRequestImpl request(response, session, _pParams);
 
 		Poco::Timestamp now;
 		response.setDate(now);
@@ -79,9 +82,8 @@ void HTTPReactorServer::onMessage(const TcpReactorConnectionPtr& conn)
 			{
 				try
 				{
-					sendErrorResponse(
-						session,
-						e.code() == 0 ? HTTPResponse::HTTP_INTERNAL_SERVER_ERROR : HTTPResponse::HTTPStatus(e.code()));
+					sendErrorResponse(session, e.code() == 0 ? HTTPResponse::HTTP_INTERNAL_SERVER_ERROR
+															 : HTTPResponse::HTTPStatus(e.code()));
 					session.popCompletedRequest();
 				}
 				catch (...)
--- a/Net/src/TCPReactorAcceptor.cpp
+++ b/Net/src/TCPReactorAcceptor.cpp
@@ -6,8 +6,8 @@ namespace Poco {
 namespace Net {
 
 
-TCPReactorAcceptor::TCPReactorAcceptor(
-	Poco::Net::ServerSocket& socket, Poco::Net::SocketReactor& reactor, TCPServerParams::Ptr pParams)
+TCPReactorAcceptor::TCPReactorAcceptor(Poco::Net::ServerSocket& socket, Poco::Net::SocketReactor& reactor,
+									   TCPServerParams::Ptr pParams)
 	: Poco::Net::SocketAcceptor<TCPReactorServerConnection>(socket, reactor),
 	  _selfReactor(reactor),
 	  _useSelfReactor(pParams->getUseSelfReactor()),
@@ -35,16 +35,17 @@ TCPReactorAcceptor::~TCPReactorAcceptor()
 	stop();
 }
 
-void TCPReactorAcceptor::stop() {
+void TCPReactorAcceptor::stop()
+{
 	if (_stopped.exchange(true))
 	{
 		return;
 	}
-	for(auto& worker: _wokerReactors)
+	for (auto& worker : _wokerReactors)
 	{
 		worker->stop();
 	}
-	if(_threadPool)
+	if (_threadPool)
 	{
 		_threadPool->joinAll();
 	}
--- a/Net/src/TCPReactorServer.cpp
+++ b/Net/src/TCPReactorServer.cpp
@@ -9,11 +9,11 @@ namespace Net {
 
 
 TCPReactorServer::TCPReactorServer(int port, TCPServerParams::Ptr pParams)
-: _threadPool("TCPRA", pParams->getAcceptorNum()),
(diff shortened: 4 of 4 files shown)


Net/include/Poco/Net/HTTPReactorServer.h
```diff
@@ -31,3 +34,4 @@ class Net_API HTTPReactorServer
 
 }} // namespace Poco::Net
 
+#endif // Net_HTTPReactorServer_INCLUDED
```
matejk:
```suggestion
#endif // Net_HTTPReactorServer_INCLUDED
```
Missing newline at end of file. The diff shows `\ No newline at end of file`. This violates POSIX and can cause compiler warnings with `-pedantic`. The file must end with a trailing newline.

matejk (commented):
## Comprehensive Review

The PR fixes three real bugs:
1. **Body reading** — `read()` override is needed and the approach is correct
2. **CTRL+C hang** — `_stopped` guard + worker reactor stop in destructor fixes the shutdown hang
3. **Include guard** — correct, follows POCO convention

However, there are several issues to address: a missing newline at EOF, code style violations (POCO uses Allman brace style), an incomplete shutdown sequence, and a questionable use of chunked encoding for empty error responses. See inline comments for details.

### Missing test coverage
There is no test that exercises the body-reading path (e.g., a POST request with body via the reactor server). The only test change adds `reactor->stop()` to an existing test. A dedicated test for body reading would significantly increase confidence in the fix.

Net/src/HTTPReactorServerSession.cpp
```diff
@@ -209,6 +209,19 @@ int HTTPReactorServerSession::peek()
...
+int HTTPReactorServerSession::read(char* buffer, std::streamsize length)
+{
+	if (_idx < _complete)
+	{
+		int n = (int) (_complete - _idx);
+		if (n > length) n = (int) length;
+		std::memcpy(buffer, _buf.c_str() + _idx, n);
+		_idx += n;
+		return n;
+	}
```
matejk, line 221:
```suggestion
int HTTPReactorServerSession::read(char* buffer, std::streamsize length)
{
	if (_idx < _complete)
	{
		int n = static_cast<int>(_complete - _idx);
		if (n > static_cast<int>(length)) n = static_cast<int>(length);
		std::memcpy(buffer, _buf.data() + _idx, n);
		_idx += n;
		return n;
	}
	return 0;
}
```
Two minor improvements:
1. Use `_buf.data()` instead of `_buf.c_str()` — semantically more appropriate for a binary copy (not a C-string operation).
2. Use `static_cast` instead of C-style casts for consistency with modern C++ style.

Net/src/TCPReactorAcceptor.cpp
```diff
@@ -28,6 +28,16 @@ TCPReactorAcceptor::TCPReactorAcceptor(
...
 {
+	stop();
+}
+
+void TCPReactorAcceptor::stop() {
+	for(auto& worker: _wokerReactors) {
+		worker->stop();
+	}
+	if(_threadPool) {
+		_threadPool->joinAll();
```
matejk, line 50:
```suggestion
void TCPReactorAcceptor::stop()
{
	for (auto& worker : _wokerReactors)
	{
		worker->stop();
	}
	if (_threadPool)
	{
		_threadPool->joinAll();
	}
}
```
**Code style:** POCO uses Allman brace style (opening brace on its own line) and requires a space before `(` in control statements (`for (`, `if (`). The current code uses K&R/attached style which is inconsistent with the rest of the codebase.

Also, `stop()` is not idempotent — it can be called explicitly and then again from the destructor. Consider adding an `std::atomic<bool> _stopped` guard (like `TCPReactorServer` does) for consistency and safety.

Net/src/HTTPReactorServer.cpp
```diff
@@ -99,6 +99,8 @@ void HTTPReactorServer::sendErrorResponse(HTTPSession& session, HTTPResponse::HT
 	response.setVersion(HTTPMessage::HTTP_1_1);
 	response.setStatusAndReason(status);
 	response.setKeepAlive(false);
+	response.setChunkedTransferEncoding(true);
```
matejk: Adding `response.send()` is critical and correct — without it the error response was never actually sent to the client.

However, using chunked transfer encoding for a bodiless error response is unnecessary overhead (it emits a `Transfer-Encoding: chunked` header plus a `0\r\n\r\n` terminator). Consider instead:

```suggestion
	response.setContentLength(0);
	response.send();
```

Net/src/TCPReactorServer.cpp
```diff
@@ -46,6 +47,10 @@ void TCPReactorServer::setRecvMessageCallback(const RecvMessageCallback& cb)
 
 void TCPReactorServer::stop()
 {
+	if (_stopped.exchange(true))
```
matejk, line 54: The `_stopped` guard is good and fixes the CTRL+C double-stop issue.

However, `stop()` only stops the main acceptor reactors — the **worker reactors** inside each `TCPReactorAcceptor` are only stopped later when the acceptors are destroyed in the `TCPReactorServer` destructor. Between `stop()` returning and destruction, worker threads are still running. Consider also stopping acceptors explicitly:

```suggestion
	if (_stopped.exchange(true))
	{
		return;
	}
	for (auto& acceptor : _acceptors)
	{
		acceptor->stop();
	}
```

This ensures all threads (both acceptor and worker) are fully stopped when `stop()` returns, rather than relying on eventual destructor cleanup.

Net/samples/HTTPReactorTimeServer/src/HTTPReactorTimeServer.cpp
```diff
@@ -89,6 +91,38 @@ class TimeRequestHandler: public HTTPRequestHandler
 };
 
 
+class BodyEchoRequestHandler: public HTTPRequestHandler
```
matejk, line 94: Nit: the empty constructor body can be replaced with `= default`:
```suggestion
class BodyEchoRequestHandler: public HTTPRequestHandler
	/// Echo the request body back to the client.
{
public:
	BodyEchoRequestHandler() = default;
```

matejk:
@sky92zwq, can you take a look at the code review comments?

IMO it makes sense merging this PR into next patch release.

sky92zwq:
> @sky92zwq, can you take a look at the code review comments?
> 
> IMO it makes sense merging this PR into next patch release.

@matejk OK. I’ve been quite busy these past two weeks, but I’ll try to address the review comments over the weekend. I’m also working on optimizing performance for a large number of concurrent connections.

matejk:
@sky92zwq 

Release for 1.15.1 is planned for next week. Shall we merge this PR and put performance optimisations to 1.15.2 or even 1.16?

sky92zwq:
> @sky92zwq
> 
> Release for 1.15.1 is planned for next week. Shall we merge this PR and put performance optimisations to 1.15.2 or even 1.16?

SGTM. Merging this now and handling the performance optimizations later makes sense.

Net/include/Poco/Net/TCPReactorAcceptor.h
```diff
@@ -40,6 +41,7 @@ class Net_API TCPReactorAcceptor : public Poco::Net::SocketAcceptor<TCPReactorSe
 	std::shared_ptr<ThreadPool>                 _threadPool;
 	RecvMessageCallback                         _recvMessageCallback;
 	TCPServerParams::Ptr                        _pParams;
+	std::atomic<bool>                           _stopped;
```
matejk: **Bug:** `_stopped` is not initialized. `std::atomic<bool>` has a trivial default constructor that leaves the value indeterminate in C++17 and earlier. If `_stopped` happens to start as `true`, `stop()` will return immediately without stopping anything.

Should be:
```suggestion
	std::atomic<bool>                           _stopped{false};
```

(Note: `TCPReactorServer` correctly initializes `_stopped(false)` in its constructor initializer list, but `TCPReactorAcceptor` does not.)

Net/src/TCPReactorAcceptor.cpp
```diff
@@ -16,18 +16,38 @@ TCPReactorAcceptor::TCPReactorAcceptor(
...
+		_useSelfReactor = true;
 	}
 }
 
 TCPReactorAcceptor::~TCPReactorAcceptor()
 {
+	stop();
+}
+
+void TCPReactorAcceptor::stop() {
```
matejk: Minor style nits still remaining:
1. Function definition brace should be on its own line (Allman style)
2. Missing space before `(` in `for(` (line 43) and `if(` (line 47)

```suggestion
void TCPReactorAcceptor::stop()
{
```

Net/src/TCPReactorServer.cpp
```diff
@@ -9,15 +9,20 @@ namespace Net {
 
 
 TCPReactorServer::TCPReactorServer(int port, TCPServerParams::Ptr pParams)
-	: _threadPool("TCPR", pParams->getAcceptorNum()),
-	  _reactors(pParams->getAcceptorNum()),
-	  _pParams(pParams),
-	  _port(port)
+: _threadPool("TCPRA", pParams->getAcceptorNum()),
```
matejk: The initializer list indentation changed from tab-indented (POCO convention) to no indentation:
```cpp
// Before (POCO style)
\t: _threadPool(...),
\t  _reactors(...),
// Now
: _threadPool(...),
  _reactors(...),
```
This is inconsistent with the rest of the codebase. The same applies to `HTTPReactorServer.cpp`.

PR #5235, opened 2026-03-09
Use C++17 nested declarations and add CppParser support, closes #5234

### LINKED ISSUE #5234: Use nested namespace declarations
The library currently uses repeated namespace declarations to nested namespaces. As of C++17, [nested namespace declarations](https://en.cppreference.com/w/cpp/language/namespace.html) were introduced and allow concisely a single declaration for nested namespaces:
```cpp
// currently...
namespace Poco {
namespace Net {
namespace ODBC { 
    // ...
}}} // namespace Poco::Data::ODBC

// proposed new-style (C++17 onward)
namespace Poco::Net::ODBC {
    // ...
} // namespace Poco::Data::ODBC
```

As this syntax was introduced in C++17, and the library is C++20, there should not be any code breaking in such a change.

This would probably be difficult to do by hand, so perhaps a Python script could be used to do this quickly.

mikomikotaishi:
This pull request closes #5234, and replaces the repeated namespace declarations with the new C++17 [nested namespace declarations](https://en.cppreference.com/w/cpp/language/namespace.html) for a concise declaration of multiple nested namespaces. (The library is C++20, and thus should not be broken by this change.)
```cpp
// currently...
namespace Poco {
namespace Net {
namespace ODBC { 
    // ...
}}} // namespace Poco::Data::ODBC

// proposed new-style (C++17 onward)
namespace Poco::Net::ODBC {
    // ...
} // namespace Poco::Data::ODBC
```

This was done using a Python script (and a small part by hand for more complex files), and the CI tests don't report any build errors so I assume it was done correctly, but as always if something is incorrect please bring it to my attention.

As `CppParser` requires nested namespace support as well to generate the new docs, I have updated it to support nested namespaces as well as a small test case to the testing suite.


mikomikotaishi:
This is a change that doesn't actually change the usage of the library at the consumer end, so could we merge this one before 1.16.0? I think it falls under code tidiness and would probably be better suited to be merged sooner rather than later.

mikomikotaishi:
The CodeQL errors seem to not be caused by anything made by this PR, as the affected errors are not mentioning changes made in this PR, so I won't address those in this PR.

obiltschnig:
This change will also require someone to add support for nested namespace declarations to CppParser, otherwise PocoDoc will fail to generate the reference documentation.

mikomikotaishi:
@obiltschnig I have added nested namespace support and a test case now

obiltschnig:
> @obiltschnig I have added nested namespace support and a test case now

Great, thank you!

CppParser/include/Poco/CppParser/NameSpace.h
```diff
@@ -117,6 +124,7 @@ class CppParser_API NameSpace: public Symbol
 	SymbolTable _symbols;
 	AliasMap _importedSymbols;
 	NameSpaceVec _importedNameSpaces;
+	bool _isInline;
```
matejk: The `_isInline` member has no default member initializer and is assigned in the constructor body. Consider using a default member initializer here:
```cpp
bool _isInline = false;
```
This avoids the risk of uninitialized reads if a new constructor is added later without initializing it, and allows the default constructor body to be empty.

CppParser/src/NameSpace.cpp
```diff
@@ -22,18 +22,19 @@ using Poco::ExistsException;
...
 NameSpace::NameSpace()
 {
+	_isInline = false;
 }
 
 
-NameSpace::NameSpace(const std::string& name, NameSpace* pNameSpace):
+NameSpace::NameSpace(const std::string& name, NameSpace* pNameSpace, bool isInline):
 	Symbol(name, pNameSpace)
 {
```
matejk, line 36: Prefer the member initializer list over assignment in the constructor body:
```cpp
NameSpace::NameSpace(const std::string& name, NameSpace* pNameSpace, bool isInline):
	Symbol(name, pNameSpace),
	_isInline(isInline)
{
}
```
Similarly, the default constructor can be simplified if `_isInline` has a default member initializer in the header.

CppParser/src/Parser.cpp
```diff
@@ -239,14 +267,31 @@ const Token* Parser::parseNameSpace(const Token* pNext)
...
+			if (undefined) 
+				pNS = new NameSpace(namespaceNames[i], currentNameSpace(), inlineFlags[i]);
+			else if (inlineFlags[i])
+			{
+				// If the namespace already exists but this declaration marks it as inline,
+				// update the inline flag
+				pNS->setInline(inlineFlags[i]);
+			}
+			// Only add the first (outermost) namespace to the global symbol table,
+			// and only if it's not already there
```
matejk: The `addToGST` logic has subtly changed from the old behavior.

**Old code** set `addGST = undefined` (true when the namespace was newly created, false when found via `lookup`).

**New code** checks `_gst.find(namespaceNames[i]) == _gst.end()`, which looks up the **simple name** in the flat global symbol table.

Since `_gst` is a `multimap`, `find()` will match **any** symbol with that name (not just namespaces), which could give false negatives if a non-namespace symbol happens to share the name. Consider using the original approach:
```cpp
pushNameSpace(pNS, -1, (i == 0 && undefined));
```
This preserves the old semantic: add to GST only when the outermost namespace was newly created.

CppParser/include/Poco/CppParser/NameSpace.h
```diff
@@ -99,6 +98,14 @@ class CppParser_API NameSpace: public Symbol
...
+	[[nodiscard]]
+	bool isInline() const;
+		/// Returns true if this is an inline namespace.
+
+	void setInline(bool isInline);
+		/// Sets the inline flag for this namespace.
+
+	[[nodiscard]]
 	Symbol::Kind kind() const;
 	std::string toString() const;
```
matejk, line 110: Nit: `[[nodiscard]]` was added to `isInline()` and `kind()` but not to `toString()` on the next line. Consider adding it for consistency, or removing from `kind()` since the existing codebase doesn't use `[[nodiscard]]` elsewhere in this class.

CppParser/testsuite/src/CppParserTest.cpp
```diff
@@ -417,6 +417,87 @@ class Foo
...
+	assertTrue (!pNSC->isInline());
+	
+	// Check that func exists within C
+	Symbol* pSymFunc = pNSC->lookup("func");
+	assertTrue (pSymFunc != nullptr);
+	assertTrue (pSymFunc->fullName() == "A::B::C::func");
+}
+
+
 void CppParserTest::setUp()
```
matejk, line 599: Good test coverage for the basic cases. Consider adding tests for:
- **Repeated declaration** of the same nested namespace (e.g., `namespace A::B { class X; }` followed by `namespace A::B { class Y; }`) to ensure both symbols end up in the same namespace
- **Single (non-nested) namespace** as a regression test for the refactored code
- **Deeply nested** (4+ levels) to verify the pop loop works correctly

matejk:
Hi @mikomikotaishi, I opened new issues for CodeQL issues to be analysed separately.

Please take a look at the comment for CppParser. Thank you for your contributions.

CppParser/include/Poco/CppParser/NameSpace.h
mikomikotaishi: I probably would have had it added in my other pull request for `[[nodiscard]]`, #5156. If you want I could just move ahead with adding it here, rather than waiting for #5156.

Edit: I think it's a good idea if we move ahead with it here, rather than waiting for #5156, seeing as it's a related functionality that touches this class.

CppParser/src/NameSpace.cpp
mikomikotaishi: Ah, good catch, fixed it.

CppParser/include/Poco/CppParser/NameSpace.h
mikomikotaishi: Good idea, I'll do this instead.

CppParser/src/Parser.cpp
mikomikotaishi: OK, I'll apply this

CppParser/testsuite/src/CppParserTest.cpp
mikomikotaishi: OK, added these.

CppParser/include/Poco/CppParser/NameSpace.h
matejk: Yes, move this to #5156.

mikomikotaishi: I will keep the one method `NameSpace::isInline()` marked as `[[nodiscard]]`, as #5156 doesn't have that method, but leave the rest to #5156 as requested, then.

dependencies/tessil/include/Poco/ordered_hash.h
```diff
@@ -117,7 +115,7 @@ const T& clamp(const T& v, const T& lo, const T& hi) {
 }
 
 template <typename T, typename U>
-static T numeric_cast(U value,
+T numeric_cast(U value,
```
matejk, line 120: This file lives under `dependencies/tessil/` and should not be modified as part of POCO changes — it's a bundled third-party library. If this `static` removal is needed, it should be done upstream or tracked as a separate dependency patch. Please revert this change.

matejk: Non-Poco files were extracted to dependencies on main and this change can't be applied directly.

mikomikotaishi: This is the only way I can get modules to work, since apparently something calls these internally linked symbols.

I'm not actually sure what to do instead here

matejk: My proposal:

1. modify ordered_hash.h in dependencies, Write a note to dependencies/README.md about the change and the reason for the change.
2. Then open PR in https://github.com/Tessil/ordered-map for the same change to get it included upstream.

mikomikotaishi: OK, I submitted a similar pull request at Tessil/ordered-map#54.

Do note that the repo has a few stale PRs, the most recent one being in May of last year, which haven't been responded to at all, though the last commit to the repo was from the author 4 months ago.

matejk:
@mikomikotaishi, please do not force push because the track of changes is then lost. We will squash merge in the end after all of the changes are done.

dependencies/tessil/include/Poco/ordered_hash.h
matejk: Some stable products change very slowly. We'll see what happens.

dependencies/cpptrace/src/symbols/dwarf/debug_map_resolver.cpp
```diff
@@ -19,8 +19,7 @@
 #include <vector>
 
 CPPTRACE_BEGIN_NAMESPACE
-namespace detail {
-namespace libdwarf {
+namespace detail::libdwarf {
```
matejk: Revert all changes in cpptrace. We use it unchanges from upstream.

dependencies/v8_double_conversion/src/cached-powers.h
```diff
@@ -30,9 +30,7 @@
 
 #include "diy-fp.h"
 
-namespace double_conversion {
-
-namespace PowersOfTenCache {
+namespace double_conversion::PowersOfTenCache {
```
matejk: Revert all changes in v8_double_conversion. We use it unchanged from upstream.

mikomikotaishi:
I only use force pushes when it's a small amendment, such as a typo or a minor fix to a broken push.

However that being said I will avoid doing this in the future.

dependencies/cpptrace/src/symbols/dwarf/debug_map_resolver.cpp
mikomikotaishi: Fixed

dependencies/v8_double_conversion/src/cached-powers.h
mikomikotaishi: Fixed

mikomikotaishi:
Sorry for the delay in the response, I didn't see it until now. I've reverted all changes made in `dependencies` (except for removing `static` from `dependencies/tessil/include/Poco/ordered_hash.h` and the notice about this in the readme file).

CppParser/src/Parser.cpp
```diff
@@ -223,11 +222,40 @@ const Token* Parser::parseNameSpace(const Token* pNext)
...
+				syntaxError("namespace name after ::");
+			}
+			
+			namespaceNames.push_back(pNext->tokenString());
+			inlineFlags.push_back(isInline);
+			pNext = next();
+		}
 
 		if (isOperator(pNext, OperatorToken::OP_ASSIGN))
 		{
```
mikomikotaishi, line 257: I just want to ensure I am not misunderstanding: should this not permit re-defining/re-opening namespaces? In the tests, I re-open namespaces by re-using them, and am not sure if this is the intended behaviour or not.

matejk: Sorry, the comment was wrong.

CppParser/src/Parser.cpp
```diff
@@ -239,14 +267,23 @@ const Token* Parser::parseNameSpace(const Token* pNext)
...
+			bool undefined = (pNS == nullptr);
+			if (undefined) 
+				pNS = new NameSpace(namespaceNames[i], currentNameSpace(), inlineFlags[i]);
+			else if (inlineFlags[i])
+			{
+				// If the namespace already exists but this declaration marks it as inline,
+				// update the inline flag
+				pNS->setInline(inlineFlags[i]);
+			}
+			pushNameSpace(pNS, -1, i == 0);
```
matejk, line 283: I think that here it should be:

```
pushNameSpace(pNS, -1, (i == 0 && undefined))
```

matejk:
There is potentially one change in CppParser. Please take a look.

Rebase (or merge) from latest main in Poco repository to resolve conflicts.

Then we are ready for merge.

CppParser/src/Parser.cpp
mikomikotaishi: I just want to ensure I am not misunderstanding: should this not permit re-defining/re-opening namespaces? In the tests, I re-open namespaces by re-using them, and am not sure if this is the intended behaviour or not. You mentioned earlier that you made a mistake in the comment you left, but this was on line 257. Could you give clarity on what the intended behaviour should be, and whether my tests should reuse namespaces?

matejk: Re-opening namespaces is perfectly fine and must keep working — that is not the concern here.

The issue is about **duplicate entries in `_gst`** (the global symbol table, a `std::multimap`), not about preventing namespace re-opening.

**Old code** (before this PR):
```cpp
NameSpace* pNS = dynamic_cast<NameSpace*>(currentNameSpace()->lookup(fullName));
bool undefined = (pNS == nullptr);
if (undefined) pNS = new NameSpace(name, currentNameSpace());
pushNameSpace(pNS, -1, undefined);  // addGST = true ONLY when newly created
```

When `namespace Poco` appeared a second time in the same file, `undefined` was `false`, so the existing `Poco` namespace was **not** inserted into `_gst` again.

**Current code:**
```cpp
bool undefined = (pNS == nullptr);
// ...
pushNameSpace(pNS, -1, i == 0);  // addGST = true ALWAYS for outermost
```

Now every time `namespace Poco::Net` appears (even the 2nd, 3rd time in the same file), the outermost namespace `Poco` is re-inserted into `_gst`. Since `_gst` is a multimap, this creates duplicate entries.

**The fix** `(i == 0 && undefined)` restores the original semantic: add to `_gst` only when the namespace is seen for the first time. Re-opening still works exactly as before — `lookup()` finds the existing namespace, symbols get added to it — we just skip the redundant `_gst` insert.

mikomikotaishi: OK, I'll apply the change then.

mikomikotaishi: Making this change seemed to cause the unit tests to fail, particularly at the line `st.find("A") != st.end()`. Any ideas on what to do?

matejk: My suggestion was actually wrong and should be reverted.

It does not work because there is a global root() singleton which is shared across runs in tests.

A apologise.

mikomikotaishi: OK, I've gone ahead and undid the edit.

Should we wait for the tests to finish before merging, or could we just go ahead and do it knowing that two commits ago the tests all passed?

matejk: I'll merge immediately after 1.15.1 is released.

mikomikotaishi: Will that be soon? I am just concerned that if pull requests continue they will have merge conflicts, which are obviously easy to resolve, but, still.

dependencies/tessil/include/Poco/ordered_hash.h
mikomikotaishi: They finally merged the PR in now, so I suppose there's no worry now about things breaking with the module.

matejk: This is good news. I'll sync the upstream code.

Issue #5237, opened 2026-03-09
OpenSSL constifying interfaces and removing direct access from ASN1_STRING types


seektechnz:
**Describe the bug**
OpenSSL has changed its interfaces. Most X509 pointers are now const and ASN1_STRING types cannot have direct member access.

**To Reproduce**
Build Poco with OpenSSL built from latest git source.

**Expected behavior**
Building with no type mismatch errors.

**Logs**
```
C:\Users\build\externals\poco\Crypto\src\PKCS12Container.cpp(131,19): error C2440: 'reinterpret_cast': cannot convert from 'const unsigned char *' to 'char *' [C:\Users\build\externals\poco\output\x64\Crypto\Crypto.vcxproj]
     C:\Users\build\externals\poco\Crypto\src\PKCS12Container.cpp(131,19):
     Conversion loses qualifiers

C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(226,16): error C2664: 'std::string Poco::Crypto::_X509_NAME_oneline_utf8(X509_NAME *)': cannot convert argument 1
from 'const X509_NAME *' to 'X509_NAME *' [C:\Users\build\externals\poco\output\x64\Crypto\Crypto.vcxproj]
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(226,60):
     Conversion loses qualifiers
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(212,13):
     see declaration of 'Poco::Crypto::_X509_NAME_oneline_utf8'
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(226,16):
     while trying to match the argument list '(const X509_NAME *)'

C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(227,17): error C2664: 'std::string Poco::Crypto::_X509_NAME_oneline_utf8(X509_NAME *)': cannot convert argument 1
from 'const X509_NAME *' to 'X509_NAME *' [C:\Users\build\externals\poco\output\x64\Crypto\Crypto.vcxproj]
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(227,62):
     Conversion loses qualifiers
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(212,13):
     see declaration of 'Poco::Crypto::_X509_NAME_oneline_utf8'
     C:\Users\build\externals\poco\Crypto\src\X509Certificate.cpp(227,17):
[...]

PR #5239, opened 2026-03-10
Constify X509* usage where needed and use ASN1_STRING accessors

seektechnz:
Fixes: https://github.com/pocoproject/poco/issues/5237

OpenSSL is moving most X509 pointers to const, this makes handlers compatible with const pointers. OpenSSL has also begun removing direct access to ASN1_STRING members, so we use the accessor functions instead of direct struct members.

### DIFF
--- a/Crypto/src/X509Certificate.cpp
+++ b/Crypto/src/X509Certificate.cpp
@@ -298,7 +298,7 @@ Poco::DateTime X509Certificate::validFrom() const
 {
 	const ASN1_TIME* certTime = X509_get0_notBefore(_pCert);
 	auto certTimeType = ASN1_STRING_type(certTime);
-	std::string dateTime(reinterpret_cast<const char*>(ASN1_STRING_get0_data(certTime)));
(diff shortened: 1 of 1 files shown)


matejk:
Nice, clean PR — the changes are correct, minimal, and backward-compatible with OpenSSL >= 1.1.0.

One suggestion: in `validFrom()` and `expiresOn()`, the `std::string` is constructed from `ASN1_STRING_get0_data()` relying on null-termination:

```cpp
std::string dateTime(reinterpret_cast<const char*>(ASN1_STRING_get0_data(certTime)));
```

It would be more robust to use the length accessor as well:

```cpp
std::string dateTime(reinterpret_cast<const char*>(ASN1_STRING_get0_data(certTime)), ASN1_STRING_length(certTime));
```

This is a pre-existing issue (the old code did the same with `certTime->data`), so not a blocker for this PR, but worth considering as part of this change since you're already touching these lines.

seektechnz:
That should be fixed now for safety. Thanks for looking it over!

PR #5254, opened 2026-03-19
add include directive support to PropertyFileConfiguration

aleks-f:
Relative paths are resolved relative to the directory of the including file. Nested includes are supported. A missing include file throws Poco::FileException.

Also adds two entries to XML/.gitignore (winconfig.h, xmlparse.c).

### DIFF
--- a/Util/src/PropertyFileConfiguration.cpp
+++ b/Util/src/PropertyFileConfiguration.cpp
@@ -179,20 +179,18 @@ void PropertyFileConfiguration::parseLine(std::istream& istr, const std::string&
 					throw Poco::FileException("Cyclic property file include detected", absPathStr);
 				}
 
-				includeStack.insert(absPathStr);
-				try
+				struct StackGuard
 				{
-					Poco::FileInputStream includeIstr(p.toString());
-					if (!includeIstr.good())
-						throw Poco::OpenFileException(p.toString());
-					loadStream(includeIstr, p.parent().toString(), includeStack);
-				}
-				catch (...)
-				{
-					includeStack.erase(absPathStr);
-					throw;
-				}
-				includeStack.erase(absPathStr);
+					std::set<std::string>& stack;
+					const std::string& key;
+					StackGuard(std::set<std::string>& s, const std::string& k): stack(s), key(k) { stack.insert(k); }
+					~StackGuard() { stack.erase(key); }
+				} guard(includeStack, absPathStr);
+
+				Poco::FileInputStream includeIstr(p.toString());
+				if (!includeIstr.good())
+					throw Poco::OpenFileException(p.toString());
+				loadStream(includeIstr, p.parent().toString(), includeStack);
 			}
 		}
 		else


Util/src/PropertyFileConfiguration.cpp
```diff
@@ -46,24 +46,38 @@ void PropertyFileConfiguration::load(std::istream& istr)
...
 	Poco::FileInputStream istr(path);
-	if (istr.good())
-		load(istr);
-	else
+	if (!istr.good())
 		throw Poco::OpenFileException(path);
+	AbstractConfiguration::ScopedLock lock(*this);
+	clear();
+	std::set<std::string> includeStack;
+	loadStream(istr, basePath, includeStack);
```
matejk, line 67: **Bug:** the root file's absolute path is never inserted into `includeStack` before this call. A file that includes itself (`!include self.properties` where `self.properties` is the file being loaded) will **not** be caught by the cycle detector and will recurse until stack overflow.

Fix: insert the root file's absolute path into `includeStack` before calling `loadStream()`:
```cpp
includeStack.insert(p.toString()); // p is already absolute at this point
loadStream(istr, basePath, includeStack);
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -121,7 +135,7 @@ void PropertyFileConfiguration::save(const std::string& path) const
 }
 
 
-void PropertyFileConfiguration::parseLine(std::istream& istr)
+void PropertyFileConfiguration::parseLine(std::istream& istr, const std::string& basePath, std::set<std::string>& includeStack)
 {
 	static const int eof = std::char_traits<char>::eof();
```
matejk: `static` is unnecessary here — the initializer is a constexpr expression. Should be:
```cpp
constexpr int eof = std::char_traits<char>::eof();
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -131,7 +145,46 @@ void PropertyFileConfiguration::parseLine(std::istream& istr)
 	{
 		if (c == '#' || c == '!')
 		{
-			while (c != eof && c != '\n' && c != '\r') c = istr.get();
+			std::string line;
+			line += (char) c;
+			while (c != eof && c != '\n' && c != '\r') { c = istr.get(); if (c != eof && c != '\n' && c != '\r') line += (char) c; }
```
matejk: This line is very dense and hard to follow — multiple statements with duplicated conditions crammed onto one line. Consider a normal multi-line loop:
```cpp
while (c != eof && c != '\n' && c != '\r')
{
    c = istr.get();
    if (c != eof && c != '\n' && c != '\r')
        line += static_cast<char>(c);
}
```
(Also: `(char) c` is a C-style cast — `static_cast<char>(c)` is preferred in new code.)

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -131,7 +145,46 @@ void PropertyFileConfiguration::parseLine(std::istream& istr)
...
+
+			static const std::string includeDirective = "!include";
```
matejk: `static const std::string` allocates on the heap on first call. Since C++17, this can be a zero-cost constant:
```cpp
constexpr std::string_view includeDirective = "!include";
```
`line.compare(0, ...)` works fine with `string_view::size()`.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -131,7 +145,46 @@ void PropertyFileConfiguration::parseLine(std::istream& istr)
 	{
 		if (c == '#' || c == '!')
 		{
-			while (c != eof && c != '\n' && c != '\r') c = istr.get();
+			std::string line;
```
matejk, line 154: **Performance:** when the first character is `#` (not `!`), there's no need to build the `line` string at all — it will just be discarded. The string is only useful for `!include` detection. Consider skipping the string building for `#` comments:
```cpp
if (c == '#')
{
    while (c != eof && c != '\n' && c != '\r') c = istr.get();
}
else if (c == '!')
{
    // build line and check for !include ...
}
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -131,7 +145,46 @@ void PropertyFileConfiguration::parseLine(std::istream& istr)
...
+				Poco::Ascii::isSpace(line[includeDirective.size()]))
+			{
+				std::string includePath = Poco::trim(line.substr(includeDirective.size()));
+				if (includePath.empty())
+					throw Poco::SyntaxException("Missing path in !include directive");
+				Poco::Path p(includePath);
+				if (p.isRelative() && !basePath.empty())
+					p = Poco::Path(basePath).resolve(p);
+
+				Poco::Path absPath(p);
```
matejk: The `Poco::Path absPath(p)` copy is redundant — you could call `p.makeAbsolute()` directly since `p` is not used in its relative form after this point (its only subsequent uses on lines 176 and 179 would work equally well with an absolute path). This would also eliminate the extra `Poco::Path` construction.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -131,7 +145,46 @@ void PropertyFileConfiguration::parseLine(std::istream& istr)
...
+				absPath.makeAbsolute();
+				const std::string absPathStr = absPath.toString();
+
+				if (includeStack.find(absPathStr) != includeStack.end())
+				{
+					throw Poco::FileException("Cyclic property file include detected", absPathStr);
+				}
+
+				includeStack.insert(absPathStr);
```
matejk: The manual insert/try/catch/erase pattern for `includeStack` is fragile. A simple scope guard would make this exception-safe by construction:
```cpp
includeStack.insert(absPathStr);
auto guard = Poco::ScopeGuard([&] { includeStack.erase(absPathStr); });
// ... no try/catch needed
```
(Or if `ScopeGuard` isn't available here, even a small RAII struct would be cleaner.)

aleks-f: f93a6a6e7

Util/src/PropertyFileConfiguration.cpp
aleks-f: d5fca1c2d

Util/src/PropertyFileConfiguration.cpp
aleks-f: 63e8e4d4f

Util/src/PropertyFileConfiguration.cpp
aleks-f: b02ab30b1

Util/src/PropertyFileConfiguration.cpp
aleks-f: 4235538eb

Util/src/PropertyFileConfiguration.cpp
aleks-f: 7f740e2df

Util/src/PropertyFileConfiguration.cpp
aleks-f: 9594a6bd5

Issue #5257, opened 2026-03-21
Preserve properties file on save


aleks-f:
**Is your feature request related to a problem? Please describe.**
Properties file saving does not preserve comments, includes and the original order of properties.

**Describe the solution you'd like**
The original file comments, includes and properties order should be preserved.

matejk:
Closed by #5258

andrewauclair:
Is this feature always enabled? To prevent this extra save logic from happening do I have to call `PropertyFileConfiguration(std::istream& istr);` instead of `PropertyFileConfiguration(const std::string& path);`?

I'm concerned this change is going to cause some serious performance issues due to the `std::map<std::string, std::map<std::string, std::string>> fileValues` and building up that map of maps every single call to save.

aleks-f:
It is currently always enabled when the configuration is loaded from a file. The reason (besides the obvious ones: lost comments, reordered entries and merged included files resulting in duplication of configuration content) is that I did not want want to break existing interface for this release. I would not be opposed to adding a bool flag controlling this behavior, defaulting to true.

andrewauclair:
Wouldn't it be better to default to false and let users opt-in? I've never even thought to have the property files preserve comments and order. And I wonder how many users of Poco are the same.

aleks-f:
It is best not to break released behavior if possible. As for the comments, they serve a purpose; removing them without explicit request to do so is not a good behavior.

matejk:
An enhancement can be planned for 1.16.0.

aleks-f:
> An enhancement can be planned for 1.16.0.

That's fine with me, I just want to make clear that (a) saving properties to the file from which they came from, (b) not reordering properties, and (c) not removing comments must remain default behavior because it is common sense.

andrewauclair:
I did some crude timing tests, and the performance hit seems to be fairly minimal.

PR #5258, opened 2026-03-21
feat(PropertyFileConfiguration): preserve original file contents and …

aleks-f:
…layout #5257

### DIFF
--- a/Util/include/Poco/Util/PropertyFileConfiguration.h
+++ b/Util/include/Poco/Util/PropertyFileConfiguration.h
@@ -116,8 +116,13 @@ public:
 		/// Sets the source file for the given key. When save() is called
 		/// with provenance-based saving, the key will be written to this file.
 
+	void clear();
+		/// Clears the configuration, including provenance information.
+
 protected:
 	~PropertyFileConfiguration() = default;
+
+	void removeRaw(const std::string& key) override;
 
 private:
 	void loadStream(std::istream& istr, const std::string& basePath, const std::string& currentFile, std::set<std::string>& includeStack);
--- a/Util/src/PropertyFileConfiguration.cpp
+++ b/Util/src/PropertyFileConfiguration.cpp
@@ -57,8 +57,6 @@ void PropertyFileConfiguration::load(std::istream& istr)
 	AbstractConfiguration::ScopedLock lock(*this);
 
 	clear();
-	_sourceMap.clear();
-	_rootFile.clear();
 	std::set<std::string> includeStack;
 	loadStream(istr, "", "", includeStack);
 }
@@ -75,7 +73,6 @@ void PropertyFileConfiguration::load(const std::string& path)
 		throw Poco::OpenFileException(path);
 	AbstractConfiguration::ScopedLock lock(*this);
 	clear();
-	_sourceMap.clear();
 	_rootFile = absPath;
 	std::set<std::string> includeStack;
 	includeStack.insert(absPath);
@@ -102,37 +99,10 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
 {
 	AbstractConfiguration::ScopedLock lock(*this);
 
-	MapConfiguration::iterator it = begin();
-	MapConfiguration::iterator ed = end();
-	while (it != ed)
+	for (auto it = begin(); it != end(); ++it)
 	{
-		ostr << it->first << ": ";
-		for (auto ch: it->second)
-		{
-			switch (ch)
-			{
-			case '\t':
-				ostr << "\\t";
-				break;
-			case '\r':
-				ostr << "\\r";
-				break;
-			case '\n':
-				ostr << "\\n";
-				break;
-			case '\f':
-				ostr << "\\f";
-				break;
-			case '\\':
-				ostr << "\\\\";
-				break;
-			default:
-				ostr << ch;
-				break;
-			}
-		}
-		ostr << "\n";
-		++it;
+		const auto& [key, value] = *it;
+		ostr << key << ": " << escapeValue(value) << "\n";
 	}
 }
 
@@ -162,22 +132,22 @@ void PropertyFileConfiguration::save(const std::string& path) const
 			}
 
 			// Assign keys with no provenance (newly added) to root file
-			MapConfiguration::iterator it = begin();
-			MapConfiguration::iterator ed = end();
-			while (it != ed)
+			for (auto it = begin(); it != end(); ++it)
 			{
-				if (_sourceMap.find(it->first) == _sourceMap.end())
-					fileValues[_rootFile][it->first] = it->second;
-				++it;
+				const auto& [key, value] = *it;
+				if (_sourceMap.find(key) == _sourceMap.end())
+					fileValues[_rootFile][key] = value;
 			}
 		}
 		else
 		{
 			// Flat save — all keys to the provided path
-			auto& values = fileValues[path];
-			MapConfiguration::iterator it = begin();
-			MapConfiguration::iterator ed = end();
-			while (it != ed) { values[it->first] = it->second; ++it; }
+			auto& vals = fileValues[path];
+			for (auto it = begin(); it != end(); ++it)
+			{
+				const auto& [key, value] = *it;
+				vals[key] = value;
+			}
 		}
 	}
 
@@ -224,10 +194,7 @@ void PropertyFileConfiguration::saveToFile(const std::string& path, const std::m
 			}
 
 			// Key-value line: extract key
-			std::string::size_type sep = line.find('=');
-			std::string::size_type sep2 = line.find(':');
-			if (sep2 != std::string::npos && (sep == std::string::npos || sep2 < sep))
-				sep = sep2;
+			std::string::size_type sep = line.find_first_of("=:");
 
 			if (sep != std::string::npos)
 			{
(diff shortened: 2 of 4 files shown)


Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+{
+	std::set<std::string> written;
+	std::ostringstream out;
+
+	Poco::File f(path);
+	if (f.exists())
+	{
+		Poco::FileInputStream istr(path);
+		std::string line;
+		while (std::getline(istr, line))
```
matejk, line 172: **Bug (HIGH): Multi-line continuation values break on provenance save**

`saveToFile` reads the original file line-by-line with `std::getline`. If the original file uses trailing `\\` for line continuation:
```
key = value part1 \\
value part2
```
Line 1 is matched as key `key` and rewritten with the full (already-joined) value. Line 2 has no separator, so it falls through to "preserve as-is". After save:
```
key = full joined value
value part2
```
Line 2 is a stale leftover that will be parsed as a bare key on next load, corrupting the configuration.

The line-by-line parser in `saveToFile` needs to handle continuation lines (detect trailing `\\` and consume following lines as part of the same key-value entry), or at least skip them.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+		return it->second;
+	return "";
+}
+
+
+void PropertyFileConfiguration::setSourceFile(const std::string& key, const std::string& path)
+{
+	Poco::Path p(path);
+	p.makeAbsolute();
+	_sourceMap[key] = p.toString();
```
matejk, line 280: **Bug (HIGH): Missing lock protection**

`setSourceFile` modifies `_sourceMap` without acquiring `ScopedLock`. Meanwhile, `save()` reads `_sourceMap` under the lock. Concurrent calls are a data race on `_sourceMap`.

```suggestion
void PropertyFileConfiguration::setSourceFile(const std::string& key, const std::string& path)
{
	AbstractConfiguration::ScopedLock lock(*this);
	Poco::Path p(path);
	p.makeAbsolute();
	_sourceMap[key] = p.toString();
}
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+			{
+				out << line << "\n";
+				continue;
+			}
+
+			// Key-value line: extract key
+			std::string::size_type sep = line.find('=');
+			std::string::size_type sep2 = line.find(':');
+			if (sep2 != std::string::npos && (sep == std::string::npos || sep2 < sep))
+				sep = sep2;
```
matejk: **Simplification (LOW): Use `find_first_of`**

The four-line separator detection logic can be a single call with the same semantics:

```suggestion
			std::string::size_type sep = line.find_first_of("=:");
```

Util/include/Poco/Util/PropertyFileConfiguration.h
```diff
@@ -99,16 +100,35 @@ class Util_API PropertyFileConfiguration: public MapConfiguration
...
-	void loadStream(std::istream& istr, const std::string& basePath, std::set<std::string>& includeStack);
-	void parseLine(std::istream& istr, const std::string& basePath, std::set<std::string>& includeStack);
+	void loadStream(std::istream& istr, const std::string& basePath, const std::string& currentFile, std::set<std::string>& includeStack);
+	void parseLine(std::istream& istr, const std::string& basePath, const std::string& currentFile, std::set<std::string>& includeStack);
+	static void saveToFile(const std::string& path, const std::map<std::string, std::string>& values);
 	static int readChar(std::istream& istr);
+	static std::string escapeValue(const std::string& value);
 
 	AbstractConfiguration::Ptr _pParentConfig;
+	std::map<std::string, std::string> _sourceMap;
```
matejk, line 135: **Design (LOW): `_sourceMap` leaks stale entries on key removal**

When `remove(key)` is called, `MapConfiguration::removeRaw` erases from `_map` but `_sourceMap` retains the entry. Functionally this is handled (skipped in `save` since `getRaw` returns false), but it accumulates garbage.

Consider overriding `removeRaw` to also clean up `_sourceMap`:
```cpp
void PropertyFileConfiguration::removeRaw(const std::string& key) override
{
	MapConfiguration::removeRaw(key);
	std::string prefix = key + '.';
	for (auto it = _sourceMap.begin(); it != _sourceMap.end(); )
	{
		if (it->first == key || it->first.compare(0, prefix.size(), prefix) == 0)
			it = _sourceMap.erase(it);
		else
			++it;
	}
}
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+	if (!ostr.good()) throw Poco::CreateFileException(path);
+	Poco::OutputLineEndingConverter lec(ostr);
+	lec << out.str();
+	lec.flush();
+	ostr.flush();
+	if (!ostr.good()) throw Poco::WriteFileException(path);
+}
+
+
+std::string PropertyFileConfiguration::escapeValue(const std::string& value)
```
matejk, line 246: **Maintainability (LOW): Duplicated escaping logic**

The inline `switch` in `save(ostream&)` (lines 101-136 above) duplicates this `escapeValue()` method. If either is updated without the other, they'll diverge. Consider refactoring `save(ostream&)` to use `escapeValue()`:

```cpp
ostr << it->first << ": " << escapeValue(it->second) << "\n";
```

This replaces the entire inner loop and delegates to this already-existing method.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+}
+
+
+std::string PropertyFileConfiguration::getSourceFile(const std::string& key) const
+{
+	auto it = _sourceMap.find(key);
+	if (it != _sourceMap.end())
+		return it->second;
+	return "";
+}
```
matejk, line 272: **Bug (HIGH): Missing lock protection**

Same issue as `setSourceFile` — `getSourceFile` reads `_sourceMap` without a lock.

```suggestion
std::string PropertyFileConfiguration::getSourceFile(const std::string& key) const
{
	AbstractConfiguration::ScopedLock lock(*this);
	auto it = _sourceMap.find(key);
	if (it != _sourceMap.end())
		return it->second;
	return "";
}
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -214,7 +373,10 @@ void PropertyFileConfiguration::parseLine(std::istream& istr, const std::string&
 				c = readChar(istr);
 				while (c != eof && c) { value += (char) c; c = readChar(istr); }
 			}
-			setRaw(trim(key), trim(value));
+				std::string trimmedKey = trim(key);
```
matejk: **Style (LOW): Indentation error**

This line has one extra level of indentation compared to the `setRaw` call on the next line.

```suggestion
			std::string trimmedKey = trim(key);
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+
+			// Assign keys with no provenance (newly added) to root file
+			MapConfiguration::iterator it = begin();
+			MapConfiguration::iterator ed = end();
+			while (it != ed)
+			{
+				if (_sourceMap.find(it->first) == _sourceMap.end())
+					fileValues[_rootFile][it->first] = it->second;
+				++it;
+			}
```
matejk: **Modernization (LOW): Use structured bindings**

```suggestion
			for (auto it = begin(); it != end(); ++it)
			{
				const auto& [key, value] = *it;
				if (_sourceMap.find(key) == _sourceMap.end())
					fileValues[_rootFile][key] = value;
			}
```

Same applies to the flat-save loop at lines 178-180.

(`begin()`/`end()` are `protected` in `MapConfiguration`, so range-for on `*this` won't compile — structured bindings on `*it` within a traditional for loop is the best available option.)

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -55,28 +57,33 @@ void PropertyFileConfiguration::load(std::istream& istr)
 	AbstractConfiguration::ScopedLock lock(*this);
 
 	clear();
+	_sourceMap.clear();
```
matejk: **Bug (MEDIUM): `clear()` doesn't reset provenance state**

`MapConfiguration::clear()` clears `_map` but does not clear `_sourceMap` or `_rootFile`. Only `load()` does that (here). If a user calls `clear()` directly (without `load()`), then manually adds keys via `setString()` and calls `save()`, stale provenance data is used.

Consider overriding `clear()` in `PropertyFileConfiguration`:
```cpp
void PropertyFileConfiguration::clear()
{
	MapConfiguration::clear();
	_sourceMap.clear();
	_rootFile.clear();
}
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -132,20 +139,172 @@ void PropertyFileConfiguration::save(std::ostream& ostr) const
...
+				// Line with no separator — preserve as-is
+				out << line << "\n";
+			}
+		}
+	}
+
+	// Append new keys not yet written
+	for (const auto& [key, value] : values)
+	{
+		if (!written.count(key))
```
matejk, line 230: **Modernization (LOW):** `count` → `contains` (C++20, default since POCO 1.15.0)

```suggestion
		if (!written.contains(key))
```

matejk: Optional if we still want to be able to compile with C++17 standard.

Issue #5266, opened 2026-03-25
PropertyFileConfiguration: add include file capability


aleks-f:
Add capability to add/remove `!include` directives

PR #5267, opened 2026-03-25
feat(PropertyFileConfiguration): include directive add/remove #5266

aleks-f:


Util/src/LoggingConfigurator.cpp
```diff
@@ -285,18 +367,18 @@ bool LoggingConfigurator::validateConfiguration(AbstractConfiguration::Ptr pConf
...
 	{
-		if (registry.hasChannel(c))
-			return false;
+		if (!registry.hasChannel(c))
+			return true;
 	}
 
-	return true;
+	return false;
 }
```
matejk, line 407: **Bug: edge-case regression when config has no formatters and no channels**

The old logic returned `true` when both loops found nothing (no collisions). The new inverted logic falls through to `return false` — rejecting configs with zero formatters AND zero channels (e.g. logger-only configs).

This means `getLogger()` with a logger-only config will now skip `configure()` and log a warning instead of configuring the logger.

Suggested fix — add early return at the top of the method:
```cpp
if (pFmtConfig->keys().empty() && pChConfig->keys().empty())
    return true;
```

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -308,6 +308,213 @@ void PropertyFileConfiguration::clear()
...
+		throw Poco::CreateFileException(_rootFile);
+	ostr << "!include " << path << "\n";
+	ostr.flush();
+	if (!ostr.good())
+		throw Poco::WriteFileException(_rootFile);
+}
+
+
+void PropertyFileConfiguration::removeIncludeFile(const std::string& path, bool removeKeys)
+{
```
matejk, line 461: **`addIncludeFile()` does not load included keys into memory**

This method appends the `!include` directive to the root file on disk and creates the target file if missing, but never loads the included file's properties into the in-memory configuration. After calling `addIncludeFile()`:
- `getIncludeFiles()` will list it (re-scans the file)
- But any keys in that file are invisible until the config is reloaded

The test works around this by doing `pConf = new PropertyFileConfiguration(rootFile.path())` after `addIncludeFile()`.

Suggestion: Either load the included file into memory (call `loadStream()` internally), or document clearly in the doc comment that a reload is required.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -308,6 +308,213 @@ void PropertyFileConfiguration::clear()
...
+			}
+		}
+	}
+	Poco::FileOutputStream ostr(_rootFile, std::ios::app);
+	if (!ostr.good())
+		throw Poco::CreateFileException(_rootFile);
+	ostr << "!include " << path << "\n";
+	ostr.flush();
+	if (!ostr.good())
+		throw Poco::WriteFileException(_rootFile);
```
matejk, line 456: **Mixed line endings between add and remove**

`addIncludeFile()` writes raw `"\n"` here, but `removeIncludeFile()` uses `Poco::OutputLineEndingConverter` which writes platform-native endings. On Windows, this produces LF in `addIncludeFile` and CRLF in `removeIncludeFile` within the same file.

Suggestion: Use `OutputLineEndingConverter` here too, or use raw `"\n"` in both methods.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -308,6 +308,213 @@ void PropertyFileConfiguration::clear()
...
+	{
+		// Just clear provenance for keys from this file
+		for (auto it = _sourceMap.begin(); it != _sourceMap.end(); )
+		{
+			if (it->second == absPath)
+				it = _sourceMap.erase(it);
+			else
+				++it;
+		}
+	}
```
matejk, line 547: **`removeKeys=false` silently migrates keys to root file on save**

When `removeKeys=false`, this erases `_sourceMap` entries for the included file's keys, making them unprovenanced. A subsequent `save()` will write those keys to the **root file** (since unprovenanced keys default to root), effectively migrating them out of the included file.

This may be intentional but is surprising — worth documenting in the method's doc comment.

Util/src/PropertyFileConfiguration.cpp
```diff
@@ -308,6 +308,213 @@ void PropertyFileConfiguration::clear()
...
+
+	// Append !include directive to root file
+	// Ensure file ends with a newline so the directive starts on its own line
+	{
+		Poco::FileInputStream fin(_rootFile);
+		if (fin.good())
+		{
+			fin.seekg(0, std::ios::end);
+			if (fin.tellg() > 0)
```
matejk, line 437: **Triple file open for one append operation**

This opens the root file for reading (seek to last byte, check newline), then opens it again for appending a newline, then opens it a third time to append the `!include` directive. This could be simplified to a single append open — read the last byte first and write the conditional newline + directive in one shot:
```cpp
Poco::FileOutputStream ostr(_rootFile, std::ios::app);
// check if we need a preceding newline
{
    Poco::FileInputStream fin(_rootFile);
    fin.seekg(0, std::ios::end);
    if (fin.tellg() > 0)
    {
        fin.seekg(-1, std::ios::end);
        char last = 0;
        fin.get(last);
        if (last != '\n' && last != '\r')
            ostr << "\n";
    }
}
ostr << "!include " << path << "\n";
ostr.flush();
```

Util/src/LoggingConfigurator.cpp
```diff
@@ -63,6 +66,85 @@ void LoggingConfigurator::configure(AbstractConfiguration::Ptr pConfig)
...
+	std::set<std::string> channelNames;
+	std::function<void(const std::string&)> collectChannels =
+		[&](const std::string& name)
+	{
+		if (name.empty() || channelNames.count(name)) return;
+		channelNames.insert(name);
+		std::string subChannels = pChConfig->getString(name + ".channels"s, ""s);
+		if (!subChannels.empty())
+		{
+			Poco::StringTokenizer tok(subChannels, ","s, Poco::StringTokenizer::TOK_TRIM);
```
matejk: **Minor: `std::function` heap allocation**

`std::function<void(const std::string&)>` involves a heap allocation on each call to this method. Since this is designed for runtime reconfiguration (potentially called repeatedly), consider using a private helper method instead of a lambda captured by `std::function`.

Issue #5275, opened 2026-03-30
Prometheus: add memory and thread metrics


aleks-f:
Add memory and thread metrics to Prometheus (linux only)

matejk:
@aleks-f, this issue is closed but its PR is still open. Shall PR be reviewed and merged?

PR #5276, opened 2026-03-30
feat(Prometheus): add thread and CPU metrics (Linux only) #5275

aleks-f:


Prometheus/include/Poco/Prometheus/ProcessCollector.h
```diff
@@ -36,6 +37,10 @@ class Prometheus_API ProcessCollector: public Collector
 	///     (actually, the time the Prometheus library was loaded).
 	///   - process_up_time_seconds: Up time of the process in seconds
 	///     (actually, time since the Prometheus library was loaded).
+	///   - process_max_fds: Maximum number of open file descriptors (Unix only).
+	///   - process_resident_memory_bytes: Resident memory size in bytes (Linux only).
+	///   - process_virtual_memory_bytes: Virtual memory size in bytes (Linux only).
+	///   - process_thread_cpu_seconds_total: Per-thread CPU time (Linux only).
```
matejk: Per [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/#metric-names), the `_total` suffix is reserved for Counter types. This metric is declared as a `Gauge`.

CPU time is monotonically increasing per-thread, but threads come and go, so Gauge was likely chosen intentionally. Suggestion: drop the `_total` suffix to match the Gauge type:
```
process_thread_cpu_seconds
```

Prometheus/include/Poco/Prometheus/ProcessCollector.h
```diff
@@ -61,7 +66,12 @@ class Prometheus_API ProcessCollector: public Collector
 private:
 	using MetricPtr = std::unique_ptr<Metric>;
 	std::vector<MetricPtr> _metrics;
+	std::unique_ptr<Gauge> _pThreadCPU;
```
matejk: This member is only initialized and used inside `#if POCO_OS == POCO_OS_LINUX` guards. It should be conditionally compiled too:
```cpp
#if POCO_OS == POCO_OS_LINUX
	std::unique_ptr<Gauge> _pThreadCPU;
#endif
```
This avoids a dead 8-byte member on macOS/BSD/Windows and keeps the platform scope explicit.

Prometheus/src/ProcessCollector.cpp
```diff
@@ -50,8 +57,45 @@ void ProcessCollector::exportTo(Exporter& exporter) const
...
 		p->exportTo(exporter);
 	}
+#if POCO_OS == POCO_OS_LINUX
+	exportThreadCPU(exporter);
+#endif
+}
+
+
+#if POCO_OS == POCO_OS_LINUX
+namespace {
```
matejk, line 66: Since C++17 is the minimum standard for Poco, consider using `std::string_view` for the `field` parameter. This eliminates the duplicate `std::strlen(field)` calls (lines 72 and 75) and removes the need for `<cstring>`:

```cpp
double readProcStatusKb(std::string_view field)
{
	std::ifstream in("/proc/self/status");
	if (!in) return 0.0;
	std::string line;
	while (std::getline(in, line))
	{
		if (line.compare(0, field.size(), field) == 0)
		{
			if (line.size() <= field.size()) return 0.0;
			const char* start = line.c_str() + field.size();
			char* end = nullptr;
			double val = std::strtod(start, &end);
			return (end != start) ? val : 0.0;
		}
	}
	return 0.0;
}
```

Prometheus/src/ProcessCollector.cpp
```diff
@@ -98,4 +170,79 @@ void ProcessCollector::buildMetrics()
...
+	if (ticksPerSec <= 0.0) return;
+	bool headerWritten = false;
+
+	struct DirGuard
+	{
+		DIR* d;
+		explicit DirGuard(DIR* p): d(p) {}
+		~DirGuard() { if (d) closedir(d); }
+	};
+
```
matejk: The `DirGuard` struct + raw `opendir`/`readdir`/`closedir` can be replaced with `Poco::DirectoryIterator`, which is the established abstraction for this in the codebase. Currently only `DirectoryIterator_UNIX.cpp` itself uses raw `opendir` — this PR would be the first external use, going against the existing pattern.

```cpp
#include "Poco/DirectoryIterator.h"
// ...
Poco::DirectoryIterator it("/proc/self/task");
Poco::DirectoryIterator end;
for (; it != end; ++it)
{
	const std::string& tid = it.name();
	if (tid[0] < '0' || tid[0] > '9') continue;
	// ... rest unchanged
}
```
This eliminates `DirGuard`, the `<dirent.h>` include, and ~10 lines of boilerplate.

PR #5284, opened 2026-04-01
fix(Redis): use shared_ptr for NotificationCenter, harden AsyncNotificationCenter::stop()

matejk:
## Summary

- Use `shared_ptr` internally in `Redis::Client` for the `AsyncNotificationCenter`. When `std::atomic<shared_ptr>` is available (C++20 `__cpp_lib_atomic_shared_ptr`), use it directly; otherwise protect the `shared_ptr` with a `std::mutex`. This replaces the raw-pointer-with-no-op-deleter approach that could leave callers with dangling pointers if the `Client` was destroyed first.
- Harden `AsyncNotificationCenter::stop()` to unconditionally wait for and join the dequeue thread, preventing potential issues if `stop()` is called after the thread has already exited.
- Add missing connection guard to `RedisTest::testPool`.
- Enable Redis tests in `windows-2025-msvc-cmake` CI job.

## Test plan

- [x] Foundation `AsyncNotificationCenterTest` passes (14 tests)
- [x] Redis library and test runner build cleanly
- [x] Verify on Windows CI (Redis tests should skip gracefully without a server)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -657,7 +657,7 @@ jobs:
           retry_on: any
           command: >-
             cd cmake-build;
-            ctest --output-on-failure -E "(DataMySQL)|(DataODBC)|(Redis)|(MongoDB)" -C Release
+            ctest --output-on-failure -E "(DataMySQL)|(DataODBC)|(MongoDB)" -C Release
 
   windows-2025-clang-cmake:
     runs-on: windows-2025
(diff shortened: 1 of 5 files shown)


matejk:
> Client may delete the pointer, leaving holders with a false ownership of a dangling pointer. It was a quick and dirty temporary hack until c++20 and atomic shared_ptr becomes available everywhere, but not a good idea really.
> 
> Proper solution is to use shared_ptr internally, protect it with mutex when atomic shared_ptr is not available.

Applied requested changes.

PR #5301, opened 2026-04-06
enh(Foundation): optimize numeric conversions with digit-pairs algorithm and std::to_chars/from_chars

matejk:
## Summary

Rewrites integer and floating-point conversion in `NumericString.h`/`.cpp` for
significantly improved performance, using modern C++17/C++20 techniques while
maintaining full API compatibility.

### Integer formatting (`intToStr`)

Based on the [implementation suggested by @aleks-f](https://gist.github.com/aleks-f/d2ea39a465a3ea1dde5ebcbcc1a159d2),
replaces the old division-based loop with the **two-digits-at-a-time** algorithm
using a 200-byte digit-pairs lookup table (as used by {fmt}, jemalloc, Facebook Folly).
Shift-and-mask for hex and power-of-2 bases. Writes directly into the result buffer
in reverse, then reverses in-place -- no intermediate buffer copy.

### Integer parsing (`strToInt`)

Replaced the manual digit loop with `std::from_chars` (C++17). Range-based
`(begin, end)` core overload avoids `strlen`. Thousand-separator slow path only
triggers when `from_chars` stops at a non-digit character.

### Float/double formatting (`floatToStr`, `doubleToStr`)

When `POCO_HAS_FLOAT_CHARCONV` is defined (GCC 11+, MSVC 19.24+, non-Apple
libc++ 20+, or Apple Clang with macOS 26.0+ deployment target), uses
`std::to_chars` for shortest representation and `std::to_chars(fixed) +
adjustPrecision` for fixed-precision formatting. Falls back to bundled
double-conversion library on older compilers.

To enable on Apple Clang, set the deployment target:
\`\`\`bash
cmake -B build -DCMAKE_OSX_DEPLOYMENT_TARGET=26.0
\`\`\`

**Caveat**: `std::to_chars(fixed)` + manual precision adjustment is slower than
double-conversion's integrated `ToFixed` for fixed-precision formatting on macOS
Apple Clang (0.6x for `format(double, prec=2)`). This is a tradeoff: the
`std::to_chars` path eliminates ~6800 lines of bundled double-conversion dependency
when the feature is available. As stdlib implementations improve, this gap should
close. The double-conversion fallback remains available for older compilers where
it is the only option.

### Float/double parsing (`strToFloat`, `strToDouble`)

When `POCO_HAS_FLOAT_CHARCONV` is defined, uses `std::from_chars` with manual
inf/nan handling (matching double-conversion's exact-match semantics).
Falls back to `strtod` for out-of-range values (extreme underflow/overflow)
where `from_chars` reports `errc::result_out_of_range`.

### Other changes

- `POCO_HAS_FLOAT_CHARCONV` feature detection macro in `Config.h`
- Float/double functions in `NumericString.h` remain `Foundation_API` exported
  (required by inline `NumberFormatter` methods compiled into other libraries)
- `if constexpr` in `isIntOverflow` and `safeMultiply` (C++17 consistency)
- `[[nodiscard]]` on `strToInt`, `intToStr`, `safeMultiply`
- `intToStr` uses `size` parameter as buffer capacity for bounds checking
- `poco_bugcheck()` in unreachable switch default
- Removed obsolete MSVC `#pragma warning(disable: 4146)`
- Moved `<cmath>` from header to .cpp, removed unused `<memory>`
- `tryParseHex`/`tryParseHex64` now skip leading whitespace before 0x prefix
- Benchmarks rewritten to use `NumberFormatter`/`NumberParser` public API with
  baselines (std::to_string, snprintf, strtol, strtod, sscanf) and round-trip
  correctness validation

### NumberFormatter -- Integer

| Operation | macOS speedup | Linux speedup |
|---|---|---|
| format(small 0-999) | 1.2x | 1.2x |
| format(int) | **1.7x** | **2.3x** |
| format(int, w=15) | **1.5x** | **1.9x** |
| format(-int) | **1.5x** | **1.9x** |
| formatHex(int) | **1.4x** | **1.7x** |
| format(Int64) | **2.7x** | **1.7x** |
| format(UInt64) | **2.3x** | **1.6x** |
| formatHex(Int64) | **2.3x** | 1.4x |

### NumberFormatter -- Float/Double

| Operation | macOS speedup | Linux speedup |
|---|---|---|
| format(float) shortest | **1.8x** | **2.5x** |
| format(double) shortest | ~same | ~same |
| format(double, prec=2) | **0.6x** (regression) | 0.9x |
| format(double, w=15, prec=4) | ~same | ~same |

### NumberParser -- Integer

| Operation | macOS speedup | Linux speedup |
|---|---|---|
| parse(uint dec) | **1.4x** | **2.2x** |
| parseHex(uint) | 1.2x | **3.7x** |
| parseUnsigned64 | **1.9x** | **2.8x** |

### NumberParser -- Float/Double

| Operation | macOS speedup | Linux speedup |
|---|---|---|
| tryParseFloat(double) | 1.2x | **1.5x** |

### vs standard library

| vs snprintf | Integer | Float |
|---|---|---|
| macOS | **2.5-4x faster** | **2-3x faster** |
| Linux | **2.8-4.3x faster** | **2.5-4.3x faster** |

| vs sscanf | Integer | Float |
|---|---|---|
| macOS | **3-5x faster** | **1.9x faster** |
| Linux | **3-5x faster** | **2.7x faster** |

## Known regressions

- `format(double, prec=2)` on macOS: 0.6x vs main. Root cause: `std::to_chars(fixed)` +
  `adjustPrecision` is slower than double-conversion's `ToFixed`. This is the cost of
  eliminating the double-conversion dependency when `POCO_HAS_FLOAT_CHARCONV` is available.
  On compilers without charconv support, the double-conversion path is unchanged.

## Test plan

- [x] Foundation tests: 946 pass on macOS C++20, macOS C++17 (fallback), Linux GCC
- [x] JSON tests: 48 pass on macOS and Linux (including extreme underflow 123e-10000000)
- [x] Benchmarks on macOS arm64 and Linux arm64 with 5M random values
- [x] Round-trip correctness validation (format -> parse, parse -> strtol/strtod)

### DIFF
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -209,9 +209,11 @@
 // - GCC 11+ (libstdc++): full support since GCC 11
 // - MSVC 19.24+: full support since VS 2019 16.4
 // - Apple Clang: requires macOS 26.0+ deployment target (availability annotations)
-// - libc++ (Android NDK, FreeBSD, Emscripten): requires libc++ 20+ (LLVM 20);
-//   libc++ 17-19 have float from_chars overloads explicitly deleted
-#if defined(__apple_build_version__)
+// - Non-Apple libc++ (Android NDK, FreeBSD, Emscripten): requires libc++ 20+
+//   (LLVM 20); libc++ 17-19 have float from_chars overloads explicitly deleted.
+//   On macOS, even non-Apple LLVM (Homebrew) uses system libc++ headers with
+//   Apple availability annotations, so __APPLE__ must also be excluded here.
+#if defined(__APPLE__)
 #include <AvailabilityMacros.h>
 #endif
 #if defined(__cpp_lib_to_chars) && __cpp_lib_to_chars >= 202306L
@@ -220,9 +222,9 @@
 	#define POCO_HAS_FLOAT_CHARCONV 1
 #elif defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 11
 	#define POCO_HAS_FLOAT_CHARCONV 1
-#elif defined(__apple_build_version__) && defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED >= 260000
+#elif defined(__APPLE__) && defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED >= 260000
 	#define POCO_HAS_FLOAT_CHARCONV 1
-#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 200000 && !defined(__apple_build_version__)
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 200000 && !defined(__APPLE__)
 	#define POCO_HAS_FLOAT_CHARCONV 1
 #endif
 
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -206,8 +206,11 @@
 #endif
 
 // Float std::to_chars/from_chars support detection.
-// Available in GCC 11+, MSVC 19.24+, non-Apple libc++ 17+.
-// Apple Clang requires macOS 26.0+ deployment target due to availability annotations.
+// - GCC 11+ (libstdc++): full support since GCC 11
+// - MSVC 19.24+: full support since VS 2019 16.4
+// - Apple Clang: requires macOS 26.0+ deployment target (availability annotations)
+// - libc++ (Android NDK, FreeBSD, Emscripten): requires libc++ 20+ (LLVM 20);
+//   libc++ 17-19 have float from_chars overloads explicitly deleted
 #if defined(__apple_build_version__)
 #include <AvailabilityMacros.h>
 #endif
@@ -219,7 +222,7 @@
 	#define POCO_HAS_FLOAT_CHARCONV 1
 #elif defined(__apple_build_version__) && defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED >= 260000
 	#define POCO_HAS_FLOAT_CHARCONV 1
-#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 170000 && !defined(__apple_build_version__)
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 200000 && !defined(__apple_build_version__)
 	#define POCO_HAS_FLOAT_CHARCONV 1
 #endif
 
--- a/Foundation/include/Poco/Config.h
+++ b/Foundation/include/Poco/Config.h
@@ -205,6 +205,24 @@
 	#define POCO_HAVE_ATOMIC_SHARED_PTR false
 #endif
 
+// Float std::to_chars/from_chars support detection.
+// Available in GCC 11+, MSVC 19.24+, non-Apple libc++ 17+.
+// Apple Clang requires macOS 26.0+ deployment target due to availability annotations.
+#if defined(__apple_build_version__)
+#include <AvailabilityMacros.h>
+#endif
+#if defined(__cpp_lib_to_chars) && __cpp_lib_to_chars >= 202306L
+	#define POCO_HAS_FLOAT_CHARCONV 1
+#elif defined(_MSC_VER) && _MSC_VER >= 1924
+	#define POCO_HAS_FLOAT_CHARCONV 1
+#elif defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 11
+	#define POCO_HAS_FLOAT_CHARCONV 1
+#elif defined(__apple_build_version__) && defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED >= 260000
+	#define POCO_HAS_FLOAT_CHARCONV 1
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 170000 && !defined(__apple_build_version__)
+	#define POCO_HAS_FLOAT_CHARCONV 1
+#endif
+
 // Option to silence deprecation warnings.
 #ifndef POCO_SILENCE_DEPRECATED
 	#define POCO_DEPRECATED(reason) [[deprecated(reason)]]
--- a/Foundation/include/Poco/NumberFormatter.h
+++ b/Foundation/include/Poco/NumberFormatter.h
@@ -692,7 +692,7 @@ inline bool NumberFormatter::isEnabled(Options options, Options opt)
 inline std::string NumberFormatter::format(int value)
 {
 	std::string result;
-	intToStr(value, 10, result);
+	if (!intToStr(value, 10, result)) result.clear();
 	return result;
 }
 
@@ -700,7 +700,7 @@ inline std::string NumberFormatter::format(int value)
 inline std::string NumberFormatter::format(int value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, ' ');
+	if (!intToStr(value, 10, result, false, width, ' ')) result.clear();
 	return result;
 }
 
@@ -708,7 +708,7 @@ inline std::string NumberFormatter::format(int value, int width)
 inline std::string NumberFormatter::format0(int value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, '0');
+	if (!intToStr(value, 10, result, false, width, '0')) result.clear();
 	return result;
 }
 
@@ -728,7 +728,7 @@ inline std::string NumberFormatter::formatHex(int value, int width, Options opti
 inline std::string NumberFormatter::format(unsigned value)
 {
 	std::string result;
-	intToStr(value, 10, result);
+	if (!intToStr(value, 10, result)) result.clear();
 	return result;
 }
 
@@ -736,7 +736,7 @@ inline std::string NumberFormatter::format(unsigned value)
 inline std::string NumberFormatter::format(unsigned value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, ' ');
+	if (!intToStr(value, 10, result, false, width, ' ')) result.clear();
 	return result;
 }
 
@@ -744,7 +744,7 @@ inline std::string NumberFormatter::format(unsigned value, int width)
 inline std::string NumberFormatter::format0(unsigned int value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, '0');
+	if (!intToStr(value, 10, result, false, width, '0')) result.clear();
 	return result;
 }
 
@@ -752,7 +752,7 @@ inline std::string NumberFormatter::format0(unsigned int value, int width)
 inline std::string NumberFormatter::formatHex(unsigned value, Options options)
 {
 	std::string result;
-	intToStr(value, 0x10, result, isEnabled(options, Options::HEX_PREFIX), -1, ' ', 0, isEnabled(options, Options::HEX_LOWERCASE));
+	if (!intToStr(value, 0x10, result, isEnabled(options, Options::HEX_PREFIX), -1, ' ', 0, isEnabled(options, Options::HEX_LOWERCASE))) result.clear();
 	return result;
 }
 
@@ -760,7 +760,7 @@ inline std::string NumberFormatter::formatHex(unsigned value, Options options)
 inline std::string NumberFormatter::formatHex(unsigned value, int width, Options options)
 {
 	std::string result;
-	intToStr(value, 0x10, result, isEnabled(options, Options::HEX_PREFIX), width, '0', 0, isEnabled(options, Options::HEX_LOWERCASE));
+	if (!intToStr(value, 0x10, result, isEnabled(options, Options::HEX_PREFIX), width, '0', 0, isEnabled(options, Options::HEX_LOWERCASE))) result.clear();
 	return result;
 }
 
@@ -768,7 +768,7 @@ inline std::string NumberFormatter::formatHex(unsigned value, int width, Options
 inline std::string NumberFormatter::format(long value)
 {
 	std::string result;
-	intToStr(value, 10, result);
+	if (!intToStr(value, 10, result)) result.clear();
 	return result;
 }
 
@@ -776,7 +776,7 @@ inline std::string NumberFormatter::format(long value)
 inline std::string NumberFormatter::format(long value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, ' ');
+	if (!intToStr(value, 10, result, false, width, ' ')) result.clear();
 	return result;
 }
 
@@ -784,7 +784,7 @@ inline std::string NumberFormatter::format(long value, int width)
 inline std::string NumberFormatter::format0(long value, int width)
 {
 	std::string result;
-	intToStr(value, 10, result, false, width, '0');
+	if (!intToStr(value, 10, result, false, width, '0')) result.clear();
 	return result;
 }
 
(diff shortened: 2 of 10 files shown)


aleks-f (changes requested):
given the tradeoff, it seems that the best approach is hybrid - for <1000 the old way, >1000 the new one. This is essentially the same pattern used by small string optimization or SpinlockMutex

matejk:
# POCO Numeric Conversion Benchmark Report

Comparison of `main` branch vs `numeric-conversion-performance-improvements` branch.

- **5 million random values** per test, fixed seed for reproducibility
- macOS: Apple Clang, arm64, RelWithDebInfo, deployment target 26.0
- Linux: GCC (OrbStack), arm64, RelWithDebInfo

## NumberFormatter — Integer Formatting

| Operation | macOS main | macOS current | speedup | Linux main | Linux current | speedup |
|---|---:|---:|---|---:|---:|---|
| format(small 0-999) | 53.8 | **45.4** | 1.2x | 66.4 | **53.7** | 1.2x |
| formatHex(small) | 57.3 | **48.7** | 1.2x | 70.3 | **63.0** | 1.1x |
| format(int) | 165.2 | **99.8** | **1.7x** | 218.6 | **95.1** | **2.3x** |
| format(int, w=15) | 188.1 | **128.0** | **1.5x** | 200.5 | **107.8** | **1.9x** |
| format0(int, w=15) | 195.9 | **133.3** | **1.5x** | 213.9 | **115.5** | **1.9x** |
| format(-int) | 135.9 | **93.5** | **1.5x** | 162.7 | **84.6** | **1.9x** |
| formatHex(int) | 73.3 | **51.6** | **1.4x** | 110.9 | **65.8** | **1.7x** |
| formatHex(int, w=10) | 76.2 | **62.3** | 1.2x | 100.8 | **73.6** | 1.4x |
| formatHex(int, prefix) | 86.4 | **65.0** | **1.3x** | 101.6 | **74.7** | **1.4x** |
| format(unsigned) | 110.5 | **70.4** | **1.6x** | 122.8 | **74.0** | **1.7x** |
| formatHex(unsigned) | 75.8 | 82.7 | 0.9x | 85.5 | **66.1** | 1.3x |
| format(Int64) | 332.0 | **121.3** | **2.7x** | 392.5 | **230.1** | **1.7x** |
| format(UInt64) | 318.5 | **136.1** | **2.3x** | 392.4 | **243.1** | **1.6x** |
| formatHex(Int64) | 174.7 | **76.1** | **2.3x** | 258.0 | **189.9** | 1.4x |
| formatHex(UInt64, w=20) | 203.4 | **106.6** | **1.9x** | 294.5 | **211.0** | 1.4x |

### vs standard library baselines (current branch)

| Operation | macOS POCO | macOS snprintf | macOS to_string | Linux POCO | Linux snprintf | Linux to_string |
|---|---:|---:|---:|---:|---:|---:|
| format(int) | **99.8** | 276.7 | 70.0 | **95.1** | 301.9 | 70.3 |
| format(int, w=15) | **128.0** | 325.4 | — | **107.8** | 328.7 | — |
| format0(int, w=15) | **133.3** | 332.4 | — | **115.5** | 321.5 | — |
| formatHex(int) | **51.6** | 204.7 | — | **65.8** | 237.2 | — |

POCO is **2.5-4x faster than snprintf** for all integer formatting options.
`std::to_string` is faster for plain decimal (no width/hex options available).

## NumberFormatter — Float/Double Formatting

| Operation | macOS main | macOS current | speedup | Linux main | Linux current | speedup |
|---|---:|---:|---|---:|---:|---|
| format(double) | 314.2 | 324.8 | ~same | 437.0 | 444.0 | ~same |
| format(double, prec=2) | **204.5** | 322.5 | **0.6x** | 334.3 | 385.9 | 0.9x |
| format(double, prec=6) | 281.4 | 296.9 | ~same | 383.6 | 375.2 | ~same |
| format(double, w=15, prec=4) | 360.3 | 366.2 | ~same | 328.4 | 341.4 | ~same |
| format(float) | 514.2 | **280.4** | **1.8x** | 975.4 | **387.4** | **2.5x** |
| format(float, prec=2) | **202.2** | 224.6 | 0.9x | 331.9 | 329.4 | ~same |

### vs standard library baselines (current branch)

| Operation | macOS POCO | macOS snprintf | macOS to_string | Linux POCO | Linux snprintf | Linux to_string |
|---|---:|---:|---:|---:|---:|---:|
| format(double) | **324.8** | 770.6 | 992.7 | **444.0** | 1115.2 | 1524.3 |
| format(double, prec=2) | **322.5** | 955.1 | — | **385.9** | 1344.5 | — |
| format(double, w=15, prec=4) | **366.2** | 934.8 | — | **341.4** | 1466.7 | — |
| format(float) | **280.4** | 677.1 | — | **387.4** | 1134.3 | — |

POCO is **2-4x faster than snprintf** for all float formatting options.
`format(float)` shortest representation improved **1.8-2.5x** vs main (std::to_chars).
`format(double, prec=2)` regressed on macOS (0.6x) — `std::to_chars(fixed)` + `adjustPrecision`
is slower than double-conversion's integrated `ToFixed` for this case.

## NumberParser — Integer Parsing

| Operation | macOS main | macOS current | speedup | Linux main | Linux current | speedup |
|---|---:|---:|---|---:|---:|---|
| parse(small dec) | 82.0 | **73.0** | 1.1x | 65.7 | **48.9** | **1.3x** |
| parse(uint dec) | 155.7 | **110.1** | **1.4x** | 163.1 | **74.5** | **2.2x** |
| parse(int neg) | 179.2 | **149.6** | 1.2x | 153.1 | **103.6** | **1.5x** |
| parseHex(uint) | 229.7 | **185.1** | 1.2x | 222.6 | **60.1** | **3.7x** |
| parseHex(0x-prefixed) | 242.0 | **170.5** | **1.4x** | 236.5 | **73.5** | **3.2x** |
| parseUnsigned64(uint64) | 319.2 | **172.3** | **1.9x** | 329.0 | **118.4** | **2.8x** |

### vs standard library baselines (current branch)

| Operation | macOS POCO | macOS strtol | macOS sscanf | Linux POCO | Linux strtol | Linux sscanf |
|---|---:|---:|---:|---:|---:|---:|
| parse(uint dec) | 110.1 | **78.3** | 386.4 | **74.5** | 88.1 | 316.8 |
| parseHex(uint) | 185.1 | **183.7** | 575.3 | **60.1** | 183.0 | 390.8 |
| parseUnsigned64 | 172.3 | **133.8** | — | **118.4** | 151.6 | — |

POCO matches or beats `strtol` on most operations. **3-5x faster than sscanf**.
Linux hex parsing is **3x faster than strtoul** thanks to `std::from_chars`.

## NumberParser — Float/Double Parsing

| Operation | macOS main | macOS current | speedup | Linux main | Linux current | speedup |
|---|---:|---:|---|---:|---:|---|
| tryParseFloat(float) | 289.0 | **257.1** | 1.1x | 323.0 | **224.7** | **1.4x** |
| tryParseFloat(double) | 287.5 | **249.4** | 1.2x | 333.3 | **225.5** | **1.5x** |
| tryParseFloat(fixed) | 297.1 | **244.5** | 1.2x | 343.9 | **224.5** | **1.5x** |

### vs standard library baselines (current branch)

| Operation | macOS POCO | macOS strtod | macOS sscanf | Linux POCO | Linux strtod | Linux sscanf |
|---|---:|---:|---:|---:|---:|---:|
| tryParseFloat(double) | 249.4 | **128.9** | 478.6 | **225.5** | 326.6 | 611.3 |

macOS: POCO is 1.9x slower than raw `strtod` (string copy + separator stripping overhead),
but **1.9x faster than sscanf**.
Linux: POCO is **1.4x faster than strtod** and **2.7x faster than sscanf**.

## Summary

### Improvements vs main

| Category | macOS | Linux |
|---|---|---|
| Integer formatting (int) | **1.5-1.7x faster** | **1.7-2.3x faster** |
| Integer formatting (Int64) | **2.3-2.7x faster** | **1.4-1.7x faster** |
| Float format(float) shortest | **1.8x faster** | **2.5x faster** |
| Integer parsing | **1.1-1.9x faster** | **1.3-3.7x faster** |
| Float parsing | **1.1-1.2x faster** | **1.4-1.5x faster** |

### Regressions vs main

| Category | macOS | Linux | Cause |
|---|---|---|---|
| format(double, prec=2) | **0.6x** (322 vs 205 ms) | 0.9x (386 vs 334 ms) | `std::to_chars(fixed)` + `adjustPrecision` slower than double-conversion's integrated `ToFixed` |
| format(float, prec=2) | 0.9x (225 vs 202 ms) | ~same (329 vs 332 ms) | Same as above, less pronounced for float |
| formatHex(unsigned) macOS | 0.9x (83 vs 76 ms) | — | Noise or minor overhead from reverse-based hex vs old forward-write |

### Slower than standard library baselines

| Operation | macOS POCO | macOS baseline | ratio | Linux POCO | Linux baseline | ratio |
|---|---:|---:|---|---:|---:|---|
| format(int) vs to_string | 99.8 | 70.0 | **0.7x** | 95.1 | 70.3 | **0.7x** |
| format(unsigned) vs to_string | 70.4 | 44.2 | **0.6x** | 74.0 | 58.8 | **0.8x** |
| format(Int64) vs to_string | 121.3 | 79.7 | **0.7x** | 230.1 | 120.5 | **0.5x** |
| tryParseFloat vs strtod | 249.4 | 128.9 | **0.5x** | — | — | — |
| parse(small dec) vs strtol | 73.0 | 44.1 | **0.6x** | 48.9 | 41.4 | **0.8x** |
| parse(uint dec) vs strtol | 110.1 | 78.3 | **0.7x** | — | — | — |

**Root causes:**
- **format(int) vs std::to_string**: `std::to_string` writes directly into a string
  with no sign/prefix/width handling. POCO writes into a char buffer, reverses, then
  copies to `std::string` — the extra copy + reversal adds ~30-40% overhead. This is
  inherent to the richer API (width, fill, prefix, hex, thSep support).
- **tryParseFloat vs strtod (macOS)**: POCO's `tryParseFloat(string)` copies the input
  string, strips thousand separators and 'f' suffix, replaces decimal separator, then
  calls `std::from_chars`. The raw `strtod` operates directly on the C string with no
  preprocessing. On Linux, GCC's `std::from_chars` is fast enough to offset this overhead.
- **parse(small dec) vs strtol**: `NumberParser::parse` dispatches through `tryParse` →
  `strToInt(string)` → whitespace skip → `strToInt(range)` → `std::from_chars`. The
  function call chain adds overhead vs a direct `strtol` call. For small numbers where
  the parsing itself is trivial, this overhead dominates.

### vs standard library (current branch)

| Comparison | Integer formatting | Float formatting | Integer parsing | Float parsing |
|---|---|---|---|---|
| vs snprintf | **2.5-4.3x faster** | **2-4.3x faster** | — | — |
| vs std::to_string | 1.1-1.4x slower | **3-3.4x faster** | — | — |
| vs strtol/strtod | — | — | ~same to **3.7x faster** | 1.9x slower (macOS) to **1.4x faster** (Linux) |
| vs sscanf | — | — | **3-5x faster** | **1.9-2.7x faster** |

### Optimization opportunities

**format(double, prec=2) regression (0.6x on macOS):**
The `std::to_chars(fixed) + adjustPrecision` path is slower than double-conversion's
integrated `ToFixed` for fixed-precision formatting. Options:
1. Fall back to double-conversion for `FixedStr` — restores performance but requires
   always compiling double-conversion even when charconv is available.
2. Accept the tradeoff — `std::to_chars` path eliminates ~6800 lines of bundled
   dependency when `POCO_HAS_FLOAT_CHARCONV` is defined.
3. Wait for stdlib improvements — Apple Clang's `std::to_chars(fixed, precision)` may
   improve in future releases.

**tryParseFloat vs strtod (1.9x slower on macOS):**
The overhead is in string preprocessing (copy, trim, separator stripping). Options:
1. Add a fast path in `tryParseFloat(string)` that skips preprocessing when `decSep='.'`
   and `thSep=','` and the string contains no separators (the common case) — check for
   separator characters before copying.
2. Provide a `tryParseFloat(const char*, double&)` overload that skips all preprocessing
   for callers who know their input is clean.

**format(int) vs std::to_string (0.7x slower):**
The overhead is from writing to a char buffer, reversing, then copying to `std::string`.
Not actionable without removing width/padding/hex/prefix support. Users who need raw
speed for plain decimal can use `std::to_string` directly.

**parse(small dec) vs strtol (0.6-0.8x slower):**
The overhead is the dispatch chain (`tryParse` → `strToInt(string)` → `strToInt(range)`
→ `from_chars`). Not actionable without flattening the API. The overhead is constant
per call and becomes negligible for larger numbers.

### Techniques used

- **Integer formatting**: two-digits-at-a-time with digit-pairs lookup table,
  shift-and-mask for hex/power-of-2 bases
- **Integer parsing**: `std::from_chars` (C++17)
- **Float shortest**: `std::to_chars` default representation (when `POCO_HAS_FLOAT_CHARCONV`),
  double-conversion fallback for older compilers
- **Float fixed-precision**: `std::to_chars(fixed)` + manual `adjustPrecision` with
  round-half-up rounding
- **Float parsing**: `std::from_chars` (when `POCO_HAS_FLOAT_CHARCONV`),
  double-conversion fallback

matejk:
> given the tradeoff, it seems that the best approach is hybrid - for <1000 the old way, >1000 the new one. This is essentially the same pattern used by small string optimization or SpinlockMutex

I pushed (almost) completely rewritten formatting/parsing. Benchmarks also indicate that in many cases it does (or would make sense to) use standard library-provided implementations.

matejk:
### Response to review comments

**`POCO_NUMERIC_STRING_PRIVATE` `#error` guard**: This is intentional. `NumericString.h` has always been an internal implementation detail -- the functions are called only by `NumberFormatter` and `NumberParser`. The `#error` prevents accidental direct inclusion by downstream code that should use the public API (`NumberFormatter.h` / `NumberParser.h`) instead. This is not an API break because the header was never part of the public API interace.

**`isSafeIntCast<F, T>` template parameter order**: Pre-existing design, not changed by this PR. Both `isSafeIntCast<F, T>` and `isIntOverflow<To, From>` require explicit template arguments. Swapping the order would break existing callers.

**`Foundation_API` on float/double functions**: Restored. These must remain exported because inline `NumberFormatter` methods (compiled into other libraries like JSON, Net, Data) call them.

All other review findings have been addressed in the latest push.

aleks-f:
> ### Response to review comments
> **`POCO_NUMERIC_STRING_PRIVATE` `#error` guard**: This is intentional. `NumericString.h` has always been an internal implementation detail -- the functions are called only by `NumberFormatter` and `NumberParser`. The `#error` prevents accidental direct inclusion by downstream code that should use the public API (`NumberFormatter.h` / `NumberParser.h`) instead. This is not an API break because the header was never part of the public API surface.

I'm not sure what is considered "public API surface", but these functions certainly were visible and accessible to users so far. We use [strToInt](https://github.com/aleph-us/devs/blob/8c5c2163550981dd3e373f59cc54cd4d214af0e4/alephone/DevsUtil/src/MultiArray.cpp#L135) directly and there may be other users using it. 

Perhaps it could have been better thought initially, but at this point I don't see a rationale for hiding it like this - if someone wants to use it directly, there is no harm.

matejk:
> I'm not sure what is considered "public API surface", but these functions certainly were visible and accessible to users so far. We use [strToInt](https://github.com/aleph-us/devs/blob/8c5c2163550981dd3e373f59cc54cd4d214af0e4/alephone/DevsUtil/src/MultiArray.cpp#L135) directly and there may be other users using it.
> 
> Perhaps it could have been better thought initially, but at this point I don't see a rationale for hiding it like this - if someone wants to use it directly, there is no harm.

Mark as deprecated if used from outside the library?

aleks-f:
> > I'm not sure what is considered "public API surface", but these functions certainly were visible and accessible to users so far. We use [strToInt](https://github.com/aleph-us/devs/blob/8c5c2163550981dd3e373f59cc54cd4d214af0e4/alephone/DevsUtil/src/MultiArray.cpp#L135) directly and there may be other users using it.
> > Perhaps it could have been better thought initially, but at this point I don't see a rationale for hiding it like this - if someone wants to use it directly, there is no harm.
> 
> Mark as deprecated if used from outside the library?

I wouldn't. There are some utility functions there that may come handy, like safeMultiply, locale etc. Maybe we could even add string param to decimal and thousand separator functions, defaulting to empty for global locale. We needed that recently and I forgot it was there and recommended the [std atrocity](https://github.com/aleph-us/devs/pull/1085#discussion_r3058103357).

matejk:
Removed private guard.

PR #5318, opened 2026-04-16
Fix: DateTimeParser trailing timezone designators & Timezone UTC offset cache invalidation

kozemcak:
Hi all,

After we update poco to new version our applications tests start failing. We identified these issues and fix it. Some of the issues are little specific, so we are interesting what you say about that. Here are quick summary of our changes.

Thanks.

## Summary

This MR contains two bug fixes for the `Foundation` library addressing datetime parsing and timezone handling regressions.

---

#### 1. `fix(Foundation): DateTimeParser: %S consume optional fractional seconds`

Parsing ISO 8601 date strings that contain both fractional seconds and a timezone offset (e.g. "2013-10-07T08:23:19.120-04:00") with the format "%Y-%m-%dT%H:%M:%S%z" raises a SyntaxException:

- %S consumes the integer seconds but stops at ., leaving .120-04:00 unconsumed.
- %z (parseTZD) is called next but sees . and returns without consuming anything.
- The trailing-garbage check then raises SyntaxException.

**Fix:** Extend the %S case to consume and discard an optional fractional-second suffix (. or , followed by one or more digits) immediately after parsing the integer seconds. This mirrors the existing %s behaviour and allows %z to see the timezone designator directly, keeping the trailing-garbage check fully effective for truly invalid input.

---

#### 2. `fix(Foundation): Timezone: invalidate utcOffset cache when /etc/localtime changes`

The `TZInfo` UTC offset cache introduced in Poco commit `1850dc16` is only invalidated when the `TZ` environment variable changes. Since `TZ` is process-local, system timezone changes via `/etc/localtime` (symlink update) go undetected, causing `Timezone::utcOffset()` to return a stale value.

**Fix:** Extend `cacheTZ()`/`tzChanged()` to track the inode and `mtime` of `/etc/localtime` via `stat(2)`. Cache is invalidated when either changes.

---

### Testing

- Verified ISO 8601 strings with fractional seconds and UTC offsets parse correctly.
- Verified `utcOffset()` reflects timezone changes after localtime is updated without modifying `TZ`.

### DIFF
--- a/Foundation/src/DateTimeParser.cpp
+++ b/Foundation/src/DateTimeParser.cpp
@@ -243,6 +243,18 @@ void DateTimeParser::parse(const std::string& fmt, const std::string& dtStr, Dat
 				case 'S':
 					it = skipNonDigits(it, end);
 					second = parseNumberN(dtStr, it, end, 2);
+					// Consume optional fractional seconds ('.NNN' or ',NNN') so that a
+					// subsequent %z specifier can reach the timezone designator.
+					// A decimal point/comma not followed by a digit is an error.
+					if (it != end && (*it == '.' || *it == ','))
+					{
+						++it;
+						if (it == end || !Ascii::isDigit(*it))
+						{
+							throw SyntaxException("Invalid DateTimeString: " + dtStr + ", missing fractional digits");
+						}
+						it = skipDigits(it, end);
+					}
 					break;
 				case 's':
 					it = skipNonDigits(it, end);
--- a/Foundation/src/Timezone_UNIX.cpp
+++ b/Foundation/src/Timezone_UNIX.cpp
@@ -18,6 +18,7 @@
 #include <ctime>
 #include <cstdlib>
 #include <string>
+#include <sys/stat.h>
 
 
 namespace Poco {
@@ -66,13 +67,44 @@ private:
 	{
 		const char* tz = std::getenv("TZ");
 		_cachedTZ = tz ? tz : "";
+		// /etc/localtime is the system-wide zone source (updated by timedatectl etc.).
+		// Stat'ing it lets us detect zone changes that don't touch the TZ env var; we
+		// don't read its contents — tzset()/libc still resolves the actual zone data.
+		cacheLocaltimeStat();
 	}
 
 	bool tzChanged() const
 	{
 		const char* tz = std::getenv("TZ");
 		std::string currentTZ = tz ? tz : "";
-		return currentTZ != _cachedTZ;
+		if (currentTZ != _cachedTZ) return true;
+		return localtimeStatChanged();
+	}
+
+	void cacheLocaltimeStat()
+	{
+		struct stat st{};
+		if (::stat("/etc/localtime", &st) == 0)
+		{
+			_localtimeIno = st.st_ino;
+			_localtimeMtime = st.st_mtime;
+		}
+		else
+		{
+			_localtimeIno = 0;
+			_localtimeMtime = 0;
+		}
+	}
+
+	bool localtimeStatChanged() const
+	{
+		struct stat st{};
+		if (::stat("/etc/localtime", &st) == 0)
+		{
+			return st.st_ino != _localtimeIno || st.st_mtime != _localtimeMtime;
+		}
+		// /etc/localtime not accessible: treat as changed only if we had it before
+		return _localtimeIno != 0;
 	}
 
 	static int computeTimeZone()
@@ -112,6 +144,8 @@ private:
 	std::mutex _mutex;
 	int _tzOffset;
 	std::string _cachedTZ;
+	ino_t _localtimeIno = 0;
+	time_t _localtimeMtime = 0;
 };
 
 
--- a/Foundation/include/Poco/DateTimeParser.h
+++ b/Foundation/include/Poco/DateTimeParser.h
@@ -63,6 +63,12 @@ public:
 		/// Throws a SyntaxException if the string cannot be successfully parsed.
 		/// Please see DateTimeFormatter::format() for a description of the format string.
 		/// Class DateTimeFormat defines format strings for various standard date/time formats.
+		///
+		/// Note: The %S specifier parses whole seconds and then silently discards any
+		/// fractional-second suffix of the form '.DDD' or ',DDD' (including a bare decimal
+		/// point that is immediately followed by a non-digit is treated as a parse error).
+		/// Callers that need the fractional digits captured must use %s (millis+micros),
+		/// %i (milliseconds only), %c (centiseconds), or %F (six-digit fractional seconds).
 
 	static DateTime parse(const std::string& fmt, const std::string& str, int& timeZoneDifferential);
 		/// Parses a date and time in the given format from the given string.
--- a/Foundation/testsuite/src/DateTimeParserTest.cpp
+++ b/Foundation/testsuite/src/DateTimeParserTest.cpp
@@ -638,6 +638,37 @@ void DateTimeParserTest::testCustom()
 }
 
 
+void DateTimeParserTest::testISO8601FracSeconds()
+{
+	// ISO8601_FORMAT uses %S which silently discards a well-formed fractional-second
+	// suffix so that the trailing %z can still reach the timezone designator.
+	int tzd;
+
+	// Dot-separated fractional seconds with negative timezone offset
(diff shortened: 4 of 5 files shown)


matejk:
The ASN.1 UTCTime issue (`X509Certificate::validFrom()`/`expiresOn()`) is already fixed in #5263 by including the `%Z` specifier in the format strings on the caller side.

The other example `"2013-10-07 08:23:19.120-04:00"` is not a valid ISO 8601 datetime -- ISO 8601 requires the `T` separator between date and time (the 2019 revision removed the earlier allowance to omit it by mutual agreement). POCO's `ISO8601_FORMAT` correctly requires it.

The real issue is that `%S` stops at the fractional-second separator, leaving `%z` unable to reach the timezone. Rather than tolerating trailing garbage after parsing, a better fix is to make `%S` consume and discard optional fractional seconds:

```cpp
case 'S':
    it = skipNonDigits(it, end);
    second = parseNumberN(dtStr, it, end, 2);
    if (it != end && (*it == '.' || *it == ','))
    {
        ++it;
        it = skipDigits(it, end);
    }
    break;
```

This way `%z` sees the timezone directly and the trailing-garbage check remains fully effective. Could you test this approach?

kozemcak:
Hi [matejk](https://github.com/matejk),

Thank you for the answer. Next week I will look at that and I will try your suggestion.

kozemcak:
Hi @matejk 

Thank you for your help. I again retest it with last main and with fix you recemented and all tests pass now.

Still need use commit `fix(Foundation): Timezone: invalidate utcOffset cache when /etc/localtime changes` which  fix last issue in the application, where we check `/etc/localtime`.  Not sure if you like that part.

Foundation/src/DateTimeParser.cpp
```diff
@@ -243,6 +243,13 @@ void DateTimeParser::parse(const std::string& fmt, const std::string& dtStr, Dat
 				case 'S':
 					it = skipNonDigits(it, end);
 					second = parseNumberN(dtStr, it, end, 2);
+					// Consume optional fractional seconds ('.NNN' or ',NNN') so that a
+					// subsequent %z specifier can reach the timezone designator.
+					if (it != end && (*it == '.' || *it == ','))
+					{
+						++it;
+						it = skipDigits(it, end);
+					}
```
matejk, line 257: Unlike `%s` below, this accepts a stray `.`/`,` not followed by a digit (e.g. `...19.-04:00`) — such inputs silently parse where `%s` rejects them. Suggest mirroring the `%s` guard:

```cpp
if (it == end || !Ascii::isDigit(*it))
    throw SyntaxException("Invalid DateTimeString: " + dtStr + ", missing fractional digits");
```

Also worth a one-liner in the class header noting that `%S` now discards any fractional-second suffix (users who need the digits captured must use `%s`/`%i`/`%c`/`%F`).

And please add `DateTimeParserTest` cases for the headline input `2013-10-07T08:23:19.120-04:00` with `ISO8601_FORMAT`, the `,` fractional variant, and the malformed negative case once the guard above is in place.

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -17,7 +17,9 @@
 #include <mutex>
 #include <ctime>
 #include <cstdlib>
+#include <cstring>
```
matejk: This include doesn't appear to be used in the diff — please remove.

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -66,13 +68,52 @@ class TZInfo
 	{
 		const char* tz = std::getenv("TZ");
 		_cachedTZ = tz ? tz : "";
+		cacheLocaltimeStat();
```
matejk, line 73: Please add a brief comment explaining the rationale here so a future reader knows why we're touching `/etc/localtime` at all. Something like:

```cpp
// /etc/localtime is the system-wide zone source (updated by timedatectl etc.).
// Stat'ing it lets us detect zone changes that don't touch the TZ env var; we
// don't read its contents — tzset()/libc still resolves the actual zone data.
```

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -66,13 +68,52 @@ class TZInfo
...
+	bool localtimeStatChanged() const
+	{
+		struct stat st{};
+		if (::stat("/etc/localtime", &st) == 0)
+		{
+			if (st.st_ino != _localtime_ino || st.st_mtime != _localtime_mtime)
+			{
+				_localtime_ino = st.st_ino;
+				_localtime_mtime = st.st_mtime;
+				return true;
```
matejk: This in-place update is redundant: when `localtimeStatChanged()` returns `true`, `tzChanged()` returns `true`, which triggers `reloadNoLock()` → `cacheTZ()` → `cacheLocaltimeStat()`, and that overwrites these fields anyway. Suggest making this a pure observer (drop the assignments) and removing `mutable` from the member declarations.

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -66,13 +68,52 @@ class TZInfo
...
+			}
+			return false;
+		}
+		// /etc/localtime not accessible: treat as changed only if we had it before
+		if (_localtime_ino != 0) {
+			_localtime_ino = 0;
+			_localtime_mtime = 0;
+			return true;
+		}
```
matejk: Brace style: the rest of the file uses Allman braces. Please reformat for consistency.

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -112,6 +153,8 @@ class TZInfo
 	std::mutex _mutex;
 	int _tzOffset;
 	std::string _cachedTZ;
+	mutable ino_t _localtime_ino = 0;
+	mutable time_t _localtime_mtime = 0;
```
matejk: Naming: rest of the file uses camelCase (`_tzOffset`, `_cachedTZ`). Suggest `_localtimeIno` / `_localtimeMtime` for consistency.

Foundation/src/Timezone_UNIX.cpp
```diff
@@ -66,13 +68,52 @@ class TZInfo
...
+		{
+			_localtime_ino = st.st_ino;
+			_localtime_mtime = st.st_mtime;
+		}
+		else
+		{
+			_localtime_ino = 0;
+			_localtime_mtime = 0;
+		}
+	}
```
matejk, line 97: Out of scope for this PR, but worth noting as future follow-ups: this detection is effectively a no-op on **Android**, where the zone is set via the `persist.sys.timezone` system property rather than `/etc/localtime` — bionic's `tzset()` reads the property. The analogous `Timezone_WIN32.cpp` could grow a similar check by re-querying `GetDynamicTimeZoneInformation` and comparing `TimeZoneKeyName` (or watching the `HKLM\SYSTEM\CurrentControlSet\Control\TimeZoneInformation` registry key).

matejk (commented):
Two well-described fixes for real bugs — thanks. A few suggestions inline before merge.

matejk:
@kozemcak, I'd include the fixes to Poco 1.15.3 that is planned for release in a week.

Can you take a look at the comments above?

kozemcak:
Hi @matejk,
Sorry I miss the notification so I miss your comments, I will look at that.

Foundation/src/DateTimeParser.cpp
kozemcak: add "code block", comment and test

Foundation/src/Timezone_UNIX.cpp
kozemcak: remove

Foundation/src/Timezone_UNIX.cpp
kozemcak: add comment

Foundation/src/Timezone_UNIX.cpp
kozemcak: change to pure observer

Foundation/src/Timezone_UNIX.cpp
kozemcak: not actual after we change it to pure observer

Foundation/src/Timezone_UNIX.cpp
kozemcak: done, change

PR #5346, opened 2026-05-07
Support quoted parameters and URI syntax in PostgreSQL connection strings

db4:
Currently POCO parses PostgreSQL connection strings itself and doesn't support quoted parameters with spaces as described in https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

This PR uses libpq to parse connection strings and also supports URI syntax.

### DIFF
--- a/Data/PostgreSQL/src/SessionImpl.cpp
+++ b/Data/PostgreSQL/src/SessionImpl.cpp
@@ -19,36 +19,9 @@
 #include "Poco/Data/Session.h"
 #include "Poco/NumberParser.h"
 #include "Poco/String.h"
+#include "libpq-fe.h"
 #include <map>
-
-
-namespace
-{
-	std::string copyStripped(std::string::const_iterator aFromStringCItr, std::string::const_iterator aToStringCItr)
-	{
-		// skip leading spaces
-		while ((aFromStringCItr != aToStringCItr) && isspace(*aFromStringCItr)) aFromStringCItr++;
-		// skip trailing spaces
-		while ((aFromStringCItr != aToStringCItr) && isspace(*(aToStringCItr - 1))) aToStringCItr--;
-
-		return std::string(aFromStringCItr, aToStringCItr);
-	}
-
-	std::string createConnectionStringFromOptionsMap(const std::map <std::string, std::string> anOptionsMap)
-	{
-		std::string connectionString;
-
-		for (auto citr = anOptionsMap.begin(); citr != anOptionsMap.end(); ++citr)
-		{
-			connectionString.append(citr->first);
-			connectionString.append("=");
-			connectionString.append(citr->second);
-			connectionString.append(" ");
-		}
-
-		return connectionString;
-	}
-}
+#include <memory>
 
 
 namespace Poco::Data::PostgreSQL {
@@ -116,27 +89,60 @@ void SessionImpl::open(const std::string& aConnectionString)
 	// Default values
 	optionsMap["connect_timeout"] = Poco::NumberFormatter::format(timeout);
 
-	const std::string& connString = connectionString();
+	std::string connString = connectionString();
 
-	for (std::string::const_iterator start = connString.begin();;)
+	// Parse the connection string using libpq's parser
+	char* errMsg = nullptr;
+	const std::unique_ptr<PQconninfoOption, decltype(&PQconninfoFree)> opts(
+		PQconninfoParse(connString.c_str(), &errMsg), &PQconninfoFree);
+	std::string errorMsg;
+	if (errMsg)
 	{
-		std::string::const_iterator finish = std::find(start, connString.end(), ' '); // space is the separator between keyword=value pairs
-		std::string::const_iterator middle = std::find(start, finish, '=');
-
-		if (middle == finish)
+		errorMsg = errMsg;
+		PQfreemem(errMsg);
+	}
+	if (!opts)
+	{
+		if (errorMsg.empty())
 		{
-			throw PostgreSQLException("create session: bad connection string format, cannot find '='");
+			errorMsg = "Unknown error parsing connection string";
 		}
+		throw PostgreSQLException("create session: bad connection string format: " + errorMsg);
+	}
 
-		optionsMap[ copyStripped(start, middle) ] = copyStripped(middle + 1, finish);
+	// Drop defaults already specified by the caller's connection string
+	// so we don't override user-provided values when we re-serialize below.
+	for (PQconninfoOption* opt = opts.get(); opt->keyword != nullptr; ++opt)
+	{
+		if (auto it = optionsMap.find(opt->keyword); it != optionsMap.end() && opt->val != nullptr)
+		{
+			optionsMap.erase(it);
+		}
+	}
 
-		if ((finish == connString.end()) || (finish + 1 == connString.end())) break;
+	char delimiter = ' ';
+	const bool isUri = connString.find("postgresql://") == 0
+					|| connString.find("postgres://") == 0;
+	if (isUri)
+	{
+		delimiter = (connString.find('?') == std::string::npos)? '?': '&';
+	}
 
-		start = finish + 1;
+	for (const auto& opt : optionsMap)
+	{
+		connString += delimiter;
+		connString += opt.first;
+		connString += '=';
+		connString += opt.second;
+
+		if (delimiter == '?')
+		{
+			delimiter = '&';
+		}
 	}
 
 	// Real connect
-	_sessionHandle.connect(createConnectionStringFromOptionsMap(optionsMap));
+	_sessionHandle.connect(connString);
 
 	addFeature("autoCommit",
 		&SessionImpl::autoCommit,
--- a/Data/PostgreSQL/testsuite/src/PostgreSQLTest.cpp
+++ b/Data/PostgreSQL/testsuite/src/PostgreSQLTest.cpp
@@ -142,12 +142,12 @@ void PostgreSQLTest::testConnectNoDB()
 }
 
 
-void PostgreSQLTest::testFailedConnect()
+void PostgreSQLTest::testFailedConnectParams()
 {
 	std::string dbConnString;
 	dbConnString +=  "host=" + getHost();
-	dbConnString += " user=invalid";
-	dbConnString +=	" password=invalid";
+	dbConnString += " user='invalid user'";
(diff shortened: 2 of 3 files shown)


Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -19,38 +19,10 @@
 #include "Poco/Data/Session.h"
 #include "Poco/NumberParser.h"
 #include "Poco/String.h"
+#include "libpq-fe.h"
```
matejk, line 22: Minor: `std::unique_ptr` is used at L95 but `<memory>` is not directly included. It currently arrives transitively, but `misc-include-cleaner` flags it. Consider adding `#include <memory>` here for self-documentation.

Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -116,27 +88,37 @@ void SessionImpl::open(const std::string& aConnectionString)
...
-	const std::string& connString = connectionString();
+	std::string connString = connectionString();
 
-	for (std::string::const_iterator start = connString.begin();;)
-	{
-		std::string::const_iterator finish = std::find(start, connString.end(), ' '); // space is the separator between keyword=value pairs
-		std::string::const_iterator middle = std::find(start, finish, '=');
+	// Parse the connection string using libpq's parser
+	char* errMsg = nullptr;
+	std::unique_ptr<PQconninfoOption, decltype(&PQconninfoFree)> opts(PQconninfoParse(connString.c_str(), &errMsg), &PQconninfoFree);
```
matejk: Minor (clang-tidy `misc-const-correctness`): `opts` is never reassigned and could be declared `const`:

```cpp
const std::unique_ptr<PQconninfoOption, decltype(&PQconninfoFree)> opts(
    PQconninfoParse(connString.c_str(), &errMsg), &PQconninfoFree);
```

Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -116,27 +88,37 @@ void SessionImpl::open(const std::string& aConnectionString)
...
+	if (!opts) {
+		std::string errorMsg = errMsg ? errMsg : "Unknown error parsing connection string";
+		if (errMsg) PQfreemem(errMsg);
+		throw PostgreSQLException("create session: bad connection string format: " + errorMsg);
+	}
```
matejk: Style nit: this file (and the rest of POCO) uses Allman braces (`if (...)\n{`). The new code introduced in this PR consistently uses K&R (`if (...) {`) -- here, L103, L110, L115. Worth reformatting for consistency.

Also (defensive, not a real bug today): `errMsg` is freed only inside the `if (!opts)` branch. Today's libpq does not set `errMsg` when `PQconninfoParse` returns non-null (verified empirically), but if a future libpq starts emitting warnings on success, that allocation would leak. A defensive `PQfreemem(errMsg);` after the success path would future-proof this.

Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -116,27 +88,37 @@ void SessionImpl::open(const std::string& aConnectionString)
...
 
-		if (middle == finish)
-		{
-			throw PostgreSQLException("create session: bad connection string format, cannot find '='");
+	// Set default values
+	for (PQconninfoOption* opt = opts.get(); opt->keyword != nullptr; ++opt) {
+		if (auto it = optionsMap.find(opt->keyword); it != optionsMap.end() && opt->val != nullptr) {
+			optionsMap.erase(it);
 		}
+	}
```
matejk: The comment `// Set default values` is misleading -- the loop does the **opposite**: it *removes* entries from `optionsMap` whenever the user has already provided that keyword, so defaults don't override user values. Suggest renaming for the next reader:

```cpp
// Drop defaults already specified by the caller's connection string
// so we don't override user-provided values when we re-serialise below.
```

The logic itself is correct: `PQconninfoParse` returns the full set of recognised keywords (~41 of them), with `val != nullptr` only for those the user supplied -- I verified empirically.

Minor (clang-tidy): `opt` could be declared `const PQconninfoOption*` since the loop body never mutates `*opt`.

Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -116,27 +88,37 @@ void SessionImpl::open(const std::string& aConnectionString)
...
 
-		optionsMap[ copyStripped(start, middle) ] = copyStripped(middle + 1, finish);
-
-		if ((finish == connString.end()) || (finish + 1 == connString.end())) break;
+	std::string delimiter = " ";
+	if (connString.find("postgresql://") == 0) {
```
matejk: **Critical -- silent corruption of `postgres://` URIs.**

libpq accepts both `postgresql://` and `postgres://` as valid URI schemes ([docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)). This check only matches the long form, so a `postgres://...` URI falls through with `delimiter = " "`.

I verified the failure mode against real libpq:

```
PQconninfoParse("postgres://user@host/db connect_timeout=30")
  user = user
  dbname = db connect_timeout=30      <-- user's dbname corrupted!
  host = host
```

The trailing space-delimited keyword is absorbed into the dbname field. The connection then fails with "database does not exist" or, worse, silently connects to an unintended database. Anyone passing `postgres://...` (the canonical short form used widely in tutorials and ORM defaults) is broken.

Suggested fix:

```cpp
const bool isUri = connString.starts_with("postgresql://")
                || connString.starts_with("postgres://");
if (isUri)
{
    delimiter = (connString.find('?') == std::string::npos) ? "?" : "&";
}
```

(`starts_with` is C++20; POCO targets C++20 by default since 1.15.0 per CLAUDE.md. If you'd rather stay on `find(...) == 0` for safety, the OR-pattern still works.)

Also note: clang-tidy `modernize-use-starts-ends-with` and `performance-faster-string-find` -- replace `find("?")` with `find('?')` (single-char overload).

Data/PostgreSQL/src/SessionImpl.cpp
```diff
@@ -116,27 +88,37 @@ void SessionImpl::open(const std::string& aConnectionString)
...
+		// URI syntax is used
+		delimiter = (connString.find("?") == std::string::npos)? "?": "&";
+	}
 
-		start = finish + 1;
+	for (auto& opt : optionsMap) {
+		connString += delimiter + opt.first  + "=" + opt.second;
+		if (delimiter == "?") delimiter = "&";
 	}
```
matejk: Minor: K&R braces (see L96 comment), and on L116 there's a stray double space `opt.first  + "="`.

L117 `if (delimiter == "?") delimiter = "&";` should have braces per `readability-braces-around-statements`:

```cpp
if (delimiter == "?")
{
    delimiter = "&";
}
```

Data/PostgreSQL/testsuite/src/PostgreSQLTest.cpp
```diff
@@ -160,6 +160,20 @@ void PostgreSQLTest::testFailedConnect()
...
+	std::string dbConnString = "postgresql://invalid%20user:invalid%20password@" + getHost() + ":" + getPort();
+
+	try
+	{
+		Session session(PostgreSQL::Connector::KEY, dbConnString);
+		failmsg ("must fail");
+	}
+	catch (ConnectionFailedException& ex) {}
+	catch (ConnectionException& ex) {}
+}
```
matejk, line 177: **Test coverage gap related to the L110 critical bug.**

`testFailedConnectURI` only exercises the `postgresql://` form. Because the test is designed to fail on auth, the `postgres://` regression is invisible to it -- the connection fails before the malformed `dbname` matters. Two suggestions:

1. Add a sibling test that uses the short scheme:

   ```cpp
   std::string dbConnString = "postgres://invalid%20user:invalid%20password@"
                            + getHost() + ":" + getPort();
   ```

2. Optionally, add a unit-test-level check that doesn't depend on a live server: feed a known URI through `SessionImpl::open`, intercept the string passed to `_sessionHandle.connect`, and assert it round-trips cleanly through `PQconninfoParse` (i.e. `dbname` field stays clean after default injection). That would have caught the `postgres://` regression mechanically.

Minor: catch parameters `ex` are unused (clang-tidy `bugprone-unused-raii` / `misc-unused-parameters`). Either name them `&` or drop the variable -- but matches existing style in this file, so feel free to ignore.

matejk (commented):
Review combining clang-tidy 21 (focused on PR-changed lines), an independent code-reviewer agent, CodeRabbit CLI 0.4.5, and empirical libpq probing.

**One critical finding** (silent corruption of `postgres://` URI connection strings -- see inline on `SessionImpl.cpp` L110), one **important** test-coverage gap, and several minor style/correctness suggestions inline.

Thanks for taking this on -- moving from the home-grown parser to libpq is the right call, and URI support is a welcome addition.

Data/PostgreSQL/testsuite/src/PostgreSQLTest.cpp
db4: Thank you very much for you review, I will address all the issues above, but I don't quite understand your suggestion (2) here. How can I intercept the string passed to `_sessionHandle.connect` without modifying ` SessionImpl::open()`? The latter always tries to connect to a real server, if it can't, it will throw an exception.

matejk: You're right -- my phrasing skipped a step. Suggestion (2) needs a small refactor: extract the parse/rebuild block from `open()` into a private static helper

```cpp
static std::string buildConnectionString(const std::string& userInput, unsigned int timeout);
```

so `open()` ends with `_sessionHandle.connect(buildConnectionString(connectionString(), timeout));`. The test then feeds `"postgres://u@h/db"` through the helper and asserts the result round-trips through `PQconninfoParse` with `dbname == "db"` -- no live server needed, and it catches the `postgres://` regression mechanically.

To clarify the closing line of my earlier reply: my original review had two test suggestions -- (1) a sibling integration test using the `postgres://` short scheme (still hits a live server, fails on auth like the existing `testFailedConnectURI`), and (2) the helper-extraction + offline unit test above. (1) is the cheap option, (2) is the thorough one. Either catches the regression in practice -- no need to do both.

db4:
@matejk I've implemented the changes based on your review comments

PR #5369, opened 2026-05-26
fix(Net): export pocoNetworkInitializer when built with clang-cl #5368

### LINKED ISSUE #5368: Net: pocoNetworkInitializer not exported when PocoNet.dll is built with clang-cl
**Describe the bug**

PR #5205 widened the `POCO_NET_FORCE_SYMBOL` guard in
`Net/include/Poco/Net/Net.h` to `#if defined(POCO_COMPILER_MINGW) || defined(__clang__)`.
`__clang__` is also defined for clang-cl, which uses the MSVC ABI. This
routes clang-cl into the MinGW branch, so a clang-cl-built `PocoNet.dll`
never emits `/export:pocoNetworkInitializer`. DLL consumers (MSVC or
clang-cl) then fail to link with `LNK2001: unresolved external symbol
pocoNetworkInitializer`.

**To Reproduce**

1. Build PocoNet shared with clang-cl.
2. `dumpbin /exports PocoNet.dll | findstr /i pocoNetwork` → no output.
3. Link any DLL that includes `Poco/Net/Net.h` → `LNK2001`.

**Expected behavior**

clang-cl should follow the MSVC `/export:` / `/include:` path; the
symbol should appear in `dumpbin /exports`.

**Environment**

- Windows 10/11 x64
- POCO `main` (regression from #5205, 21c95af)
- clang-cl (LLVM), consumer built with MSVC `cl.exe`
- Shared build (`POCO_DLL`, `Net_EXPORTS`)

pavledragisic:


matejk:
**Alternative approach worth considering**

The consumer-side `/INCLUDE:pocoNetworkInitializer` machinery in `Net.h` exists only to force-reference the `pocoNetworkInitializer` global so its constructor (`NetworkInitializer::NetworkInitializer()` in `Net.cpp`, which calls `WSAStartup`) runs. For a CMake-managed shared build that is redundant: `target_link_libraries(consumer PocoNet)` makes `PocoNet.dll` a load-time dependency, and the Windows loader runs CRT init for the DLL, which invokes that constructor for every consumer regardless of whether the symbol is referenced. Per the MSVC runtime docs, on process attach `_DllMainCRTStartup` "initializes and calls constructors for static and nonlocal data" ([source](https://learn.microsoft.com/en-us/cpp/build/run-time-library-behavior)).

Two small changes would resolve #5368 without the `__imp_` ladder:

1. `Foundation/CMakeLists.txt`: set `POCO_NO_AUTOMATIC_LIBS` PUBLIC for every Win32 CMake build, not just clang-cl. There is no MSVC vs clang-cl difference here -- under CMake the build system handles linking either way, and clang-cl is MSVC-ABI compatible and honors `__declspec(dllexport)` / `/EXPORT` the same way ([Clang MSVC compatibility](https://clang.llvm.org/docs/MSVCCompatibility.html)). The clang-cl-only condition is what currently makes MSVC users still rely on the pragma path.

   ```cmake
   if(WIN32)
       target_compile_definitions(Foundation PUBLIC
           POCO_OS_FAMILY_WINDOWS UNICODE _UNICODE POCO_NO_AUTOMATIC_LIBS)
       target_link_libraries(Foundation PUBLIC iphlpapi)
   endif()
   ```

2. `Net/include/Poco/Net/Net.h`: short-circuit the force-symbol under that flag for the shared consumer:

   ```cpp
   #if defined(POCO_NO_AUTOMATIC_LIBS) && defined(POCO_DLL) && !defined(Net_EXPORTS)
       // CMake-managed shared consumer: PocoNet.dll's loader runs
       // NetworkInitializer::NetworkInitializer() automatically.
       #define POCO_NET_FORCE_SYMBOL(s)
   #elif defined(POCO_COMPILER_MINGW) || defined(__clang__)
       // ...existing arms unchanged...
   ```

Why this is attractive:
- No producer-side rework, no `__imp_` mangling, one added arm.
- Aligns MSVC and clang-cl CMake builds on the same well-understood DLL-load path.
- Static-lib paths keep working: \`POCO_DLL\` is unset for static builds, so the existing \`__clang__\` static-pointer arm still fires. That reference is what pulls \`Net.obj\` from the archive -- per the MSVC linker docs, \`/INCLUDE\` "resolves symbol by adding the object that contains the symbol definition to the program ... useful for including a library object that otherwise would not be linked" ([source](https://learn.microsoft.com/en-us/cpp/build/reference/include-force-symbol-references)), and the static-pointer trick has the same effect at the source level.

Known limitation: a clang-cl-built \`PocoNet.dll\` consumed by a non-CMake project without \`POCO_NO_AUTOMATIC_LIBS\` (e.g. an MSVC solution using the auto-link pragmas) would still hit \`LNK2001\` because the DLL itself does not export the symbol. This PR also fixes that mixed-toolchain case; the alternative does not. \`#pragma comment(linker, "/export:...")\` is the documented MSVC way to export and is one of four supported export methods ([/EXPORT docs](https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function)), so PR #5369's producer-side approach is correct on its own merits.

Just sharing as an option.

Net/include/Poco/Net/Net.h
```diff
@@ -104,7 +104,7 @@ std::string Net_API htmlize(const std::string& str);
 
 extern "C" const struct NetworkInitializer Net_API pocoNetworkInitializer;
 
-#if defined(POCO_COMPILER_MINGW) || defined(__clang__)
+#if defined(POCO_COMPILER_MINGW) || (defined(__clang__) && !defined(_MSC_VER))
```
aleks-f, line 115: this is the only thing needed here.

others are:

- unnecessary/cosmetic (`pocoNetworkInitializer` is already extern "C") 
- DLL needs no `/include`; that is required for static auto-linking at build time only

aleks-f (changes requested):

aleks-f (approved):

PR #5372, opened 2026-05-27
5370 sync upstream sql parser

### LINKED ISSUE #5370: Sync upstream SQLParser
- [ ] sync upstream changes
- [ ] add support for $N (postgres) and :named binding placeholders

### LINKED ISSUE #5371: Bound SQL statement string
There is no way to get a full SQL statement as executed, only the one with placeholders. This is also the case for the ones returned by native APIs, which makes it hard to troubleshoot statement execution failures

A set of utility functions returning fully assembled SQL statement, with bound values.

**Additional context**
Will not be useful for custom TypeHandlers.

### LINKED ISSUE #5373: SQLite in-memory DB class
In-memory SQLite class. 

Can be used transparently as Data::Session
Automatic persistence, sharding and retention policy.

### LINKED ISSUE #5374: Data/SQLite: UB in update-hook callback type on LP64 platforms
## Summary

`Poco::Data::SQLite::Utility::UpdateCallbackType` is declared with `Poco::Int64`
for the rowid parameter, which on LP64 platforms (Linux x86_64/aarch64) is
`long`. SQLite's `sqlite3_update_hook` expects a callback whose rowid parameter
is `sqlite3_int64`, which is `long long`. The current code bridges them with
`reinterpret_cast` in `Utility::eventHookRegister`, but calling a function
through a pointer of a different type than its actual signature is **undefined
behavior** in C++ — even when the integer widths happen to match.

UBSan (`-fsanitize=function`) flags this on every invocation of the update hook,
e.g. running `SQLiteTest::testUpdateCallback`:

```
src/sqlite3.c:102277:5: runtime error: call to function (unknown) through
pointer to incorrect function type
'void (*)(void *, int, const char *, const char *, long long)'
  #1 sqlite3_step (libPocoDataSQLite.so.123)
  #2 Poco::Data::SQLite::SQLiteStatementImpl::hasNext()
  ...
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/sqlite3.c:102277:5
```

## Affected platforms

Anywhere `Poco::Int64` is `long` (Linux LP64: x86_64, aarch64, ...). Windows
LLP64 is unaffected because `Poco::Int64` is already `long long` there, so the
function-pointer types incidentally match.

## Fix

Change the rowid parameter type from `Poco::Int64` to `long long` (exactly what
`sqlite3_int64` is typedef'd to) in:

- `Data/SQLite/include/Poco/Data/SQLite/Utility.h` — `UpdateCallbackType` typedef
- `Data/SQLite/src/Utility.cpp` — `eventHookRegister`: the `reinterpret_cast`
  goes away, direct call to `sqlite3_update_hook`
- `Data/SQLite/include/Poco/Data/SQLite/Notifier.h` — `sqliteUpdateCallbackFn`
  declaration
- `Data/SQLite/src/Notifier.cpp` — matching definition
- `Data/SQLite/testsuite/src/SQLiteTest.{h,cpp}` — test's
  `sqliteUpdateCallbackFn` to match the new signature

Internal `Notifier::_row` / `setRow(Poco::Int64)` can stay; the implicit
`long long → Poco::Int64` conversion at `_row = row` is value-preserving on
every supported platform.

## API impact

Minor: any external code that registered a callback declared with `Poco::Int64`
needs a one-line signature swap to `long long`. Same width and calling
convention everywhere, so the migration is trivial. Windows users see no
change at all (their `Poco::Int64` was already `long long`).

## Verification

After the change, the full Data/SQLite testsuite (160 tests) runs cleanly under
`-fsanitize=undefined,address` with `halt_on_error=1`, no diagnostics.

### LINKED ISSUE #5376: Data/SQLite: race in `SQLiteStatementImpl::compileImpl` — bare `sqlite3_errmsg` outside db mutex
## Summary

`SQLiteStatementImpl::compileImpl` reads the C-string returned by `sqlite3_errmsg` outside the SQLite connection mutex. The pointer is valid after the call, but its **contents** are only protected while `db->mutex` is held. Copying it into a `std::string` after `sqlite3_errmsg` returns races with any other thread that mutates `db->pErr` — for example, a failing `sqlite3_step` on the same connection in another thread.

## Affected code

[Data/SQLite/src/SQLiteStatementImpl.cpp:79-86](Data/SQLite/src/SQLiteStatementImpl.cpp#L79-L86)

```cpp
rc = sqlite3_prepare_v2(_pDB, pSql, -1, &pStmt, &pLeftover);
if (rc != SQLITE_OK)
{
    if (pStmt) sqlite3_finalize(pStmt);
    pStmt = nullptr;
    std::string errMsg = sqlite3_errmsg(_pDB);   // <-- racy copy
    Utility::throwException(_pDB, rc, errMsg);
}
```

`sqlite3_errmsg` internally acquires and releases `db->mutex` around its read of `db->pErr`. The pointer it returns is into `db->pErr`'s UTF-8 buffer and remains valid after the call, but the buffer contents are only safe to read while the mutex is held. The implicit `std::string` construction here copies from that pointer **after** the mutex has been released.

### LINKED ISSUE #5377: Public include of sqlite3.h
**Describe the bug**

Two public headers in `Data/SQLite` include `<sqlite3.h>` directly:

- `Data/SQLite/include/Poco/Data/SQLite/Binder.h`
- `Data/SQLite/include/Poco/Data/SQLite/Extractor.h`

Both did it solely so a small inline template body (`Binder::bindLOB<T>`, `Extractor::extractLOB<T>`) could call three sqlite3 free functions (`sqlite3_bind_blob`, `sqlite3_column_bytes`, `sqlite3_column_blob`). The remaining public headers (`Utility.h`, `MemoryDB.h`, `SessionImpl.h`, `SQLiteStatementImpl.h`) had already moved to opaque forward declarations and do not pull in `<sqlite3.h>`.

Consequences for downstream users:

1. Anyone including `Poco/Data/SQLite/Session.h` (or any other entry point that transitively pulls Binder.h / Extractor.h) must have an `sqlite3.h` reachable on the compile include path, even if they never call the SQLite C API.
2. If the downstream project has a different `sqlite3.h` on its include path than the one POCO was built against (system SQLite vs the bundled copy, mismatched versions, amalgamation vs split headers, custom `SQLITE_API` decoration), it gets ODR / declaration mismatches at compile or link time.
3. On Windows static builds (`POCO_STATIC`), this was the same family of include-path conflicts that already caused [#5373](https://github.com/pocoproject/poco/issues/5373) headaches around the `manifest.db` path and `sqlite3_limit` symbol.

**To Reproduce**

Build any consumer of `Poco::Data::SQLite::Session` against a POCO compiled with the bundled SQLite, then place a different `sqlite3.h` (e.g. a system one with different `SQLITE_VERSION`) earlier on the consumer's include path. Compilation of the consumer's translation unit pulls the consumer-side `sqlite3.h` through Binder.h/Extractor.h and surfaces declaration mismatches.

**Expected behavior**

Public Data/SQLite headers should not pull in `<sqlite3.h>`. The SQLite C API is an implementation detail of the library; only the .cpp files should include it.

**Fix**

Apply the type-erased-helper pattern already used by the rest of the SQLite public headers:

- Replace `#include <sqlite3.h>` with `extern "C" { typedef struct sqlite3_stmt sqlite3_stmt; }` in both headers.
- Wrap the three sqlite3 calls in non-template `static` helpers on the owning class (`Binder::bindBlobStatic`, `Extractor::columnBytes`, `Extractor::columnBlobPtr`) declared in the header and defined in the .cpp.
- Have the inline template bodies route through the helpers; they no longer name a sqlite3 function or macro, so the forward declaration alone is sufficient.

The bodies stay inline (no explicit template instantiation needed, no observable API change for callers).

aleks-f:


matejk:
Shall SQLParser be moved to dependencies as well? I was not aware that this is not original Poco work.

aleks-f:
> Shall SQLParser be moved to dependencies as well? I was not aware that this is not original Poco work.

I am not sure yet. Given the upstream scarce activity and non-responsiveness (it seems that all you can get is [likes](https://github.com/hyrise/sql-parser/issues/236)), we may take over our fork and break with upstream (we have lots of adaptive code already anyway). I gave it [one last attempt](https://github.com/hyrise/sql-parser/pull/264) to keep the sync and if there's no response, we'll go our own merry way.

Data/src/Utility.cpp
```diff
@@ -0,0 +1,521 @@
...
+		// left the binding's _bound flag set (canBind() would otherwise return
+		// false and we'd render nothing). Safe to call: RenderingBinder::reset()
+		// is intentionally a no-op so the cascade in scalar Binding<T>::reset()
+		// (Binding.h:93-98) won't wipe captures from earlier walked bindings.
+		binding->reset();
+
+		for (std::size_t r = 0; r < rowsToCapture; ++r)
+		{
+			if (!binding->canBind()) break;
+			binding->bind(pos);
```
matejk, line 349: **P1 -- exception safety in the binder-swap loop.**

If `binding->bind(pos)` throws (a user `TypeHandler<T>::bind` can), the saved binder is never restored for this binding: control unwinds past the `setBinder(savedBinder)` below. The binding then keeps `pBinder` pointing at the `RenderingBinder`, which is destroyed when `rbPtr` goes out of scope -- a dangling binder reference until the next driver attaches its own on `execute()`.

Suggest a tiny RAII guard so the binder is restored on every exit path:

```cpp
struct BinderGuard {
    AbstractBinding* b;
    AbstractBinder::Ptr saved;
    ~BinderGuard() { if (!saved.isNull()) b->setBinder(saved); }
};
```

No observable change on the happy path; fixes the unhappy path cleanly.

Data/include/Poco/Data/Utility.h
```diff
@@ -0,0 +1,406 @@
...
+		/// SAFETY: the returned string contains the actual bound values.
+		/// Bound values frequently carry sensitive data (passwords, PII,
+		/// tokens). Do not log this unconditionally — call it only where the
+		/// operator has authorised exposure of bound values, typically inside
+		/// a catch block where diagnostic detail outweighs PII risk.
+		///
+		/// The helper does not consume the bindings. It temporarily swaps
+		/// each binding's binder to a RenderingBinder, drives bind() row by
+		/// row, and restores the original binder plus calls reset(). Safe to
+		/// call before or after stmt.execute().
```
matejk, line 289: **P1 -- thread-safety contract.**

This function `const_cast`s the Statement and mutates each binding (binder swap, two `reset()`s, an iterator walk). Calling it concurrently with `stmt.execute()` -- or with another `boundSQL(stmt)` -- on the same Statement is a data race.

The existing line says "Safe to call before or after `stmt.execute()`", which addresses *temporal* ordering only. Suggest adding one explicit sentence:

> Not thread-safe with respect to the Statement: do not call concurrently with `stmt.execute()` or with another `boundSQL(stmt)` on the same Statement.

(This just makes the existing single-driver assumption explicit at the point where a reader is most likely to be tempted to violate it.)

Data/include/Poco/Data/Utility.h
```diff
@@ -0,0 +1,406 @@
...
+
+	static std::string boundSQLBulk(const Statement& stmt, std::size_t maxRows);
+		/// As above, but renders up to maxRows rows joined by ";\n". maxRows
+		/// == 0 is treated as 1. The "-- (+N more rows)" suffix appears when
+		/// the largest binding has more rows than maxRows.
+
+	template <typename... Args>
+	static std::size_t executeSQL(Session& session,
+		const std::string& sql,
+		std::unique_ptr<std::string> context,
```
matejk: **P2 -- API ergonomics.**

`std::unique_ptr<std::string>` to mean "optional, owned" forces every caller to heap-allocate just to pass a tag string. `std::optional<std::string>` says the same thing more idiomatically, removes one allocation per call, and reads better at call sites (`Utility::executeSQL(s, sql, std::nullopt, cb, ...)` vs constructing a unique_ptr). Pure C++17, no dependency change.

If you want to keep zero-copy at the boundary, `std::string_view` (empty == no context) works too -- but then the result struct's lifetime must outlive the view; `optional<string>` is simpler.

Data/include/Poco/Data/Utility.h
```diff
@@ -0,0 +1,406 @@
...
+{
+public:
+	struct ExecResult
+		/// Outcome of Utility::executeSQL, handed to the callback.
+	{
+		std::string  sql;                  /// Bound SQL (placeholders substituted).
+		bool         ok = false;
+		std::size_t  rowsAffected = 0;
+		std::string  error;                /// Populated iff !ok.
+		std::int64_t tsMicros = 0;         /// Wall-clock at execute start.
```
matejk, line 90: **P2 -- comment drift.**

"Wall-clock at execute start" suggests the timestamp is captured next to `stmt.execute()`. In the implementation it is captured at function entry, *before* rendering and Statement construction -- which on a slow render or contended allocator can be noticeably earlier than the actual execute. Either:
- rename the field to `tsCallMicros` / `tsEnteredMicros`, or
- move the `system_clock::now()` capture next to the `steady_clock` `t0` so the doc and the code agree.

(`executionMicros` itself is fine -- the `steady_clock` delta is correctly the `stmt.execute()` duration.)

Data/include/Poco/Data/Utility.h
```diff
@@ -0,0 +1,406 @@
...
+	static std::string formatVar(const Poco::Dynamic::Var& v);
+		/// Inspects a Poco::Dynamic::Var and renders it as a SQL literal,
+		/// recursively dispatching to the appropriate branch (NULL for
+		/// empty; booleans, integers, floats; strings via quoteString;
+		/// DateTime / LocalDateTime / Date / Time via the corresponding
+		/// helpers; otherwise stringified via Var::convert<std::string>
+		/// and quoted).
+
+	static std::string formatBLOB(const BLOB& b);
+		/// Renders Poco::Data::BLOB as the SQL hex literal X'AABB…'.
```
matejk, line 407: **P2 -- dialect caveat.**

`X'AABB...'` is the SQLite / MySQL / standard-SQL form. PostgreSQL prefers `'\xAABB'` for `bytea`; MS SQL uses `0xAABB`. As a diagnostic helper that is OK, but a sentence in the docstring ("output uses standard-SQL hex literal form; not portable to PostgreSQL `bytea` / MS SQL") would save the next reader a debugging round.

Data/include/Poco/Data/RenderingBinder.h
```diff
@@ -0,0 +1,230 @@
...
+	void bind(std::size_t pos, const std::deque<Poco::UInt32>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::list<Poco::UInt32>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::vector<Poco::Int64>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::deque<Poco::Int64>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::list<Poco::Int64>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::vector<Poco::UInt64>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::deque<Poco::UInt64>& val, Direction dir = PD_IN) override;
+	void bind(std::size_t pos, const std::list<Poco::UInt64>& val, Direction dir = PD_IN) override;
+#ifndef POCO_INT64_IS_LONG
+	void bind(std::size_t pos, const std::vector<long>& val, Direction dir = PD_IN) override;
```
matejk, line 124: **P2 -- possible missing override.**

The `#ifndef POCO_INT64_IS_LONG` block adds scalar overrides for both `long` and `unsigned long` (lines 80-82), but the container block only declares `std::vector<long>` / `std::deque<long>` / `std::list<long>`. If `AbstractBinder` declares the matching `unsigned long` container overloads (worth checking on a platform where `POCO_INT64_IS_LONG` is *not* defined), then `vector<unsigned long>` bindings would silently miss this override and render as `"?"` instead of their actual values.

Quick check: if `AbstractBinder` has them, mirror the trio here too.

Data/SQLParser/src/sql/SelectStatement.h
```diff
@@ -7,6 +7,7 @@
 
 namespace hsql {
 enum OrderType { kOrderAsc, kOrderDesc };
```
matejk, line 9: **P2 -- promote to `enum class`.**

As written, this is an *unscoped* enum that injects three very common identifiers (`First`, `Last`, `Undefined`) directly into namespace `hsql`. Macro / identifier collisions are realistic in downstream code that includes this header. Since every caller already uses scoped form (`NullOrdering::First` etc., e.g. in `bison_parser.y` and the new tests), changing to `enum class NullOrdering` is a one-line header change with zero call-site churn.

(For the record: Copilot's earlier comment that "the code won't compile" because callers use `NullOrdering::First` is wrong -- C++11 allows scoped resolution on unscoped enums too. The reason to make this `enum class` is collision hygiene, not compilability.)

Data/SQLParser/src/sql/CreateStatement.h
```diff
@@ -28,29 +28,46 @@ struct SQLParser_API TableConstraint : TableElement {
...
   bool nullable;
+  std::vector<ReferencesSpecification*>* references;
+};
+
+struct SQLParser_API ColumnConstraints {
+  explicit ColumnConstraints();
+
+  std::unordered_set<ConstraintType>* constraints;
+  std::vector<ReferencesSpecification*>* references;
 };
```
matejk, line 71: **P2 -- ownership lives in bison, not on the type.**

The ctor heap-allocates `constraints` and `references` but the type has no destructor. Cleanup is externalised to the bison `%destructor` for `<column_constraints_t>` (added in 7cc3fc4f9). That destructor is correct for the parse-error path, but the type is fragile: any code that constructs a `ColumnConstraints` outside the bison context leaks both members.

The textbook fix is:
- give `ColumnConstraints` a proper RAII destructor that owns `constraints` and `references`;
- add an explicit `release()` that nulls the pointers, called by the happy-path action that transfers ownership to `ColumnDefinition`.

Then the bison `%destructor { delete ($$); } <*>` fallback handles this type generically and the dedicated `<column_constraints_t>` destructor block can go away.

matejk (commented):
Design / code-quality review for the Poco-authored surface (`Utility`, `RenderingBinder`, `Statement::bindings()`) and the vendored design choices the Poco fork now owns. Auto-generated `bison_parser.cpp` / `flex_lexer.cpp` not reviewed line by line.

Leaving 8 inline comments split P1 (worth before merge) and P2 (design polish):
- **P1**: RAII guard for the binder-swap loop; thread-safety docstring.
- **P2**: API ergonomics on `executeSQL(context)`, `NullOrdering` -> `enum class`, RAII for `ColumnConstraints`, doc fixes on `tsMicros` and `formatBLOB` dialect, possible missing `vector<unsigned long>` override.

P3 items (TSAN CI job, scanner-fallback hardening, fork-extension tags in `Expr.h`) intentionally not commented inline -- happy to file as follow-up issues if useful.

PR #5380, opened 2026-06-01
fix(FastLogger): enable SplitterChannel #5379

### LINKED ISSUE #5379: FastLogger can not use SplitterChannel
**Component:** Foundation / Logging (FastLogger)
**Affects:** loggers configured with `type = fast` whose channel is a `SplitterChannel`

## Summary
A logger using `type = fast` (FastLogger) with a `SplitterChannel` only emits **console** output, and **without the configured pattern** (no timestamp/level). File output is silently lost. Commenting out `type = fast` restores correct behavior.

## Repro
```properties
logging.loggers.root.type     = fast
logging.loggers.root.channel  = splitter
logging.channels.splitter.class    = SplitterChannel
logging.channels.splitter.channels = console, file
logging.channels.console.class   = ColorConsoleChannel
logging.channels.console.pattern = %Y-%m-%d %H:%M:%S.%i [%p] %t
logging.channels.file.class      = FileChannel
logging.channels.file.path       = app.log
logging.channels.file.pattern    = %Y-%m-%d %H:%M:%S.%i [%p] %t
```
Expected: formatted lines on console **and** in `app.log`. Actual: unformatted console only; `app.log` never written.

## Root cause
`FastLogger` maps Poco channels onto Quill sinks in `collectSinksFromChannel()`. `SplitterChannel` exposed `count()` but **no accessor for its children**, so the splitter branch couldn't enumerate them and fell back to a single, unformatted console sink â€” losing the file sink and the per-child `PatternFormatter`. Separately, `EventChannel` (e.g. a web console) cannot be a Quill sink at all (FastLogger bypasses Poco's Message/Channel pipeline), and hit the generic "unknown â†’ console" fallback, adding a duplicate console sink.

## Fix
- **`SplitterChannel`** add `Channel::Ptr getChannel(int index) const` (pairs with the existing `count()`) so callers can enumerate children. Follows Poco's per-container convention (`AsyncChannel`/`FormattingChannel` expose a no-arg `getChannel()`; the multi-child splitter takes an index).
- **`FastLogger`** recurse into `SplitterChannel` children (collects every sink and picks up per-child patterns); skip `EventChannel` with a one-line stderr warning instead of the duplicate-console fallback; de-duplicate sinks (the console sink is a process-wide singleton).

## Behavior after fix
- `console` + `file` both receive formatted output.
- `EventChannel`/web-console children are omitted with a clear warning; use a regular `Logger` for those.

## Tests (`Foundation/testsuite/.../FastLoggerChannelsTest`)
- Rewrite `testSplitterChannel`: asserts a splitter with two children writes **both** files (regression guard) and exercises the new `count()`/`getChannel()` accessors.
- Add `testEventChannelSkipped`: a splitter with an `EventChannel` + `FileChannel`; the file sibling still works and nothing crashes.

## Files
- `Foundation/include/Poco/SplitterChannel.h`
- `Foundation/src/SplitterChannel.cpp`
- `Foundation/src/FastLogger.cpp`
- `Foundation/testsuite/src/FastLoggerChannelsTest.{h,cpp}`

### LINKED ISSUE #5381: CMake Debug executables do not have 'd' postfix
Debug libraries get 'd' postfix, but executables do not.

aleks-f:
Makes the FastLogger `type=fast` logging path production-ready and removes the SplitterChannel limitation. Rebased on `main`.

Fixes #5379. Also fixes #5381 (Debug `d` postfix for executables).

## Feature
- Enable SplitterChannel as a FastLogger sink; a splitter maps to multiple Quill sinks.
- Per-source routing: each `Poco::Logger` source gets its own Quill logger, preserving `%s` / `%(logger)`.
- Make `type=fast` usable from `Poco::Logger` by installing the FastLogger as the logger's channel, preserving source names.
- Honor per-logger levels under `type=fast`.
- PatternFormatter translation: local-time timestamps by default, honor `times` / `%L`, keep timestamp separators.
- Flush per-source bridge loggers, bound their count, and route EventChannel-only configs to a null sink.

## Correctness and robustness fixes
- Reconfigure routing: `setChannel()` releases the prior per-source Quill loggers, so a reconfigure rebinds each source to the new sinks instead of reusing stale ones.
- `setPattern()` / `addFileSink()` throw `NotImplementedException` instead of a silent no-op (`addFileSink` no longer truncates the target file as a side effect); sinks and pattern come from the Channel passed to `setChannel()`.
- `_level` and the opaque Quill/source-state pointers are atomic for the lock-free reads on the logging path.
- Mirror the `Poco::Logger` level only when one is configured; otherwise leave hierarchy inheritance intact.
- Skip a malformed numeric `quill.*` backend option with a warning instead of throwing from the first log call.

## Build (#5381)
- Give Poco's own executables the Debug `d` postfix via a `poco_add_executable` macro instead of overriding the built-in `add_executable`. The global override leaked into projects that build Poco through `add_subdirectory` and captured `CMP0155`, disabling C++20 module scanning for the executables it created. The now-redundant per-target `DEBUG_POSTFIX` lines are removed.
- Remove the unused `build.ps1` dev script.

## Tests
- Per-source routing, reconfigure routing, multiple sources, unsupported setters, concurrent logging, per-logger level across a splitter, malformed options, and the EventChannel-only null-sink path.

Also includes minor Data/SQLite test and documentation additions (#5370).


matejk:
@aleks-f, I reviewed the code and made come corrections. Can you please verify that my changes are sane? Then I'd merge to main.

aleks-f:
> @aleks-f, I reviewed the code and made come corrections. Can you please verify that my changes are sane? Then I'd merge to main.

the main reason for fast logger was speedup, which was partially because of pattern formatting. if we completely disregard quill pattern formatting, part of the speedup goes away and fast logger loses at least a part its original purpose. 

To see what we currently have in this PR, we need [benchmark results ](https://github.com/pocoproject/poco/tree/main/Benchmark/src) on linux and windows to tell whether fast logger as it is now actually brings any benefit. 

Also, FastLogger must remain transparent - adding `.type = fast` to ordinary Poco logger properties must work

matejk:
Benchmark results comparing FastLogger with the regular Logger (Linux, OrbStack 18-core VM, null sinks, same pattern). Numbers are cost per logging call; CPU is time on the logging thread. Cases are in Benchmark/src/LoggerBench.cpp.

Single sink, typical pattern: regular Logger 472 ns; FastLogger direct API 8.4 ns CPU (87 ns wall); FastLogger through Poco::Logger (type=fast) 437 ns.

SplitterChannel fan-out, cost per call by sink count:

| sinks | regular Logger | FastLogger direct (CPU) | FastLogger via Poco::Logger |
|---|---|---|---|
| 1 | 731 ns | 14.0 ns | 455 ns |
| 2 | 838 ns | 13.6 ns | 447 ns |
| 4 | 966 ns | 13.5 ns | 450 ns |
| 8 | 1302 ns | 13.4 ns | 446 ns |

Regular logging cost grows linearly with the sink count, because it formats once per child on the logging thread. FastLogger cost is independent of the sink count: the logging thread enqueues once and the backend renders and fans out. The direct API is 52x faster at one sink and 97x at eight; through Poco::Logger it is 1.6x at one sink and 2.9x at eight.

With a single sink the Poco::Logger path is not faster than the regular Logger, because Poco::Logger builds the full Message (timestamp, thread id, source) synchronously before the channel, and that cost is not deferred to Quill. The direct API avoids it. For multiple sinks the Poco::Logger path still gains, since the regular path pays per-sink formatting and FastLogger does not.

matejk (approved):
Verified performance.

PR #5396, opened 2026-07-01
chore(deps): update bundled expat 2.8.1 -> 2.8.2 (13 CVEs)

### LINKED ISSUE #5395: Expat 2.8.2 update
Expat before 2.8.2 release is affected by 13 vulnerabilities all listed here https://blog.hartwork.org/posts/expat-2-8-2-released/

Would it be possible that Expat could be updated to the latest 2.8.2 one?

Release https://github.com/libexpat/libexpat/releases/tag/R_2_8_2

beatabaranowska:
Security fixes from upstream (Expat 2.8.2), all CVE-2026:

Integer overflows (9):
- CVE-2026-56403, CVE-2026-56404, CVE-2026-56405, CVE-2026-56406, CVE-2026-56407, CVE-2026-56408, CVE-2026-56409, CVE-2026-56410, CVE-2026-56411

Missing control flow integrity checks (3):
- CVE-2026-50219, CVE-2026-56131, CVE-2026-56412

Out-of-bounds write (1):
- CVE-2026-56132

The integer overflows land in xmlparse.c; the control-flow-integrity and bounds checks touch xmlparse.c, xmltok.c and xmltok_impl.c.

Other upstream changes carried in:
- New helper sources added to the bundle: fallthrough.h, memory_sanitizer.h and xcsinc.c (the latter is #included from the char-type variants, not a standalone translation unit).
- errno-propagation and siphash.h cleanups; minor random_*.c churn.
- Library VERSION/SOVERSION moved from 1.12.1 to 1.12.2 (LIBCURRENT 13 / LIBREVISION 2 / LIBAGE 12).

Bundled-build adjustments:
- dependencies/expat/CMakeLists.txt tracks upstream's reworked lib/CMakeLists.txt; Poco's portable expat_config.h is preserved (the update did not overwrite it).
- dependencies/README.md dependency table: 2.8.1 -> 2.8.2.

Ref: https://blog.hartwork.org/posts/expat-2-8-2-released/

Verified: XML-testrunner -all passes (138 tests) on Windows x64
(MSVC 19.44, VS 2022) in a minimal cmake build (POCO_MINIMAL_BUILD +
ENABLE_XML + ENABLE_TESTS).

### DIFF
--- a/XML/src/ParserEngine.cpp
+++ b/XML/src/ParserEngine.cpp
@@ -274,11 +274,11 @@ void ParserEngine::parse(const char* pBuffer, std::size_t size)
 	while (processed < size)
 	{
 		const int bufferSize = static_cast<int>(std::min(static_cast<std::size_t>(PARSE_BUFFER_SIZE), size - processed));
-		if (!XML_Parse(_parser, pBuffer + processed, bufferSize, 0))
+		if (!XML_Parse(_parser, pBuffer + processed, bufferSize, 0) || _exception)
 			checkError(_parser);
 		processed += bufferSize;
 	}
-	if (!XML_Parse(_parser, pBuffer+processed, 0, 1))
+	if (!XML_Parse(_parser, pBuffer+processed, 0, 1) || _exception)
 		checkError(_parser);
 	if (_pContentHandler) _pContentHandler->endDocument();
 	popContext();
@@ -290,14 +290,14 @@ void ParserEngine::parseByteInputStream(XMLByteInputStream& istr)
 	std::streamsize n = readBytes(istr, _pBuffer, PARSE_BUFFER_SIZE);
 	while (n > 0)
 	{
-		if (!XML_Parse(_parser, _pBuffer, static_cast<int>(n), 0))
+		if (!XML_Parse(_parser, _pBuffer, static_cast<int>(n), 0) || _exception)
 			checkError(_parser);
 		if (istr.good())
 			n = readBytes(istr, _pBuffer, PARSE_BUFFER_SIZE);
 		else
 			n = 0;
 	}
-	if (!XML_Parse(_parser, _pBuffer, 0, 1))
+	if (!XML_Parse(_parser, _pBuffer, 0, 1) || _exception)
 		checkError(_parser);
 }
 
@@ -307,14 +307,14 @@ void ParserEngine::parseCharInputStream(XMLCharInputStream& istr)
 	std::streamsize n = readChars(istr, reinterpret_cast<XMLChar*>(_pBuffer), PARSE_BUFFER_SIZE/sizeof(XMLChar));
 	while (n > 0)
 	{
-		if (!XML_Parse(_parser, _pBuffer, static_cast<int>(n*sizeof(XMLChar)), 0))
+		if (!XML_Parse(_parser, _pBuffer, static_cast<int>(n*sizeof(XMLChar)), 0) || _exception)
 			checkError(_parser);
 		if (istr.good())
 			n = readChars(istr, reinterpret_cast<XMLChar*>(_pBuffer), PARSE_BUFFER_SIZE/sizeof(XMLChar));
 		else
 			n = 0;
 	}
-	if (!XML_Parse(_parser, _pBuffer, 0, 1))
+	if (!XML_Parse(_parser, _pBuffer, 0, 1) || _exception)
 		checkError(_parser);
 }
 
@@ -339,14 +339,14 @@ void ParserEngine::parseExternalByteInputStream(XML_Parser extParser, XMLByteInp
 		std::streamsize n = readBytes(istr, pBuffer, PARSE_BUFFER_SIZE);
 		while (n > 0)
 		{
-			if (!XML_Parse(extParser, pBuffer, static_cast<int>(n), 0))
+			if (!XML_Parse(extParser, pBuffer, static_cast<int>(n), 0) || _exception)
 				checkError(extParser);
 			if (istr.good())
 				n = readBytes(istr, pBuffer, PARSE_BUFFER_SIZE);
 			else
 				n = 0;
 		}
-		if (!XML_Parse(extParser, pBuffer, 0, 1))
+		if (!XML_Parse(extParser, pBuffer, 0, 1) || _exception)
 			checkError(extParser);
 	}
 	catch (...)
@@ -366,14 +366,14 @@ void ParserEngine::parseExternalCharInputStream(XML_Parser extParser, XMLCharInp
 		std::streamsize n = readChars(istr, pBuffer, PARSE_BUFFER_SIZE/sizeof(XMLChar));
 		while (n > 0)
 		{
-			if (!XML_Parse(extParser, reinterpret_cast<char*>(pBuffer), static_cast<int>(n*sizeof(XMLChar)), 0))
+			if (!XML_Parse(extParser, reinterpret_cast<char*>(pBuffer), static_cast<int>(n*sizeof(XMLChar)), 0) || _exception)
 				checkError(extParser);
 			if (istr.good())
 				n = readChars(istr, pBuffer, static_cast<int>(PARSE_BUFFER_SIZE/sizeof(XMLChar)));
 			else
 				n = 0;
 		}
-		if (!XML_Parse(extParser, reinterpret_cast<char*>(pBuffer), 0, 1))
+		if (!XML_Parse(extParser, reinterpret_cast<char*>(pBuffer), 0, 1) || _exception)
 			checkError(extParser);
 	}
 	catch (...)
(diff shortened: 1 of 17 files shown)


dependencies/expat/CMakeLists.txt
matejk, line 1: Keep the CMakeLists.txt from Poco. Do not replace it.

beatabaranowska: Reverted with only numbers updated

beatabaranowska: Please feel free to close this PR if it would be easier to update expat by yourself :)

matejk: Did you try to compile it yourself at all?

beatabaranowska: I guess I did misunderstood. I though I should not touch CMakeLists.txt as you prefer to manage it by yourself but I guess I was supposed just to keep the current layout of the file.
Anyway I pushed the changes but if you prefer to put them into patch to apply please let me know.
I have compiled the files in my project.  Are there any required configurations that need to be checked? I am afraid I am not able to check every possible ones :)

aleks-f (changes requested):
It leaks memory. I don't think we've seen it before and I'm not sure if it is legit. If it is, ideally it should be fixed upstream before merging, but perhaps we can temporarily patch it. If false positive, should be suppressed.

```
==10914==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 6818 byte(s) in 27 object(s) allocated from:
    #0 0x7f45ecafd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x7f45ec892793 in expat_malloc /home/runner/work/poco/poco/dependencies/expat/src/xmlparse.c:911

Indirect leak of 2048 byte(s) in 1 object(s) allocated from:
    #0 0x7f45ecafd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x7f45ec8a0f29 in XML_GetBuffer /home/runner/work/poco/poco/dependencies/expat/src/xmlparse.c:2581

Indirect leak of 1104 byte(s) in 1 object(s) allocated from:
    #0 0x7f45ecafd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x7f45ec89c179 in parserCreate /home/runner/work/poco/poco/dependencies/expat/src/xmlparse.c:1313

SUMMARY: AddressSanitizer: 9970 byte(s) leaked in 29 allocation(s).
```

dependencies/expat/CMakeLists.txt
```diff
@@ -11,8 +11,14 @@ else()
...
 	# those are added conditionally below to mirror the upstream
 	# CMakeLists gating (see expat-2.8.1/CMakeLists.txt).
+	# Also exclude xcsinc.c: unlike xmltok_impl.c/xmltok_ns.c (which guard
+	# themselves with #ifdef XML_TOK_IMPL_C/XML_TOK_NS_C and compile to an
+	# empty translation unit standalone), xcsinc.c has no include guard and
+	# is only meant to be #include-d from xmlparse.c; compiling it as its
+	# own translation unit fails since XML_Char is undefined.
 	file(GLOB SRCS_G "src/*.c")
 	list(FILTER SRCS_G EXCLUDE REGEX "/random_.*\\.c$")
+	list(FILTER SRCS_G EXCLUDE REGEX "/xcsinc\\.c$")
```
matejk, line 21: Good catch, and thanks for the comment explaining why `xcsinc.c` must be excluded while `xmltok_impl.c`/`xmltok_ns.c` need not be -- that distinction is easy to trip over on the next update.

XML/src/ParserEngine.cpp
```diff
@@ -639,6 +652,34 @@ void ParserEngine::handleError(int errorNo)
...
+	// C call stack: Expat (>= 2.8.2) tracks a handler-call-depth counter
+	// that is incremented before and decremented after every callback
+	// invocation, and refuses to free (XML_ParserFree() becomes a no-op)
+	// a parser it believes is still inside a handler call. A C++ exception
+	// propagating out of a handler skips the decrement, which would
+	// otherwise permanently leak the parser (and everything it owns).
+	// XML_StopParser() is Expat's own supported way of aborting a parse
+	// from within a handler.
+	_exception = ex;
+	XML_StopParser(parser, XML_FALSE);
```
matejk, line 679: Minor: `XML_StopParser()` can in theory fail (it returns `XML_STATUS_ERROR` when the parser is already finished). If that ever happened, `XML_Parse()` could return success and the exception captured in `_exception` would be silently dropped. It cannot happen from inside a handler today, but a cheap guard would make this future-proof: also check `_exception` after the final successful `XML_Parse()` call, or assert on the return value here.

XML/src/ParserEngine.cpp
```diff
@@ -639,6 +652,34 @@ void ParserEngine::handleError(int errorNo)
...
+	{
+		std::exception_ptr ex = _exception;
+		_exception = nullptr;
+		std::rethrow_exception(ex);
+	}
+	handleError(XML_GetErrorCode(parser));
+}
+
+
+void ParserEngine::abortParse(XML_Parser parser, std::exception_ptr ex)
```
matejk, line 667: For completeness: `handleUnknownEncoding()` is now the only callback that can still throw through Expat (`XMLString` allocation, or `TextEncoding::find()` under `bad_alloc`). Practically negligible, but wrapping it as well (returning `XML_STATUS_ERROR` on exception) would close the last gap. `handleDefault` is empty and `handleEntityDecl` only dispatches to already-wrapped functions, so those are fine.

XML/src/ParserEngine.cpp
```diff
@@ -757,45 +842,51 @@ int ParserEngine::handleExternalEntityRef(XML_Parser parser, const XML_Char* con
...
 			XML_ParserFree(extParser);
-			throw;
+			return XML_STATUS_OK;
 		}
-		pEntityResolver->releaseInputSource(pInputSource);
-		XML_ParserFree(extParser);
-		return XML_STATUS_OK;
+		else return XML_STATUS_ERROR;
+	}
+	catch (...)
```
matejk, line 880: This `catch (...)` also fixes a pre-existing leak: the old code only caught `XMLException&`, so a non-XML exception (e.g. an I/O failure from the entity stream) escaped through Expat without releasing the input source or freeing `extParser`. Releasing the resources here and stopping the outer parser via the handler's own `parser` argument is also correct for nested external entities. Nice.

matejk (commented):
Thanks for the update and the follow-up fix. I reviewed the changes and reproduced the CI failures locally.

- All files in `dependencies/expat/src` are byte-identical to the upstream `R_2_8_2` release; only Poco's own `expat_config.h` differs, as intended.
- The ASAN leak is real and is not an Expat bug. Expat 2.8.2 tracks a handler-call-depth counter, and `XML_ParserFree()` etc. silently do nothing while it is nonzero. A C++ exception thrown from a handler through Expat's C frames skips the decrement, so the later `XML_ParserFree()` in `ParserEngine::init()` becomes a no-op and the parser, DTD and buffer (~10 KB) leak.
- Reproduced at 66308c0: `XML-testrunner -all` leaks 9970 bytes in 29 allocations (same as CI), triggered by `ParserWriterTest::testMaxElementDepth`. At 3141aa0: 138/138 tests pass, no leaks.
- @aleks-f: the leaks are reported as "indirect" only because the parser and DTD blocks form a pointer cycle. Not a false positive: every parse that ends with a throwing handler leaks the parser. Nothing to suppress or fix upstream.
- The `XML_StopParser()` approach is the correct fix, and exception behavior is preserved (verified against the test suite). Good analysis and clean implementation.

Remaining asks:

1. Rewrite the second commit message ("Patch to potentially revert ?"), e.g. `fix(XML): abort parse via XML_StopParser instead of throwing through Expat callbacks`. Squashing the two commits would keep history bisectable, since the first commit alone fails CI.
2. Optional: add a regression test with a deliberately throwing `ContentHandler`, asserting the exception type still propagates and the parser stays reusable.

Inline comments follow, mostly minor.

XML/src/ParserEngine.h
```diff
@@ -220,6 +221,33 @@ class XML_API ParserEngine: public Locator
...
 		/// to the given Expat error code.
 
+	void checkError(XML_Parser parser);
+		/// Called after a failed XML_Parse()/XML_ParseBuffer() call.
+		/// If a C++ exception thrown from within a handler has been
+		/// captured by abortParse(), rethrows it (and clears it);
+		/// otherwise translates the parser's current Expat error code
+		/// into an exception via handleError().
+
+	void abortParse(XML_Parser parser, std::exception_ptr ex);
```
matejk, line 231: Clear and accurate description of the failure mechanism -- this will save the next person a debugging session.

matejk:
I pushed 89cd28d directly to this branch (maintainer edit) implementing the review follow-ups:

- `checkError()` is now also invoked when `XML_Parse()` succeeds but an exception was captured, so a captured handler exception can never be silently dropped if `XML_StopParser()` ever fails.
- `handleUnknownEncoding()` now captures exceptions via `abortParse()` instead of letting them unwind through Expat.
- New regression test `SAXParserTest::testContentHandlerThrows`: a `std::runtime_error` thrown from `startElement()` propagates unchanged, an `XMLException` surfaces as `SAXParseException`, and the parser remains usable afterwards. Under the ASAN CI jobs this test would have caught the original leak.

Verified locally on Linux (gcc, `-fsanitize=address`): 139/139 XML tests pass, no leaks.

matejk:
History note: I force-pushed this branch to clean up commit metadata -- content is unchanged (identical trees):

- `3141aa0` ("Patch to potentially revert ?") is now `e90689c` with a proper message: `fix(XML): abort parse via XML_StopParser instead of throwing through Expat callbacks`.
- `89cd28d` is now `acc60ab`, credited to @beatabaranowska as author with a `Co-authored-by` trailer for me.

@beatabaranowska: your local `main` has diverged from your fork. Before making further changes, please run `git fetch && git reset --hard origin/main` in your clone (where `origin` is your fork) -- do not `git pull`, it would create a merge.

PR #5401, opened 2026-07-09
fix(SQLChannel): initialize _flush to DEFAULT_FLUSH_SECONDS

### LINKED ISSUE #5400: SQLChannel: _flush is never initialized
`std::atomic<int> _flush` in `SQLChannel` (Data/include/Poco/Data/SQLChannel.h) has no default member initializer and is not initialized in either constructor. It is only assigned when the `flush` property is explicitly set. However, `shouldFlush()` reads it unconditionally on the logging thread:

```cpp
inline bool SQLChannel::shouldFlush() const
{
	return (_flush > 0 && _source.size() &&
		(_flushTimer.elapsedSeconds() >= _flush));
}
```

Before C++20, a default-constructed `std::atomic<int>` holds an indeterminate value, so a channel whose `flush` property was never configured reads uninitialized memory and flushes unpredictably (C++20 value-initializes it to 0, i.e. never flush).

`_flush`, `shouldFlush()` and `DEFAULT_FLUSH_SECONDS = 10` were all introduced in #4530, but the constant is never used — it appears the intended constructor initialization was left out. `_flush` should be initialized to `DEFAULT_FLUSH_SECONDS` in both constructors.

matejk:
`_flush` was never initialized: not in either constructor and with no default member initializer, it holds an indeterminate value (pre-C++20) until the `flush` property is explicitly set, while `shouldFlush()` reads it unconditionally on the logging thread. `DEFAULT_FLUSH_SECONDS` was introduced together with `_flush` in #4530 but never used.

This initializes `_flush` to `DEFAULT_FLUSH_SECONDS` in both constructors, documents the default, and asserts it in `testSQLChannel`.

Closes #5400


matejk:
Folded into #5399.

PR #5433, opened 2026-08-10
enh(Crypto,NetSSL_OpenSSL): support OpenSSL 4.0 #5420

### LINKED ISSUE #5420: Support OpenSSL 4.0
OpenSSL 4.0.0 has been released. Poco should build cleanly and behave
correctly against it. This issue tracks the required work for 2.0.

Reported downstream by Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138410
("poco: FTBFS with openssl 4.0", package version 1.14.2-6)

## Status of the reported build failure

The Debian FTBFS is the const-correctness change in `X509_alias_get0()`:

```
Crypto/src/PKCS12Container.cpp:132: error: reinterpret_cast from type
'const unsigned char*' to type 'char*' casts away qualifiers
```

This is already fixed by #5239, which constified the `X509*` usage and
switched to the `ASN1_STRING` accessors -- commit 413d2b9b5 on `main`,
and ee292b113 on the 1.15.x branch. No action needed for 1.16.

## Remaining work for OpenSSL 4.0

1. **`libcrypto` no longer runs `OPENSSL_cleanup()` from an `atexit()`
   handler.** Checked: no code change is needed. OpenSSL 4.0 leaves its
   global data to the operating system at exit, and `OPENSSL_cleanup()`
   is meant for applications only (OpenSSL cannot be initialized again
   afterwards). `OpenSSLInitializer::uninitialize()` keeps the providers
   loaded and reachable; this is leak-free under LeakSanitizer, valgrind
   and macOS `leaks` with OpenSSL 3.0.13, 3.5.5, 3.6.4 and 4.0.2. Only
   the comments and the documentation were wrong; #5433 corrects them.

2. **FIPS API -- partly done.** #5460 converted
   `OpenSSLInitializer::isFIPSEnabled()` and `SSLManager::isFIPSEnabled()`
   to `EVP_default_properties_is_fips_enabled()`, and made FIPS usable in
   `Context::initDH()` (RFC 7919 ffdhe groups) and `Context::initECDH()`.

   What remains: `OpenSSLInitializer::enableFIPSMode()` and the
   `openSSL.fips` configuration wiring in `SSLManager::initDefaultContext()`
   are still behind `#ifdef OPENSSL_FIPS`, which has not been defined since
   1.0.2. So Poco can now detect FIPS but cannot enable it. Either wire
   `EVP_default_properties_enable_fips()` up unconditionally, or drop the
   setter and document that FIPS is enabled through OpenSSL configuration.

3. **`Context::createSSLContext()` FIPS branch.** Now that
   `isFIPSEnabled()` is live (#5460), the branch at
   `NetSSL_OpenSSL/src/Context.cpp:524` is reachable for the first time
   since OpenSSL 1.0.2. It uses `TLS_method()` for both client and server
   and leaves `minTLSVersion` at 0, so `SSL_CTX_set_min_proto_version()`
   is skipped entirely: a peer with `fips=yes` in its default properties
   silently loses the minimum-TLS-version floor. Either set
   `minTLSVersion` as the `else` path does, or remove the branch, since
   FIPS selection no longer needs a different method function.

4. **CI coverage.** Add an OpenSSL 4.0 build job so regressions are
   caught before the next release.

5. **`X509_NAME_get_text_by_NID()` is deprecated in 4.0.**
   `X509Certificate::issuerName(NID)` and `subjectName(NID)`
   (`Crypto/src/X509Certificate.cpp`) use it; #5433 replaces it with
   `X509_NAME_get_index_by_NID()` and `ASN1_STRING_to_UTF8()`.

## Related

- #5433 -- addresses items 1 to 5 (contributor PR, continued by the maintainers).
- #5460 -- FIPS fixes, merged; covers most of item 2 and creates item 3.
- #5464 -- restores a `#else` that #5460 mis-nested, and adds an
  OpenSSL 1.1.1 CI job covering the pre-3.0 preprocessor branches.
- #5415 / #5416 -- building against `no-deprecated`, merged.

## Checked, no action required

- ENGINE API removal: Poco does not use `ENGINE_*`.
- Fixed SSL/TLS version method removal (`SSLv3_method()` etc.):
  not used; Poco selects protocols via `SSL_CTX_set_min/max_proto_version`.
- `ASN1_STRING` made opaque: already accessed through
  `ASN1_STRING_get0_data()` / `ASN1_STRING_length()` / `ASN1_STRING_type()`.
- `ERR_get_state()` / `ERR_remove_state()` / `ERR_remove_thread_state()`
  removal: not used.
- `BIO_f_reliable()` removal: not used.
- Deprecated low-level key APIs (`RSA_*`, `EC_KEY_*`, `DH_*`,
  `PEM_*_bio_RSA*`, `SSL_CTX_set_tmp_dh()`): every remaining use in
  `RSAKeyImpl.cpp`, `ECKeyImpl.cpp`, `RSACipherImpl.cpp`, `EVPPKey.cpp`
  and `Context::initDH()` is inside a
  `#else // !POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)` branch, so none of it
  is compiled against 3.0+ or 4.0. The 3.0+ path already uses
  `EVP_PKEY`. The one EC_KEY accessor left on the 3.0+ path
  (`ECKeyImpl.cpp:224-251`) is guarded by
  `#ifndef OPENSSL_NO_DEPRECATED_3_0`. These functions are also still
  present in the 4.0 headers as `OSSL_DEPRECATEDIN_3_0`, not removed.

ssooffiiaannee:
Closes #5420. Continues the work of @ssooffiiaannee, who asked the maintainers to take the PR over; the CI job is based on the original commit. The branch was rebased onto `main`, because #5460 and #5464 had already changed the same FIPS code.

Built and tested with OpenSSL 1.1.1w, 3.5.5, 3.6.4 and 4.0.2 (Linux and macOS; 4.0.2 also with AddressSanitizer and with an active FIPS provider).

### Changes

- **Shutdown with OpenSSL 4.0 (item 1 of #5420): no code change is needed.** OpenSSL 4.0 no longer frees its global data in an `atexit()` handler; the memory is left to the operating system, and `OPENSSL_cleanup()` is meant for applications only, because OpenSSL cannot be initialized again afterwards. `OpenSSLInitializer::uninitialize()` keeps the providers loaded, as before. Measured with LeakSanitizer, valgrind and macOS `leaks` on OpenSSL 3.0.13, 3.5.5, 3.6.4 and 4.0.2: no leak on any version. The comments and the class documentation now state the real reasons: unloading the default provider leaves the process without one (OpenSSL does not activate its fallback provider again), and resetting the pointers without unloading leaks the providers on the next `initialize()`. The earlier note that `OSSL_PROVIDER_unload()` leaks child library contexts did not reproduce on any of these versions. New test `testReinitialize` covers the transition of the reference count to zero.
- **FIPS mode can be enabled again (item 2).** `OpenSSLInitializer::enableFIPSMode()` uses `EVP_default_properties_enable_fips()` with OpenSSL 3.0+. It throws a `CryptoException` if the FIPS provider is not active; the message contains the reasons reported by OpenSSL. The `openSSL.fips` configuration property of `SSLManager` is active again (the property has no `server`/`client` prefix; this is now documented). The `OPENSSL_FIPS` code for OpenSSL 1.0.x is removed. `SSLManager::isFIPSEnabled()` delegates to `OpenSSLInitializer`.
- **Minimum TLS version in FIPS mode (item 3).** `Context::createSSLContext()` no longer has a separate FIPS branch, which skipped `SSL_CTX_set_min_proto_version()`.
- **`X509_NAME_get_text_by_NID()` is deprecated in OpenSSL 4.0.** `X509Certificate::issuerName(NID)`, `subjectName(NID)` and `commonName()` use `X509_NAME_get_index_by_NID()` and `ASN1_STRING_to_UTF8()`.
- **CI (item 4).** `linux-gcc-cmake-openssl4` builds OpenSSL 4.0.2 with the FIPS provider from source (cached, checksum verified), builds Crypto, NetSSL_OpenSSL and JWT with ASan/LSan, checks that Poco is linked against `libcrypto.so.4` from that build, runs the three test suites, and runs the `OpenSSLInitializer` tests with the FIPS provider active.

### Behaviour changes

- `openSSL.fips = true` and `enableFIPSMode(true)` were ignored silently. They now throw if FIPS mode cannot be enabled: no active FIPS provider, or OpenSSL older than 3.0 (which has no FIPS provider). `enableFIPSMode(false)` still does nothing with OpenSSL older than 3.0.
- Certificate name values are no longer cut at 255 bytes, multi-byte ASN.1 string types are converted to UTF-8, and a value with an embedded NUL character is returned in full instead of being cut at the NUL. A value that OpenSSL cannot convert is returned as it is stored.

### Review points of 2026-08-19

1. Dangling provider pointers after unload: the unload is gone, see "Shutdown" above.
2. `CONF_modules_unload(1)` on 4.0: still called on all versions.
3. and 4. FIPS branch in `createSSLContext()` and the missing configuration path: fixed as described.
5. Stale documentation: updated in `SSLManager.h` and `OpenSSLInitializer.h`.

CI points (no matrix `strategy`, version-derived paths, `lib` vs `lib64`, `install_ssldirs`, check against a silent fallback to the system OpenSSL 3): all applied; the job follows `linux-gcc-cmake-openssl111`.

### Not part of this PR

- `Poco::Crypto::DigestEngine` did not check the results of the OpenSSL digest functions; found while testing FIPS mode. Fixed for 1.15.4 in #5488; the same change will be proposed for `main`.
- valgrind reports "uninitialised value" errors inside OpenSSL 4.0.2 (`OBJ_obj2txt()` during `EVP_MD_fetch()`), not present with 3.5.5. No Poco code is involved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -354,6 +354,104 @@ jobs:
         with:
           build-dir: build
 
+  linux-gcc-cmake-openssl4:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+    env:
+      OPENSSL4_VERSION: "4.0.1"
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set OpenSSL 4.0 paths
+        run: |
+          prefix="${GITHUB_WORKSPACE}/.openssl-4.0.1"
+          echo "OPENSSL4_PREFIX=${prefix}" >> "$GITHUB_ENV"
+          echo "OPENSSL_ROOT_DIR=${prefix}" >> "$GITHUB_ENV"
+      - name: Cache OpenSSL 4.0 build
+        uses: actions/cache@v5
+        id: openssl4-cache
+        with:
+          path: ${{ github.workspace }}/.openssl-4.0.1
+          key: openssl4-${{ env.OPENSSL4_VERSION }}-${{ runner.os }}-${{ runner.arch }}-v2
+      - name: Check OpenSSL 4.0 install
+        id: openssl4-ready
+        run: |
+          libdir=""
+          for d in "${OPENSSL4_PREFIX}/lib" "${OPENSSL4_PREFIX}/lib64"; do
+            if [ -f "${d}/libssl.so.4" ]; then
+              libdir="$d"
+              break
+            fi
+          done
+          if [ -n "$libdir" ]; then
+            echo "ready=true" >> "$GITHUB_OUTPUT"
+            echo "libdir=${libdir}" >> "$GITHUB_OUTPUT"
+            echo "LD_LIBRARY_PATH=${libdir}" >> "$GITHUB_ENV"
+            echo "OPENSSL_MODULES=${libdir}/ossl-modules" >> "$GITHUB_ENV"
+            echo "OPENSSL4_LIBDIR=${libdir}" >> "$GITHUB_ENV"
+          else
+            echo "ready=false" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Build OpenSSL 4.0
+        if: steps.openssl4-ready.outputs.ready != 'true'
+        env:
+          LD_LIBRARY_PATH: ""
+        run: |
+          sudo apt -y update
+          sudo apt -y install build-essential curl perl patchelf
+          curl -fsSL -LO "https://www.openssl.org/source/openssl-${OPENSSL4_VERSION}.tar.gz"
+          tar xzf "openssl-${OPENSSL4_VERSION}.tar.gz"
+          cd "openssl-${OPENSSL4_VERSION}"
+          ./config --prefix="${OPENSSL4_PREFIX}" --openssldir="${OPENSSL4_PREFIX}/ssl" shared "-Wl,-rpath,${OPENSSL4_PREFIX}/lib"
+          make -j"$(nproc)"
+          make install_sw
+          libdir="${OPENSSL4_PREFIX}/lib"
+          if [ ! -f "${libdir}/libssl.so.4" ]; then
+            libdir="${OPENSSL4_PREFIX}/lib64"
+          fi
+          if [ ! -f "${libdir}/libssl.so.4" ]; then
+            echo "OpenSSL install incomplete: libssl.so.4 not found under ${OPENSSL4_PREFIX}" >&2
+            find "${OPENSSL4_PREFIX}" -name 'libssl.so*' >&2 || true
+            exit 1
+          fi
+          patchelf --set-rpath "${libdir}" "${OPENSSL4_PREFIX}/bin/openssl"
+          echo "LD_LIBRARY_PATH=${libdir}" >> "$GITHUB_ENV"
+          echo "OPENSSL_MODULES=${libdir}/ossl-modules" >> "$GITHUB_ENV"
+          echo "OPENSSL4_LIBDIR=${libdir}" >> "$GITHUB_ENV"
+      - name: Verify OpenSSL 4.0
+        run: |
+          libdir="${OPENSSL4_PREFIX}/lib"
+          if [ ! -f "${libdir}/libssl.so.4" ]; then
+            libdir="${OPENSSL4_PREFIX}/lib64"
+          fi
+          export LD_LIBRARY_PATH="${libdir}:${LD_LIBRARY_PATH:-}"
+          ls -la "${libdir}/libssl.so"* "${libdir}/libcrypto.so"*
+          "${OPENSSL4_PREFIX}/bin/openssl" version
+      - name: Install build dependencies
+        run: sudo apt -y update && sudo apt -y install cmake ninja-build
+      - run: >-
+          cmake -S. -Bcmake-build -GNinja -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
+          ${{ env.POCO_CMAKE_FOUNDATION_UTIL }}
+          -DENABLE_CRYPTO=ON -DENABLE_NETSSL=ON -DENABLE_JWT=ON
+          -DOPENSSL_ROOT_DIR=${{ env.OPENSSL4_PREFIX }}
+          -DCMAKE_BUILD_RPATH=${{ env.OPENSSL4_LIBDIR }}
+          -DCMAKE_INSTALL_RPATH=${{ env.OPENSSL4_LIBDIR }}
+      - run: >-
+          cmake --build cmake-build --parallel $(nproc) --target
+          Foundation-testrunner Util-testrunner Crypto-testrunner NetSSL-testrunner JWT-testrunner
+      - uses: ./.github/actions/retry-action
+        with:
+          timeout_minutes: 90
+          max_attempts: 3
+          retry_on: any
+          command: >-
+            cd cmake-build &&
+            PWD=`pwd`
+            ctest ${{ env.POCO_CTEST_COMMON }} --parallel $(nproc) -R "^(Foundation|Util|Crypto|NetSSL|JWT)$"
+      - uses: ./.github/actions/upload-test-report
+        with:
+          suffix: openssl4
+
   # Classic GNU make build. Not officially supported anymore, but kept as a
   # sanity check so the legacy ./configure + make path keeps compiling. All
   # standard components are built (--everything); tests are compiled but not
--- a/Crypto/include/Poco/Crypto/OpenSSLInitializer.h
+++ b/Crypto/include/Poco/Crypto/OpenSSLInitializer.h
@@ -21,6 +21,7 @@
 #include "Poco/Crypto/Crypto.h"
 #include "Poco/AtomicCounter.h"
 #include <openssl/crypto.h>
+#include <openssl/evp.h>
 #if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
 #include <openssl/provider.h>
 #include <atomic>
@@ -73,18 +74,21 @@ private:
 //
 inline bool OpenSSLInitializer::isFIPSEnabled()
 {
-#ifdef OPENSSL_FIPS
-	return FIPS_mode() ? true : false;
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+	return EVP_default_properties_is_fips_enabled(NULL);
 #else
 	return false;
 #endif
 }
 
 
-#ifdef OPENSSL_FIPS
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
 inline void OpenSSLInitializer::enableFIPSMode(bool enabled)
 {
-	FIPS_mode_set(enabled);
+	OSSL_LIB_CTX* libctx = OSSL_LIB_CTX_get0_global_default();
+
+	if(libctx != NULL)
+		EVP_default_properties_enable_fips(libctx,enabled);
 }
 #else
 inline void OpenSSLInitializer::enableFIPSMode(bool /*enabled*/)
--- a/Crypto/src/OpenSSLInitializer.cpp
+++ b/Crypto/src/OpenSSLInitializer.cpp
@@ -104,7 +104,10 @@ void OpenSSLInitializer::uninitialize()
 {
 	if (--_rc == 0)
 	{
-#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+#if POCO_OPENSSL_VERSION_PREREQ(4, 0, 0)
+		OSSL_PROVIDER_unload(_defaultProvider);
+		OSSL_PROVIDER_unload(_legacyProvider);
+#elif POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
 		// Provider cleanup is deliberately left to OpenSSL's internal
 		// atexit handler (OPENSSL_cleanup). We must NOT:
 		//  - call OSSL_PROVIDER_unload(): leaks OSSL_LIB_CTX child
--- a/NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
+++ b/NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
@@ -353,11 +353,6 @@ private:
 	static const std::string CFG_DH_PARAMS_FILE;
 	static const std::string CFG_ECDH_CURVE;
 
-#ifdef OPENSSL_FIPS
-	static const std::string CFG_FIPS_MODE;
-	static const bool        VAL_FIPS_MODE;
-#endif
-
 	friend class Poco::SingletonHolder<SSLManager>;
 	friend class Context;
 	friend class SecureSocketImpl;
@@ -381,8 +376,8 @@ inline CertificateHandlerFactoryMgr& SSLManager::certificateHandlerFactoryMgr()
 
 inline bool SSLManager::isFIPSEnabled()
 {
-#ifdef OPENSSL_FIPS
-	return FIPS_mode() ? true : false;
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+	return EVP_default_properties_is_fips_enabled(NULL);
 #else
 	return false;
 #endif
--- a/NetSSL_OpenSSL/src/SSLManager.cpp
+++ b/NetSSL_OpenSSL/src/SSLManager.cpp
@@ -66,10 +66,6 @@ const std::string SSLManager::CFG_REQUIRE_TLSV1_3("requireTLSv1_3");
 const std::string SSLManager::CFG_DISABLE_PROTOCOLS("disableProtocols");
 const std::string SSLManager::CFG_DH_PARAMS_FILE("dhParamsFile");
 const std::string SSLManager::CFG_ECDH_CURVE("ecdhCurve");
-#ifdef OPENSSL_FIPS
-const std::string SSLManager::CFG_FIPS_MODE("openSSL.fips");
-const bool        SSLManager::VAL_FIPS_MODE(false);
-#endif
 
 
 SSLManager::SSLManager():
@@ -414,13 +410,6 @@ void SSLManager::initDefaultContext(bool server)
 	initEvents(server);
 	Poco::Util::AbstractConfiguration& config = appConfig();
 
-#ifdef OPENSSL_FIPS
-	bool fipsEnabled = config.getBool(CFG_FIPS_MODE, VAL_FIPS_MODE);
-	if (fipsEnabled && !Poco::Crypto::OpenSSLInitializer::isFIPSEnabled())
-	{
-		Poco::Crypto::OpenSSLInitializer::enableFIPSMode(true);
-	}
-#endif
 
 	std::string prefix = server ? CFG_SERVER_PREFIX : CFG_CLIENT_PREFIX;
 


Crypto/src/OpenSSLInitializer.cpp
```diff
@@ -104,7 +104,10 @@ void OpenSSLInitializer::uninitialize()
 {
 	if (--_rc == 0)
 	{
-#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+#if POCO_OPENSSL_VERSION_PREREQ(4, 0, 0)
+		OSSL_PROVIDER_unload(_defaultProvider);
+		OSSL_PROVIDER_unload(_legacyProvider);
+#elif POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
```
matejk, line 110: **Blocking.** Two problems in this block.

**1. The static pointers are not reset after unloading.** `initialize()` guards on `if (_defaultProvider == nullptr)` (line 88), so once these are unloaded the pointers stay non-null and dangling, and a later `initialize()` reloads nothing. I reproduced this by forcing this branch on against 3.6:

```
cycle 1 after init    | haveLegacyProvider()=1 | available(default)=1 | available(legacy)=1
cycle 1 after uninit  | haveLegacyProvider()=1 | available(default)=0 | available(legacy)=0
cycle 2 after init    | haveLegacyProvider()=1 | available(default)=0 | available(legacy)=0
```

Note the third line: after re-initializing, the default provider is still not available. `haveLegacyProvider()` also returns `true` for an unloaded provider, which breaks the existing assertion in `Crypto/testsuite/src/OpenSSLInitializerTest.cpp:87`:

```cpp
assertTrue(haveLegacy == OSSL_PROVIDER_available(nullptr, "legacy"));
```

That test does not fail today only because `CryptoInitializer` in the Crypto test driver's `main()` holds a reference for the whole run, so the refcount never reaches zero. CI will not catch this.

Each subsequent `uninitialize()` also drops another provider reference that Poco never took.

**2. `CONF_modules_unload(1)` is no longer called on 4.0.** `initialize()` calls `CONF_modules_load(nullptr, nullptr, 0)` unconditionally (line 85), and 4.0 removed the `atexit()` cleanup, so config modules now leak. `CONF_modules_unload` is still present and not deprecated in 4.0 (`include/openssl/conf.h.in:151`), so nothing forces dropping it.

Suggested shape:

```cpp
#if POCO_OPENSSL_VERSION_PREREQ(4, 0, 0)
		// OpenSSL 4.0 no longer runs OPENSSL_cleanup() from an atexit
		// handler, so the providers must be released explicitly.
		OSSL_PROVIDER_unload(_legacyProvider);
		_legacyProvider = nullptr;
		OSSL_PROVIDER_unload(_defaultProvider);
		_defaultProvider = nullptr;
		CONF_modules_unload(1);
#elif POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
```

Please also add a test covering an uninitialize/re-initialize cycle, and update the note at the end of `OpenSSLInitializerTest.cpp`, which still says cleanup is delegated to the atexit handler.

Crypto/include/Poco/Crypto/OpenSSLInitializer.h
```diff
@@ -73,18 +74,21 @@ class Crypto_API OpenSSLInitializer
 //
 inline bool OpenSSLInitializer::isFIPSEnabled()
 {
-#ifdef OPENSSL_FIPS
-	return FIPS_mode() ? true : false;
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+	return EVP_default_properties_is_fips_enabled(NULL);
```
matejk, line 78: Two small things:

- `NULL` should be `nullptr`. The codebase was converted in #5043 and `CLAUDE.md` requires it.
- `EVP_default_properties_is_fips_enabled()` returns `int`; the code being replaced used the explicit `? true : false` form. Please keep that rather than relying on the implicit conversion.

```suggestion
	return EVP_default_properties_is_fips_enabled(nullptr) ? true : false;
```

Crypto/include/Poco/Crypto/OpenSSLInitializer.h
```diff
@@ -73,18 +74,21 @@ class Crypto_API OpenSSLInitializer
...
 
-#ifdef OPENSSL_FIPS
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
 inline void OpenSSLInitializer::enableFIPSMode(bool enabled)
 {
-	FIPS_mode_set(enabled);
+	OSSL_LIB_CTX* libctx = OSSL_LIB_CTX_get0_global_default();
+
+	if(libctx != NULL)
+		EVP_default_properties_enable_fips(libctx,enabled);
```
matejk, line 91: A few points here:

- `OSSL_LIB_CTX_get0_global_default()` plus the null check is redundant: passing `nullptr` already means the global default library context, which is what `isFIPSEnabled()` does two functions up. The two should be consistent.
- The return value is discarded. `EVP_default_properties_enable_fips()` returns 0 on failure, so a failed FIPS enable is silently swallowed.
- `if(` and `(libctx,enabled)` need spaces per the surrounding style.
- The declaration at line 56 still says "If FIPS is not available, this method doesn't do anything." That is no longer accurate: this sets the default property whether or not the FIPS provider is loaded, so subsequent operations will fail rather than no-op. Please update the doc comment.

```suggestion
inline void OpenSSLInitializer::enableFIPSMode(bool enabled)
{
	if (!EVP_default_properties_enable_fips(nullptr, enabled ? 1 : 0))
		throw CryptoException("Failed to change OpenSSL FIPS mode");
}
```

NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
```diff
@@ -381,8 +376,8 @@ inline CertificateHandlerFactoryMgr& SSLManager::certificateHandlerFactoryMgr()
 
 inline bool SSLManager::isFIPSEnabled()
 {
-#ifdef OPENSSL_FIPS
-	return FIPS_mode() ? true : false;
+#if POCO_OPENSSL_VERSION_PREREQ(3, 0, 0)
+	return EVP_default_properties_is_fips_enabled(NULL);
```
matejk, line 380: **Blocking.** Beyond the `NULL` -> `nullptr` point, this change has a side effect outside the diff.

`isFIPSEnabled()` has effectively returned `false` since OpenSSL 1.0.2, because `OPENSSL_FIPS` is never defined. Making it live activates this branch in `NetSSL_OpenSSL/src/Context.cpp:524`:

```cpp
	if (SSLManager::isFIPSEnabled())
	{
		_pSSLContext = SSL_CTX_new(TLS_method());
	}
	else
	{
		switch (_usage)
		...
```

That path uses `TLS_method()` for both client and server and leaves `minTLSVersion` at 0, so the `SSL_CTX_set_min_proto_version()` call at line 598 is skipped entirely. Anyone with `fips=yes` in their default properties silently loses the minimum-TLS-version floor. This code has been unreachable for a decade and is untested.

Please either fix that branch to set `minTLSVersion` as the `else` path does, or remove it now that FIPS selection no longer needs a different method function.

NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
```diff
@@ -353,11 +353,6 @@ class NetSSL_API SSLManager
 	static const std::string CFG_DH_PARAMS_FILE;
 	static const std::string CFG_ECDH_CURVE;
 
-#ifdef OPENSSL_FIPS
-	static const std::string CFG_FIPS_MODE;
-	static const bool        VAL_FIPS_MODE;
-#endif
-
```
matejk, line 360: Removing these is fine on its own (they were never compiled with OpenSSL 1.1.1+, so this is not an ABI break in practice), but the class documentation at lines 150-151 still describes the property they backed:

```
///    - fips: Enable or disable OpenSSL FIPS mode. Only supported if the OpenSSL version
///      that this library is built against supports FIPS mode.
```

That needs to go too, or to be updated if the configuration wiring is restored.

NetSSL_OpenSSL/src/SSLManager.cpp
```diff
@@ -414,13 +410,6 @@ void SSLManager::initDefaultContext(bool server)
 	initEvents(server);
 	Poco::Util::AbstractConfiguration& config = appConfig();
 
-#ifdef OPENSSL_FIPS
-	bool fipsEnabled = config.getBool(CFG_FIPS_MODE, VAL_FIPS_MODE);
-	if (fipsEnabled && !Poco::Crypto::OpenSSLInitializer::isFIPSEnabled())
-	{
-		Poco::Crypto::OpenSSLInitializer::enableFIPSMode(true);
-	}
-#endif
```
matejk, line 423: This was the only caller of `OpenSSLInitializer::enableFIPSMode()` in the whole library. With it gone, the method is reachable only from application code, so the PR makes the getter live while removing the only way to turn FIPS on through configuration.

#5420 did offer "or remove" as an option, but this is half of each. Please pick one:

- Restore this block unconditionally (it no longer needs `#ifdef OPENSSL_FIPS`, since `enableFIPSMode()` now works on 3.0+), keeping `CFG_FIPS_MODE` / `VAL_FIPS_MODE`; or
- Remove the FIPS API surface as a whole and record the removal in the release notes.

NetSSL_OpenSSL/src/SSLManager.cpp
```diff
@@ -414,13 +410,6 @@ void SSLManager::initDefaultContext(bool server)
...
```
matejk, line 413: Leftover double blank line from the removed block.

.github/workflows/ci.yml
```diff
@@ -354,6 +354,104 @@ jobs:
         with:
           build-dir: build
 
+  linux-gcc-cmake-openssl4:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
```
matejk, line 360: `strategy: fail-fast: false` has no effect without a matrix -- please drop it.

Related: `suffix: openssl4` on the `upload-test-report` step at line 453 is also unnecessary. Per that action's own documentation the suffix is "Required for matrix jobs, where github.job alone is not unique"; here `github.job` is already `linux-gcc-cmake-openssl4`.

.github/workflows/ci.yml
```diff
@@ -354,6 +354,104 @@ jobs:
...
+    env:
+      OPENSSL4_VERSION: "4.0.1"
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set OpenSSL 4.0 paths
+        run: |
+          prefix="${GITHUB_WORKSPACE}/.openssl-4.0.1"
```
matejk, line 367: `OPENSSL4_VERSION` is defined just above, but the directory name is hardcoded here and again in the cache `path:` at line 374. Bumping the variable would silently keep using the old prefix and cache entry. Please derive both from the variable, for example:

```
prefix="${GITHUB_WORKSPACE}/.openssl-${OPENSSL4_VERSION}"
```

and use a `path:` that matches.

.github/workflows/ci.yml
```diff
@@ -354,6 +354,104 @@ jobs:
...
+          LD_LIBRARY_PATH: ""
+        run: |
+          sudo apt -y update
+          sudo apt -y install build-essential curl perl patchelf
+          curl -fsSL -LO "https://www.openssl.org/source/openssl-${OPENSSL4_VERSION}.tar.gz"
+          tar xzf "openssl-${OPENSSL4_VERSION}.tar.gz"
+          cd "openssl-${OPENSSL4_VERSION}"
+          ./config --prefix="${OPENSSL4_PREFIX}" --openssldir="${OPENSSL4_PREFIX}/ssl" shared "-Wl,-rpath,${OPENSSL4_PREFIX}/lib"
+          make -j"$(nproc)"
+          make install_sw
```
matejk, line 407: Two issues in this step:

- The rpath passed to `./config` hardcodes `${OPENSSL4_PREFIX}/lib`, while the rest of the script carefully handles both `lib` and `lib64`. If the install lands in `lib64` the baked-in rpath is wrong; it currently works only because `LD_LIBRARY_PATH` is set, and the `patchelf` fix below is applied to the `openssl` binary only.
- `make install_sw` installs the software but not the ssl directories, so no `openssl.cnf` is written to `${OPENSSL4_PREFIX}/ssl`. That is survivable for the current tests, but it means the job runs without a config file, which is not representative. Consider `make install_sw install_ssldirs`.

.github/workflows/ci.yml
```diff
@@ -354,6 +354,104 @@ jobs:
...
+          export LD_LIBRARY_PATH="${libdir}:${LD_LIBRARY_PATH:-}"
+          ls -la "${libdir}/libssl.so"* "${libdir}/libcrypto.so"*
+          "${OPENSSL4_PREFIX}/bin/openssl" version
+      - name: Install build dependencies
+        run: sudo apt -y update && sudo apt -y install cmake ninja-build
+      - run: >-
+          cmake -S. -Bcmake-build -GNinja -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
+          ${{ env.POCO_CMAKE_FOUNDATION_UTIL }}
+          -DENABLE_CRYPTO=ON -DENABLE_NETSSL=ON -DENABLE_JWT=ON
+          -DOPENSSL_ROOT_DIR=${{ env.OPENSSL4_PREFIX }}
```
matejk, line 436: ubuntu-24.04 ships OpenSSL 3, so it would be good to fail loudly if CMake ever picks up the system library instead of the one built here. A check after configure -- asserting the resolved `OPENSSL_VERSION` starts with 4, or checking `ldd` on a built test runner -- would make this job trustworthy. Right now a silent fallback to OpenSSL 3 would look like a passing OpenSSL 4 job.

matejk (changes requested):
Thanks for taking this on -- the FIPS direction is the right one and matches what #5420 asked for. However, I cannot merge this as it stands: the provider-cleanup change, which is the core of the issue, introduces a bug that is worse than the leak it is meant to fix, and the CI job that would have caught it has never run (no checks reported on the branch).

## What I verified locally

Built the branch against OpenSSL 3.6.2 with Crypto + NetSSL + JWT. It compiles clean and Crypto (47 tests) and JWT (33 tests) both pass, so there is no regression on the 3.x path.

I then forced the 4.0 branch on (the logic is version-independent, `OSSL_PROVIDER_unload` exists in 3.0 too) and exercised an init/uninit/init cycle:

```
cycle 1 after init    | haveLegacyProvider()=1 | available(default)=1 | available(legacy)=1
cycle 1 after uninit  | haveLegacyProvider()=1 | available(default)=0 | available(legacy)=0
cycle 2 after init    | haveLegacyProvider()=1 | available(default)=0 | available(legacy)=0
```

After a full uninitialize, `initialize()` no longer reloads the providers, and `haveLegacyProvider()` reports `true` for a provider that is unloaded. Details inline.

## Summary

Blocking:

1. Providers are unloaded but the static pointers are not reset, so `initialize()` silently stops reloading them and `haveLegacyProvider()` lies. See `Crypto/src/OpenSSLInitializer.cpp`.
2. `CONF_modules_unload(1)` is no longer called on 4.0, while `initialize()` still calls `CONF_modules_load()` unconditionally. `CONF_modules_unload` is still present and non-deprecated in 4.0 (`conf.h.in:151`), so this is an avoidable leak of exactly the kind the PR is fixing.
3. Making `SSLManager::isFIPSEnabled()` live activates a branch in `Context::createSSLContext()` that has been unreachable since OpenSSL 1.0.2 and skips the minimum-TLS-version floor.

Design:

4. The FIPS configuration feature is removed rather than migrated: `enableFIPSMode()` is left with no caller anywhere in the library. #5420 allowed "or remove", but this is half of each -- the getter goes live while the only configuration path to set it disappears. Please either restore the config wiring (it no longer needs `#ifdef OPENSSL_FIPS`) or drop the FIPS API entirely and note it in the release notes.
5. Documentation is now stale: `SSLManager.h:150-151` still documents the deleted `fips` property, and `OpenSSLInitializer.h:56` still says "If FIPS is not available, this method doesn't do anything", which is no longer true.

Conventions: `NULL` should be `nullptr` (the codebase was converted in #5043), plus some spacing and a leftover blank line -- all noted inline.

CI: please make sure the new job actually runs and goes green before this is merged. As it stands nothing in the PR demonstrates that Poco compiles or passes tests against OpenSSL 4.0.1, which is the whole point of the change.

One thing worth confirming: `POCO_OPENSSL_VERSION_PREREQ(4, 0, 0)` is correct here. OpenSSL 3.0+ encodes `OPENSSL_VERSION_NUMBER` as `(major<<28)|(minor<<20)|(patch<<4)`, so 4.0.1 is `0x40000010` and 3.6.x stays below the threshold.

matejk:
Hi @ssooffiiaannee — heads up before you pick this back up: main has moved since the review. #5460 (merged Aug 21) independently made isFIPSEnabled() live for OpenSSL 3.0+ and touches the same lines in OpenSSLInitializer.h/SSLManager.h this PR does, so it'll need a rebase and will conflict there.

That also changes review point 3 above — the "activates a dead branch" concern is now true on main regardless of this PR, so it's not something this PR need fix on its own. Points 1, 2, 4 and 5 still stand, and point 4 is sharper now: enableFIPSMode() and its config wiring (SSLManager.cpp) are still #ifdef OPENSSL_FIPS-only, so there's currently no way to enable FIPS via config on OpenSSL 3.0+/4.0 even though it can be queried. Worth deciding whether this PR restores that path or formally drops it.

Let us know if you're still able to pick this up.

ssooffiiaannee:
Hi @matejk, thank you so much for all the correction and improvement suggestions.
I may not be able to pick this up in the short term, so please feel free to make the necessary corrections directly.

Crypto/src/OpenSSLInitializer.cpp
matejk: The provider unload is gone. Measured with LeakSanitizer, valgrind and macOS `leaks` on OpenSSL 3.0.13, 3.5.5, 3.6.4 and 4.0.2: keeping the providers loaded, as `uninitialize()` did before, is leak-free on all of them, and with 4.0 `OPENSSL_cleanup()` is meant for applications only. `CONF_modules_unload(1)` is still called on every version. The comments now give the real reasons (unloading the default provider leaves the process without one; resetting the pointers without unloading leaks on the next `initialize()`). `testReinitialize` takes the reference count through zero, and the stale note in `OpenSSLInitializerTest.cpp` is replaced. c310e8618

Crypto/include/Poco/Crypto/OpenSSLInitializer.h
matejk: Taken over by #5460 on main, which already has `EVP_default_properties_is_fips_enabled(nullptr) ? true : false`; the branch was rebased onto it.

Crypto/include/Poco/Crypto/OpenSSLInitializer.h
matejk: `enableFIPSMode()` now calls `EVP_default_properties_enable_fips(nullptr, ...)` and throws a `CryptoException` when the FIPS provider is not available or OpenSSL rejects the change; the message contains the OpenSSL errors and the error queue is left empty. With OpenSSL < 3.0 enabling throws and disabling does nothing. The doc comment describes this. c77010789

NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
matejk: `createSSLContext()` no longer has a FIPS branch, so the minimum TLS version is always set. 4d4605831

NetSSL_OpenSSL/include/Poco/Net/SSLManager.h
matejk: The property is kept and documented under its real name `openSSL.fips` (no server/client prefix): it needs OpenSSL 3.0+ and an activated FIPS provider, otherwise creating the default context fails with a `CryptoException`. c77010789

NetSSL_OpenSSL/src/SSLManager.cpp
matejk: Restored unconditionally: `CFG_FIPS_MODE`, `VAL_FIPS_MODE` and the block in `initDefaultContext()` no longer depend on `OPENSSL_FIPS`. c77010789

NetSSL_OpenSSL/src/SSLManager.cpp
matejk: Gone together with the restored block.

.github/workflows/ci.yml
matejk: Removed; the job now follows `linux-gcc-cmake-openssl111`. 84d9216db

.github/workflows/ci.yml
matejk: The prefix and the cache path and key (`openssl4-fips-<version>-<os>-<arch>`) are derived from `OPENSSL4_VERSION`. 84d9216db

.github/workflows/ci.yml
matejk: `--libdir=lib` removes the lib/lib64 case (the rpath argument and the `patchelf` step are gone), and the install is `make install_sw install_ssldirs install_fips`. 84d9216db

.github/workflows/ci.yml
matejk: The job fails unless `openssl version` matches `^OpenSSL 4[.]` and `ldd` with `LD_LIBRARY_PATH` unset resolves `libcrypto.so.4` from the job's own build. First run: OpenSSL 4.0.2 built with the FIPS provider, Crypto/NetSSL/JWT green under ASan, and the `OpenSSLInitializer` tests green with the FIPS provider active. 84d9216db

matejk:
Continued here as discussed above: the branch was rebased onto main (#5460 and #5464 had changed the same FIPS code), and all points of the review of 2026-08-19 are addressed; the replies in the threads name the commits, and the description lists the changes.

CI is green, including the first run of `linux-gcc-cmake-openssl4`: OpenSSL 4.0.2 built from source with the FIPS provider, Crypto, NetSSL and JWT with AddressSanitizer, and the `OpenSSLInitializer` tests with the FIPS provider active.

PR #5438, opened 2026-08-15
fix(File): honor supplementary groups in permission checks

### LINKED ISSUE #5361: fix(File): canRead/canWrite/canExecute on POSIX ignore supplementary groups
On POSIX, `FileImpl::canReadImpl`, `canWriteImpl` and `canExecuteImpl`
in `Foundation/src/File_UNIX.cpp` check only the effective gid:

```cpp
else if (st.st_gid == ::getegid())
    return (st.st_mode & S_IRGRP) != 0;
```

This ignores supplementary group membership, so a user in a
secondary group that owns the file gets a false negative.

### Reproducer

```
$ id
uid=1000(alice) gid=1000(alice) groups=1000(alice),27(sudo)
$ ls -l /etc/sudoers.d/foo
-rw-r----- 1 root sudo 42 ... /etc/sudoers.d/foo
$ # Poco::File("/etc/sudoers.d/foo").canRead() returns false
$ # POSIX access(path, R_OK) returns 0 (readable)
```

### History

Previously reported in #301 and closed citing TOCTOU on `access(2)`.
The objection does not address the underlying bug: the current
`stat`-based code has the same TOCTOU window, and three callers
silently return the wrong answer for any user whose membership is
via a supplementary group.

### Proposed fix

Keep the `stat`-based structure (no TOCTOU regression vs today),
but extend the group check to consult `getgroups(2)`:

```cpp
static bool inGroup(gid_t gid)
{
    if (gid == ::getegid()) return true;
    int n = ::getgroups(0, nullptr);
    if (n <= 0) return false;
    std::vector<gid_t> groups(n);
    if (::getgroups(n, groups.data()) < 0) return false;
    return std::find(groups.begin(), groups.end(), gid) != groups.end();
}
```

and replace `st.st_gid == ::getegid()` with `inGroup(st.st_gid)` in
all three functions.

Note: this still does not honor ACLs; callers that need full
effective-permission semantics should use `access(2)` directly.

m0g3r:
## Summary

- include supplementary group membership when evaluating POSIX file permissions
- retry the `getgroups()` query if the group list changes between sizing and retrieval
- cover effective, supplementary, and non-member group IDs on Unix

Fixes #5361.

## Testing

- Linux aarch64, GCC 13: reproduced current `main` returning false for `canRead()`, `canWrite()`, and `canExecute()` on a root-owned, GID-27, mode-0070 file from `euid=1000`, `egid=1000`, supplementary group 27; the patched branch returns true for all three
- `Foundation-testrunner FileTest::testGroupMembership`
- macOS Foundation test suite: 935 tests passed
- Linux Foundation test suite: 934/935 passed in a Docker container without an init process; the sole unrelated `ProcessRunnerTest::testKillTreeWithChild` timeout passed separately under `tini -s`

### DIFF
--- a/Foundation/src/File_UNIX.cpp
+++ b/Foundation/src/File_UNIX.cpp
@@ -38,6 +38,9 @@
 #include <utime.h>
 #include <cstring>
 #include <vector>
+#if POCO_OS == POCO_OS_MAC_OS_X
+#include <membership.h>
+#endif
 
 #if (POCO_OS == POCO_OS_SOLARIS) || (POCO_OS == POCO_OS_QNX)
 #define STATFSFN ::statvfs
@@ -134,22 +137,67 @@ bool FileImpl::existsImpl() const
 }
 
 
-bool FileImpl::isGroupMemberImpl(gid_t gid)
+namespace
 {
-	if (gid == ::getegid()) return true;
-
-	for (int attempt = 0; attempt < 2; ++attempt)
+	bool isGroupMember(gid_t gid)
+		/// Returns true if the effective user belongs to gid, counting supplementary
+		/// groups. This mirrors access(2): the kernel tests the effective GID and every
+		/// supplementary group, not the effective GID alone.
 	{
-		const int count = ::getgroups(0, nullptr);
-		if (count <= 0) return false;
+		if (gid == ::getegid()) return true;
 
-		std::vector<gid_t> groups(count);
-		const int actualCount = ::getgroups(count, groups.data());
-		if (actualCount >= 0)
-			return std::find(groups.begin(), groups.begin() + actualCount, gid) != groups.begin() + actualCount;
-		if (errno != EINVAL) return false;
+		// A permission query must not disturb the caller's errno.
+		const int savedErrno = errno;
+		bool member = false;
+
+#if POCO_OS == POCO_OS_MAC_OS_X
+		// getgroups() is capped at NGROUPS_MAX (16) on Darwin and truncates silently,
+		// so a user in more groups than that would look like a non-member. The
+		// membership resolver consults Directory Services and has no such cap.
+		uuid_t userUuid;
+		uuid_t groupUuid;
+		int isMember = 0;
+		if (::mbr_uid_to_uuid(::geteuid(), userUuid) == 0
+			&& ::mbr_gid_to_uuid(gid, groupUuid) == 0
+			&& ::mbr_check_membership(userUuid, groupUuid, &isMember) == 0)
+		{
+			errno = savedErrno;
+			return isMember != 0;
+		}
+		// Resolver unavailable: fall through to getgroups().
+#endif
+
+		enum { INLINE_GROUPS = 64 };
+		gid_t inlineGroups[INLINE_GROUPS];
+
+		// getgroups() reports a group set that grew between the sizing call and the
+		// fetch as EINVAL. One retry is enough: losing that race twice yields false,
+		// the same answer further attempts would settle on.
+		for (int attempt = 0; attempt < 2; ++attempt)
+		{
+			const int count = ::getgroups(0, nullptr);
+			if (count <= 0) break;
+
+			gid_t* groups = inlineGroups;
+			std::vector<gid_t> heapGroups;
+			if (count > INLINE_GROUPS)
+			{
+				heapGroups.resize(count);
+				groups = heapGroups.data();
+			}
+
+			const int actualCount = ::getgroups(count, groups);
+			if (actualCount >= 0)
+			{
+				member = std::find(groups, groups + actualCount, gid) != groups + actualCount;
+				break;
+			}
+			if (errno != EINVAL) break;
+		}
+
+		errno = savedErrno;
+		return member;
 	}
-	return false;
 }
 
 
@@ -160,12 +208,16 @@ bool FileImpl::canReadImpl() const
 	struct stat st;
 	if (::stat(_path.c_str(), &st) == 0)
 	{
-		if (st.st_uid == ::geteuid())
+		if (::geteuid() == 0)
+			return true;
+		else if (st.st_uid == ::geteuid())
 			return (st.st_mode & S_IRUSR) != 0;
-		else if (isGroupMemberImpl(st.st_gid))
+		else if (((st.st_mode & S_IRGRP) != 0) == ((st.st_mode & S_IROTH) != 0))
+			return (st.st_mode & S_IROTH) != 0;
+		else if (isGroupMember(st.st_gid))
 			return (st.st_mode & S_IRGRP) != 0;
 		else
-			return (st.st_mode & S_IROTH) != 0 || ::geteuid() == 0;
+			return (st.st_mode & S_IROTH) != 0;
 	}
 	else if (const auto err = errno; err == ENOENT)
 		return false;
@@ -182,12 +234,16 @@ bool FileImpl::canWriteImpl() const
 	struct stat st;
 	if (::stat(_path.c_str(), &st) == 0)
 	{
-		if (st.st_uid == ::geteuid())
+		if (::geteuid() == 0)
+			return true;
+		else if (st.st_uid == ::geteuid())
 			return (st.st_mode & S_IWUSR) != 0;
-		else if (isGroupMemberImpl(st.st_gid))
+		else if (((st.st_mode & S_IWGRP) != 0) == ((st.st_mode & S_IWOTH) != 0))
+			return (st.st_mode & S_IWOTH) != 0;
+		else if (isGroupMember(st.st_gid))
 			return (st.st_mode & S_IWGRP) != 0;
 		else
-			return (st.st_mode & S_IWOTH) != 0 || ::geteuid() == 0;
+			return (st.st_mode & S_IWOTH) != 0;
 	}
 	else if (const auto err = errno; err == ENOENT)
 		return false;
@@ -209,7 +265,9 @@ bool FileImpl::canExecuteImpl(const std::string& absolutePath) const
 		return (st.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH)) != 0;
 	else if (st.st_uid == ::geteuid())
 		return (st.st_mode & S_IXUSR) != 0;
-	else if (isGroupMemberImpl(st.st_gid))
+	else if (((st.st_mode & S_IXGRP) != 0) == ((st.st_mode & S_IXOTH) != 0))
+		return (st.st_mode & S_IXOTH) != 0;
+	else if (isGroupMember(st.st_gid))
 		return (st.st_mode & S_IXGRP) != 0;
 	else
 		return (st.st_mode & S_IXOTH) != 0;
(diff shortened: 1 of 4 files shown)


matejk:
@m0g3r please rebase PR on latest main.

Foundation/src/File_UNIX.cpp
```diff
@@ -142,7 +162,7 @@ bool FileImpl::canReadImpl() const
 	{
 		if (st.st_uid == ::geteuid())
 			return (st.st_mode & S_IRUSR) != 0;
-		else if (st.st_gid == ::getegid())
+		else if (isGroupMemberImpl(st.st_gid))
 			return (st.st_mode & S_IRGRP) != 0;
 		else
 			return (st.st_mode & S_IROTH) != 0 || ::geteuid() == 0;
```
matejk: For root this is a regression: the `|| ::geteuid() == 0` override exists only in the else branch (added in 0ca3bbc84 for SF#3478665 "Permission checks in Poco::File not correct for root"), and widening the group check diverts root into the group branch whenever the file's gid is one of root's supplementary groups. Example: euid 0 with supplementary group `adm`; file uid 1000, gid `adm`, mode 0600 -> `canRead()` now returns false while `open()` succeeds. The kernel grants root read/write regardless of the mode bits, so check root first, as `canExecuteImpl` already does:

```suggestion
		if (::geteuid() == 0)
			return true;
		else if (st.st_uid == ::geteuid())
			return (st.st_mode & S_IRUSR) != 0;
		else if (isGroupMemberImpl(st.st_gid))
			return (st.st_mode & S_IRGRP) != 0;
		else
			return (st.st_mode & S_IROTH) != 0;
```

Foundation/src/File_UNIX.cpp
```diff
@@ -164,7 +184,7 @@ bool FileImpl::canWriteImpl() const
 	{
 		if (st.st_uid == ::geteuid())
 			return (st.st_mode & S_IWUSR) != 0;
-		else if (st.st_gid == ::getegid())
+		else if (isGroupMemberImpl(st.st_gid))
 			return (st.st_mode & S_IWGRP) != 0;
 		else
 			return (st.st_mode & S_IWOTH) != 0 || ::geteuid() == 0;
```
matejk: Same root regression as in `canReadImpl`: a file with mode 0644, a foreign uid, and a gid in root's supplementary groups now returns false for root. Check root first here as well:

```suggestion
		if (::geteuid() == 0)
			return true;
		else if (st.st_uid == ::geteuid())
			return (st.st_mode & S_IWUSR) != 0;
		else if (isGroupMemberImpl(st.st_gid))
			return (st.st_mode & S_IWGRP) != 0;
		else
			return (st.st_mode & S_IWOTH) != 0;
```

Foundation/src/File_UNIX.cpp
```diff
@@ -133,6 +134,25 @@ bool FileImpl::existsImpl() const
 }
 
 
+bool FileImpl::isGroupMemberImpl(gid_t gid)
+{
+	if (gid == ::getegid()) return true;
+
+	for (int attempt = 0; attempt < 2; ++attempt)
+	{
+		const int count = ::getgroups(0, nullptr);
```
matejk: On macOS this is capped: `getgroups()` returns at most `NGROUPS_MAX` (16) entries unless `_DARWIN_UNLIMITED_GETGROUPS` is defined before including `<unistd.h>`, and Apple documents the truncated list as unsuitable for access-control decisions (directory-service group memberships are not reflected). A user in more than 16 groups still gets the wrong answer for any group beyond the cap, and the new test cannot detect it because it reads the same capped list. On Darwin please resolve membership via `getgrouplist(3)` or `mbr_check_membership()` instead.

Foundation/testsuite/src/FileTest.cpp
```diff
@@ -40,6 +41,24 @@ using Poco::Timestamp;
 using Poco::Thread;
 
 
+#if defined(POCO_OS_FAMILY_UNIX)
+namespace {
```
matejk, line 53: This guard breaks the VxWorks testsuite build: `Platform.h` defines `POCO_OS_FAMILY_UNIX` for VxWorks, but `File.h` selects `File_VX.h` there, whose `FileImpl` has no `isGroupMemberImpl` (and `getgroups` is unavailable). Please use `#if defined(POCO_OS_FAMILY_UNIX) && !defined(POCO_VXWORKS)` here and at the `testGroupMembership` definition below. The new declaration in FileTest.h and the registration in `suite()` sit inside guards shared with `testGetExecutablePathThreadSafety` (which does build on VxWorks), so those two need a nested `#if !defined(POCO_VXWORKS)` around just the new lines.

Foundation/testsuite/src/FileTest.cpp
```diff
@@ -277,6 +296,32 @@ void FileTest::testFileAttributes3()
 }
 
 
+#if defined(POCO_OS_FAMILY_UNIX)
+void FileTest::testGroupMembership()
```
matejk: This verifies the helper against the same `getgroups` list the helper itself reads, so it cannot fail for any `getgroups`-based implementation -- and nothing covers the changed group branch in `canReadImpl`/`canWriteImpl`/`canExecuteImpl`: reverting the three call sites to `st.st_gid == ::getegid()` leaves the whole suite green. Please add a facade-level test: pick a gid from `getgroups()` that differs from `getegid()` (skip the test when there is none, or when running as root), `::chown()` a `TemporaryFile` to that gid (permitted unprivileged when the caller owns the file and the gid is in its own group set), then assert both directions through the public API: mode 0040 -> `File::canRead()` true (the fix), mode 0604 -> `File::canRead()` false (group bits take precedence over the other bits).

matejk (changes requested):
Thanks, the supplementary-group extension itself is correct and matches kernel (`access(2)`) semantics: group membership takes precedence over the "other" bits, so files whose group bits are stricter than their other bits (e.g. mode 0604) now correctly report `false` for group members where they previously reported `true`. That direction is intended but breaking; I have labeled the PR and #5361 `breaking`, release notes will cover it during release preparation.

Requested changes (details inline):

1. Root regression in `canReadImpl`/`canWriteImpl`: the `|| ::geteuid() == 0` override lives only in the else branch, so root is now denied on files whose gid is one of its supplementary groups. Hoist the root check, as `canExecuteImpl` already does.
2. macOS: `getgroups()` is capped at `NGROUPS_MAX` (16), so membership beyond the cap stays invisible and the fix is incomplete on Darwin. Use `getgrouplist(3)`/`mbr_check_membership()` there.
3. The new test code breaks the VxWorks testsuite build (`POCO_OS_FAMILY_UNIX` is defined for VxWorks, but VxWorks uses `File_VX.h`, which has no `isGroupMemberImpl`).
4. Please add facade-level tests through `File::canRead()`/`canWrite()` covering both directions; `testGroupMembership` only compares the helper against the same `getgroups` list the helper reads, so it cannot catch a regression in the three call sites.

Non-blocking notes:

- All `getgroups` failure paths silently collapse to "not a member", and the helper leaves `errno` set on the success path of the three public methods; consider saving/restoring `errno`.
- The two-attempt `EINVAL` retry deserves a short comment explaining the grow-race it guards against; alternatively it can be dropped, since losing the race twice returns `false` either way.
- Every non-owner check now costs two `getgroups` syscalls plus a heap allocation, multiplied by `getExecutablePathImpl`'s PATH loop. A fixed stack buffer, and skipping the membership test when the group bit equals the other bit (valid once the root check is hoisted), would remove most of that.
- `Foundation_API` on a single member of the otherwise unexported `FileImpl` adds an internal helper to the shared-library ABI only for the test; with facade-level tests the helper can become a file-local function in File_UNIX.cpp and the header change disappears.

m0g3r:
All four are addressed in `e2568c4b9`, rebased onto current `main` (`888bb406a`).

**1. Root regression.** Hoisted to the top of both `canReadImpl` and `canWriteImpl`, in the same
shape `canExecuteImpl` already uses. You were right that this was reachable: once the group branch
widened to supplementary groups, root landed in it for any file whose gid is one of root's groups
and got the group bits instead of an override. Read and write now return `true` for root before any
mode inspection, which is what the old `|| ::geteuid() == 0` tail meant to express.

**2. macOS.** The Darwin path now goes through `mbr_uid_to_uuid` / `mbr_gid_to_uuid` /
`mbr_check_membership` from `<membership.h>`, so membership past `NGROUPS_MAX` resolves. If the
resolver is unavailable it falls through to `getgroups()` rather than failing closed.

**3. VxWorks.** Fixed by removing the cause rather than guarding it: the helper is now a file-local
function in `File_UNIX.cpp` instead of a `FileImpl` member, so nothing in the testsuite reaches for
it and `File_VX.h` is unaffected.

That also resolves your last non-blocking note — `Foundation/include/Poco/File_UNIX.h` is now
**byte-identical to `main`**, so the `Foundation_API` export is gone from the ABI and the
`[[nodiscard]]` collision with #5156 that the previous rebase had to resolve no longer exists.

**4. Facade-level tests.** `testGroupMembership` is replaced by `testPermissionsMatchAccess`, which
compares `File::canRead()`, `canWrite()` and `canExecute()` against `access(2)` across 22 modes
(`0000`, `0400`, `0604`, `0640`, `0064`, `0046`, `0620`, `0602`, `0755`, …). `access(2)` is the
kernel answering the same question these three ask, so it is the oracle, and each mode is asserted
in whichever direction the kernel decides — a wrongly granted permission and a wrongly denied one
both fail.

I checked the test is not vacuous rather than assuming it: with `S_IRUSR` changed to `S_IRGRP` in
the `canReadImpl` owner branch it fails at the `canRead()` assertion, and passes with that reverted.

One limitation stated plainly. The group branch is only reachable for a file the caller does not
own — for an owned file the owner branch answers first — and a test cannot create one without
privilege. So the test looks for a regular file under `/usr/bin`, `/usr/sbin`, `/etc` or `/usr/lib`
owned by another user whose group is one of the caller's supplementary groups, and asserts against
`access(2)` if it finds one. When the machine has no such file, or the caller is root, that half
skips rather than pretending to cover it.

**Non-blocking notes, all applied:** `errno` is saved and restored across the query; the two-attempt
`EINVAL` retry carries a comment explaining the grow-race and why losing it twice is a settled
`false`; the group list goes into a 64-entry stack buffer and only heap-allocates beyond that; and
the membership call is skipped entirely when the group and other bits agree, since it cannot change
the answer then — valid now that the root check is hoisted, as you noted.

**What I verified, and what I did not.** Built and ran on macOS arm64 (Apple clang 21, CMake
Release, Foundation + testsuite): clean build with no new warnings, `Foundation-testrunner FileTest`
`OK (32 tests)`. So the Darwin branch above is compiled and exercised. I could **not** run as root,
so the item-1 fix is verified by construction rather than execution, and the `getgroups()` fallback
path is not compiled by a macOS build — CI covers both, and I would rather flag that than imply a
run I did not do.

## AI assistance

Claude Code authored this change and this comment. It is agent-authored and has not had a separate
human review. Every result quoted above is from a run actually performed, not an estimate.

Issue #5429, opened 2026-08-04
Extend SQLParser T-SQL dialect coverage


sreckovic1987:
**Problem description:**
`SELECT TOP (10) ...` fails to parse because `TOP` only accepts a bare
integer literal today, not a parenthesized one. The same gap applies to a
few other common, valid T-SQL SELECT forms the grammar doesn't recognize.

**Solution:**
Extend the grammar/lexer to correctly parse:
1. **`SELECT TOP (10) ...`**  parenthesized TOP (only bare
   `TOP int_literal` is currently accepted)
2. **`SELECT ... OFFSET n ROWS FETCH NEXT m ROWS ONLY`**  standard paging
   syntax (`FETCH`/`NEXT`/`ONLY` aren't recognized keywords currently)
3. **`SELECT * FROM [dbo].[Table]`**  bracket-quoted identifiers (only
   double-quoted identifiers are currently supported)
4. **`SELECT * FROM db.dbo.Table`**  three-part (database.schema.table)
   names (only 2-part `schema.table` is currently accepted)

New tests covering these forms will be added as part of the fix.

**Alternatives considered**
Leaving these unsupported works in the sense that `parse()` just fails, but
a syntactically valid SELECT then goes undetected as such, the parser's
narrow dialect coverage causes it to reject SQL that is otherwise correct.
Pre-processing/rewriting T-SQL input before handing it to the parser was
considered but is more fragile than extending the grammar directly.

PR #5442, opened 2026-08-18
5441 extend sqlparser dialect coverage for t sql and oracle select statements

### LINKED ISSUE #5441: Extend SQLParser dialect coverage for T-SQL and Oracle SELECT statements
**Problem description:**
Following up on #5429, a further set of valid SELECT statements still fails to
parse, which leaves them unclassified for callers that rely on `parse()` +
`isSelect()` to tell reads from writes. Production deployments require correct
SELECT detection for the forms listed below - they are all valid SQL Server or
Oracle SELECT statements in active use.

One of the gaps is worse than a plain parse failure: scientific notation is not
lexed at all, so `SELECT 1.5e3 FROM t` parses successfully as
`SELECT 1.5 AS e3` - a silently different statement rather than an error.

**Solution:**
Extend the grammar/lexer to correctly parse:

1. `SELECT 1.5e3`, `WHERE x = 0.0e+21` scientific notation numeric literals
   (the lexer has no exponent rule at all: in a select list the exponent is
   consumed as an alias, elsewhere it is a syntax error)
2. `DATEADD(MINUTE, -10, GETDATE())`, `ISNULL(a, 0)`, `CHAR(10)`,
   `CONVERT(INT, x)`, `CONVERT(DATETIME, x)`, `FORMAT(x, '00')` - non-reserved
   keywords used as identifiers, function names and function arguments
   (`SECOND`/`MINUTE`/`HOUR`/`DAY`/`MONTH`/`YEAR` and their plural forms,
   `ISNULL`, `CHAR`, `INT`, `INTEGER`, `VARCHAR`, `DATETIME`, `TIMESTAMP`,
   `FORMAT`, `NEXT`). The same date part written as an abbreviation
   (`DATEADD(mi, ...)`) parses today, so the behaviour depends on spelling.
3. `SELECT col AS 'Display Name'` string literal used as an alias (only
   identifiers are accepted today)
4. `SELECT db.dbo.Table.Column`, `SELECT schema.seq.NEXTVAL` multi-part column
   references (only `column` and `table.column` are accepted, while #5429
   already added multi-part names on the table side)
5. `WHERE (a, b) IN ((1, 2), (3, 4))` row constructors in `IN`
6. `SELECT * FROM STRING_SPLIT(@s, ',')` table-valued function in `FROM`
7. `SELECT * FROM (VALUES (0, 'Any'), (1, 'One')) AS t(id, name)` table value
   constructor in `FROM`

The following three are Oracle-specific, so they are listed separately:

8. `WHERE a.id (+) = b.id` legacy Oracle outer join operator
9. `LISTAGG(x, ', ') WITHIN GROUP (ORDER BY y)` ordered-set aggregate syntax
10. `FROM DUAL CONNECT BY LEVEL <= n`, `START WITH`, `PRIOR` hierarchical query
    clauses

New tests covering these forms will be added as part of the fix.

**Additionally:**

Three more forms were found while measuring against a production statement set.
The first is covered by this issue, the other two are listed for completeness:

11. `SELECT TOP (?) ...`, `SELECT TOP (:limit) ...`, `SELECT TOP (n + 1) ...`
    an expression as the row count of a parenthesized `TOP` (only an integer
    literal is accepted today, while `OFFSET`/`FETCH NEXT` already take
    expressions). T-SQL requires the parentheses exactly when `TOP` takes an
    expression rather than a constant, so the documented `TOP (@n)` form is the
    one being rejected. The bare form stays literal-only, since `TOP a` cannot
    be told apart from a select list starting with a column named `a`.

12. `SELECT * FROM {oj T RIGHT OUTER JOIN U ON T.id = U.id}` and
    `{? = call some_proc(?, ?)}` ODBC escape sequences - the lexer rejects `{`.

13. `SELECT NEXT VALUE FOR mydb.dbo.seq` T-SQL sequence expressions.

14. `SELECT a,\r\n b\r\nFROM t` statements with CRLF line endings - carriage    return is not in the lexer's whitespace class, so it falls through to the    catch-all rule, which ends the token stream. The statement is then cut off
    at the first line break: sometimes into a syntax error, sometimes into a
    shorter statement that still parses, which is the worse case since it is
    reported as valid.

sreckovic1987:
## Summary

Extends the vendored `SQLParser` (`Data/SQLParser`) grammar/lexer to support the
SQL Server, Oracle and ODBC forms listed in #5441 that previously failed to
parse, plus one standard-SQL literal form that parsed into the wrong statement.

## Changes

- **`flex_lexer.l`**
  - `WS [ \t\r\n]` definition, used by every rule that skips or spans
    whitespace. Carriage return was not matched by any rule but the catch-all
    one, which ends the token stream, so a statement with CRLF line endings was
    cut off at the first line break — sometimes into a syntax error, sometimes
    into a shorter statement that still parsed and was therefore reported as
    valid. A carriage return inside a string literal is unaffected, since string
    content is lexed in its own start condition.
  - `COMMENT` is an exclusive start condition. As an inclusive one it also ran
    the whitespace rule, which matches longer than the newline ending the
    comment, so the state never returned to `INITIAL` and the line after a
    comment was silently swallowed.
  - scientific notation numeric literals (`1.5e3`, `1.5E+3`, `.5e-3`, `2e10`) —
    there was no exponent rule at all
  - new single tokens so the spellings below cannot collide with ordinary
    identifiers: `(+)` (Oracle outer join marker), `{oj` (ODBC outer join escape
    prefix, with trailing context so it is not taken out of a name such as
    `{ojtest`), `NEXT VALUE FOR` (multi-word, like the existing `CURRENT ROW`)
  - multi-word tokens `START WITH`, `CONNECT BY`, `CONNECT BY NOCYCLE` and
    `WITHIN GROUP`, each with trailing context. The phrases are lexed as units
    so `START`, `CONNECT` and `WITHIN` remain usable as ordinary names.
  - new keyword `PRIOR` (see *Reserved words* below)

- **`bison_parser.y`**
  - `nonreserved_keyword`: date parts (`SECOND` … `YEAR` and plural forms),
    `ISNULL`, `FORMAT`, `CHAR`, `VARCHAR`, `INT`, `INTEGER`, `DATETIME`,
    `TIMESTAMP`, `NEXT` usable as column names, function names and function
    arguments — `DATEADD(MINUTE, ...)` used to depend on whether the date part
    was spelled out or abbreviated
  - `name_or_keyword`: the same keywords are accepted wherever the grammar
    names a database object — `table_name`, `ident_commalist`, `update_clause`,
    `column_def`, `table_alias`, `with_description`, `DROP INDEX` / `DROP
    COLUMN` and `opt_index_name` — as well as in qualified column references
    (`t.YEAR`, `dbo.t.MONTH`) and after `AS`
  - `alias`: `AS 'string literal'` and `AS <non-reserved keyword>`
  - `column_name`: three- and four-part references (`db.dbo.T.col`,
    `schema.seq.NEXTVAL`) and the matching qualified `*` forms — the column-side
    counterpart of the multi-part table names added in #5429
  - `in_expr`: row constructor on the left of `IN`, against both a subquery
    (`(a, b) IN (SELECT x, y FROM u)`) and a literal list
    (`(a, b) IN ((1, 2), (3, 4))`)
  - `nonjoin_table_ref_atomic`: table-valued function in `FROM`
    (`FROM STRING_SPLIT(s, ',')`) and table value constructor
    (`FROM (VALUES (1, 'a'), (2, 'b')) AS t(id, name)`)
  - `scalar_expr`: Oracle `(+)` outer join marker on a column reference
  - `function_expr` is defined in terms of `table_function_expr`, so
    `WITHIN GROUP (ORDER BY ...)` ordered-set aggregates and `OVER (...)` window
    clauses apply to every call form, including qualified and no-argument calls
  - `opt_hierarchical_clause`: Oracle `START WITH` / `CONNECT BY` in either
    order, with or without `NOCYCLE`, and `PRIOR` as a unary operator
  - `opt_top`: parenthesized `TOP` takes an expression (`TOP (?)`,
    `TOP (:limit)`), which is exactly the case T-SQL requires the parentheses
    for; the bare form stays literal-only since `TOP a` is ambiguous with the
    select list. A parenthesized `TOP` holding a placeholder cannot be combined
    with `LIMIT`/`OFFSET`: the placeholder is also referenced from the parameter
    list, so replacing the `LimitDescription` there would leave that reference
    dangling. `TOP` with a literal is unaffected.
  - ODBC escapes: `FROM {oj ...}` and `{? = call proc(...)}` /
    `{call proc(...)}`, the latter yielding the same `ExecuteStatement` an
    `EXECUTE` does. The return marker is a real placeholder, so ODBC's
    positional numbering is preserved and `{? = call p(?, ?)}` reports three
    parameters. `odbc_call_statement` and `execute_statement` share
    `opt_call_args`.
  - `next_value_expr`: T-SQL `NEXT VALUE FOR db.schema.seq`

- **Data model** — new `TableRefType` values `kTableFunc` / `kTableValues` with
  `TableRef::func` / `TableRef::values`, new `OperatorType` values
  `kOpOuterJoin` / `kOpPrior`, `Expr::withinGroupOrder`,
  `SelectStatement::startWith` / `connectBy`, `ExecuteStatement::returnValue`.
  All enum values appended at the end; all new members initialised in the
  constructors, released in the destructors and printed by `sqlhelper`.

- `bison_parser.cpp/.h`, `flex_lexer.cpp/.h` regenerated with bison 3.8.2 /
  flex 2.6.4. The `POCO-LOCAL` `io.h` conditional in the generated flex output
  was re-applied after each regeneration.

## Reserved words

`START`, `CONNECT` and `WITHIN` are **not** reserved. They are lexed only as
part of the phrases above, so `SELECT * FROM start`, `UPDATE t SET start = 1`
and `SELECT a AS within FROM t` all still parse.

`PRIOR` **is** reserved. It is a prefix operator with no neighbouring keyword to
key off, so it cannot be recognised contextually. This is a behaviour change:
`SELECT prior FROM t` and `DELETE FROM t WHERE prior = 1` parsed before this PR
and no longer do. Both Oracle and T-SQL reserve `PRIOR`.

One form is deliberately not accepted: a non-reserved keyword as an alias
written without `AS` (`SELECT a year`), which would be ambiguous with an
interval literal such as `SELECT 1 YEAR`. It does not parse on `main` either.

## Tests

Every grammar feature is covered in both the upstream suite and
`SQLParserTest`, and negative cases are in `queries-bad.sql`. Each test also
pins the pre-existing use of the same tokens (`EXTRACT(MINUTE FROM d)`, postfix
`ISNULL`, `CAST(a AS INT)`, `FETCH NEXT`, `INSERT ... VALUES`, `oj` / `value` /
`connect` / `start` as ordinary names, `a.b` references, `(a + b)` arithmetic).

On the Poco side `SQLParserTest` covers what the upstream suite cannot: that a
statement parses and classifies correctly, CRLF handling, and the placeholder
counts reported by `parameters()`.

## Changes after review

- Carriage return handled consistently, `%x COMMENT`, `{oj` delimiter.
- `START` / `CONNECT` / `WITHIN` de-reserved via multi-word tokens; `PRIOR` kept
  reserved and documented above.
- Non-reserved keywords accepted in qualified and aliased positions via
  `name_or_keyword`.
- Hierarchical clauses accept either order plus `NOCYCLE`.
- Use-after-free fixed: parenthesized `TOP` with a placeholder combined with
  `LIMIT`/`OFFSET` is rejected instead of freeing a still-referenced expression.
- ODBC return marker counted as a parameter, held on
  `ExecuteStatement::returnValue`.
- `function_expr` defined in terms of `table_function_expr`, removing four
  duplicated productions; `kTableFunc` mirrors `schema` as well as `name`;
  `execute_statement` and `odbc_call_statement` share `opt_call_args`.
- `qualifiedName` made static, and the six copies of the dotted-name join
  folded into one `dotJoin` helper.
- The fourteen dialect tests in `SQLParserTest` consolidated into one
  table-driven test plus the Poco-specific cases.

Full suites re-run: upstream 261/261 grammar tests, `SQLParserTest` 17/17,
bison 3.8.2 with 0 conflicts, flex 2.6.4 with no warnings.


aleks-f (approved):
see copilot comment, it sounds like it makes sense

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1377,13 +1531,45 @@ table_ref : table_ref_atomic | table_ref_commalist ',' table_ref_atomic {
...
   $$ = tbl;
+}
+// Table-valued function, e.g. FROM STRING_SPLIT(s, ','). name mirrors the
+// function name so getName() behaves like it does for a plain table.
+| function_expr opt_table_alias {
+  auto tbl = new TableRef(kTableFunc);
+  tbl->func = $1;
+  if ($1->name) tbl->name = strdup($1->name);
+  tbl->alias = $2;
+  $$ = tbl;
```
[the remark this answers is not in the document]
sreckovic1987: fixed by adding a dedicated `table_function_expr` that matches plain calls only, so `OVER` and `WITHIN GROUP` are now parse errors in `FROM` while staying valid in a select list; negative cases added to `queries-bad.sql` and both test suites.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1087,10 +1137,14 @@ opt_null_ordering : /* empty */ { $$ = NullOrdering::Undefined; }
...
-// TODO: TOP and LIMIT can take more than just int literals.
+// TODO: LIMIT can take more than just int literals.
 
+// T-SQL requires the parentheses exactly when TOP takes an expression rather
+// than a constant, e.g. TOP (@n) or the TOP (?) a parameterised statement ends
+// up with. The bare form stays literal-only: "TOP a" cannot be told apart from
+// a select list starting with a column named a.
 opt_top : TOP int_literal { $$ = new LimitDescription($2, nullptr); }
-| TOP '(' int_literal ')' { $$ = new LimitDescription($3, nullptr); }
+| TOP '(' expr ')' { $$ = new LimitDescription($3, nullptr); }
```
matejk, line 1190: Use-after-free. `expr` may be a placeholder (`param_expr` pushes it to `yyloc.param_list`). When a LIMIT/OFFSET follows, `select_no_paren` does `delete $$->limit; $$->limit = $3;`, which deletes this `LimitDescription` and the placeholder `Expr`, but `param_list` still holds the pointer; the `input` action then reads/writes it (`expr->type`, `expr->ival = ...`) and hands it out via `SQLParserResult::parameters()`.

valgrind on `SELECT TOP (?) * FROM t LIMIT 5;`: invalid read at bison_parser.y:405 / invalid write at :406, block freed by `LimitDescription::~LimitDescription` called from :1011. main rejects `TOP (?)`, so this is new here.

Suggested fix in `select_no_paren` (keeps the upstream behaviour for `select top 20 a from t1 limit 10;`, which `SetLimitOffset` asserts):

```c++
if ($3) {
  if ($$->limit && $$->limit->limit->type != kExprLiteralInt) {
    delete $1;
    if ($2) { for (auto ptr : *$2) delete ptr; delete $2; }
    delete $3;
    if ($4) { for (auto ptr : *$4) delete ptr; delete $4; }
    yyerror(&yyloc, result, scanner, "TOP with an expression cannot be combined with LIMIT.");
    YYERROR;
  }
  delete $$->limit;
  $$->limit = $3;
}
```

(`param_list` is only dereferenced in the `input` action, which runs on success only, so deleting here is safe.) Please add `!SELECT TOP (?) * FROM t LIMIT 5;` to queries-bad.sql.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -271,13 +283,14 @@
 %token RANGE ROWS GROUPS UNBOUNDED FOLLOWING PRECEDING CURRENT_ROW
 %token FETCH NEXT ONLY
 %token UNIQUE PRIMARY FOREIGN KEY REFERENCES
+%token OUTERJOIN WITHIN CONNECT PRIOR START ODBC_OJ NEXT_VALUE_FOR
```
matejk: Regression: `START`, `CONNECT`, `PRIOR` and `WITHIN` were identifiers on main and are reserved words now. Verified with a probe (VALID on main, INVALID on this branch): `SELECT * FROM start;`, `UPDATE t SET start = 1;`, `INSERT INTO t (start, connect) VALUES (1, 2);`, `CREATE TABLE t (start INT);`, `SELECT a AS within FROM t;`, `SELECT * FROM t AS connect;`, `SELECT start.a FROM t start;`, `SELECT prior FROM t;`, `DELETE FROM t WHERE prior = 1;`. `nonreserved_keyword` only rescues START/CONNECT as unqualified column or function names. In Poco this also changes behaviour: `Statement::checkBeginTransaction()` calls `parse()` and only begins a transaction when the parse succeeds, so `UPDATE t SET start = ?` on a non-autocommit session no longer starts one.

Suggestion (0 bison conflicts, checked): do not reserve the single words; lex the phrases as multi-word tokens like CURRENT ROW / NEXT VALUE FOR:

```
START{WS}+WITH/[^A-Za-z0-9_]             TOKEN(START_WITH)
CONNECT{WS}+BY/[^A-Za-z0-9_]             TOKEN(CONNECT_BY)
CONNECT{WS}+BY{WS}+NOCYCLE/[^A-Za-z0-9_] TOKEN(CONNECT_BY_NOCYCLE)
WITHIN{WS}+GROUP/[^A-Za-z0-9_]           TOKEN(WITHIN_GROUP)
```

with `WS [ \t\r\n]` defined in section 1. The trailing context matters: without it `CONNECT BY nocycleflag = 1` lexes as CONNECT_BY_NOCYCLE + `flag`, and `FROM start withinx` fails. Then `opt_within_group : WITHIN_GROUP '(' ORDER BY order_list ')'`, and START/CONNECT can be dropped from `nonreserved_keyword`. `PRIOR` is a unary operator and cannot be lexed contextually; keeping it reserved is defensible (Oracle and T-SQL reserve it), but please say so in the PR description since `SELECT prior FROM t` parsed before.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -479,6 +497,22 @@ prepare_statement : PREPARE IDENTIFIER FROM prepare_target_query {
...
+  $$ = new ExecuteStatement();
+  $$->name = qualifiedName($4);
+  $$->parameters = $6;
+}
+| '{' opt_odbc_return CALL table_name '}' {
+  $$ = new ExecuteStatement();
+  $$->name = qualifiedName($4);
+};
+
+opt_odbc_return : '?' '=' | /* empty */;
```
matejk: The return marker is consumed as a bare `'?'` token, so it never reaches `yyloc.param_list` and `SQLParserResult::parameters()` omits it: `{? = call p(?, ?)}` reports 2 parameters (columns 12 and 15) although ODBC numbers the marker as parameter 1 and Poco binds it positionally (`session << "{? = call storedFunction(?)}", out(result), in(i)` in the ODBC test suites). `Utility::boundSQL` builds its arity check from `parameters()`, so it now sees one placeholder for two bound positions.

```suggestion
opt_odbc_return : param_expr '=' { $$ = $1; } | /* empty */ { $$ = nullptr; };
```

plus `%type <expr> opt_odbc_return`, a new `Expr* returnValue` on `ExecuteStatement` (initialised to nullptr, deleted in the destructor) set from `$2` in both `odbc_call_statement` alternatives. With that, `{? = call p(?, ?)}` yields 3 parameters with ival 0,1,2 and ival2 1,15,18 (checked).

Data/SQLParser/src/parser/flex_lexer.l
```diff
@@ -61,7 +61,7 @@ static thread_local std::stringstream strbuf;
 <COMMENT>[^\n]* /* skipping comment content until a end of line is read */;
 <COMMENT>\n     BEGIN(INITIAL);
 
-[ \t\n]+        /* skip whitespace */;
+[ \t\r\n]+      /* skip whitespace, CR included: a CRLF file would otherwise hit the catch-all rule below, which ends the token stream */;
```
matejk: Related, pre-existing, and worth fixing while this line is touched: `COMMENT` is declared with `%s`, i.e. inclusive, so this rule is also active inside a `--` comment. When the newline that ends the comment is followed by more whitespace (indented next line, blank line, trailing spaces), this rule matches longer than `<COMMENT>\n`, the state never returns to INITIAL, and the next line is eaten by `<COMMENT>[^\n]*`.

`SELECT a FROM t\n-- c\n   WHERE b = 1;` and `SELECT a FROM t -- c\n\nWHERE b = 1;` parse as VALID without the WHERE clause, on this branch and on main - the same silent-truncation class as the CR problem this line fixes.

Fix: `%x COMMENT` (exclusive) at line 52; the two `<COMMENT>` rules cover all input and EOF falls through to yyterminate. Checked with flex 2.6.4: no warnings, `flex -s` clean, all cases above keep the WHERE.

Data/SQLParser/src/parser/flex_lexer.l
```diff
@@ -228,10 +231,12 @@ VIRTUAL     TOKEN(VIRTUAL)
 WHEN        TOKEN(WHEN)
 WHERE       TOKEN(WHERE)
 WITH        TOKEN(WITH)
+WITHIN      TOKEN(WITHIN)
 YEAR        TOKEN(YEAR)
 YEARS       TOKEN(YEARS)
 
 CURRENT[ \t\n]+ROW       TOKEN(CURRENT_ROW)
+NEXT[ \t\n]+VALUE[ \t\n]+FOR TOKEN(NEXT_VALUE_FOR)
 CHARACTER[ \t\n]+VARYING TOKEN(CHARACTER_VARYING)
```
matejk: `\r` is whitespace now, but the multi-word tokens still use `[ \t\n]`, so `SELECT NEXT VALUE\r\nFOR seq;` fails (NEXT lexed as keyword, VALUE as identifier); same for `CHARACTER\r\nVARYING`, `CURRENT\r\nROW` and `<afterarray>[ \t\n]+` at line 320. The new `{oj` / `(+)` patterns already use `[ \t\r\n]`. A `WS [ \t\r\n]` definition in section 1 used by all of them keeps this in one place; minimal version:

```suggestion
CURRENT[ \t\r\n]+ROW       TOKEN(CURRENT_ROW)
NEXT[ \t\r\n]+VALUE[ \t\r\n]+FOR TOKEN(NEXT_VALUE_FOR)
CHARACTER[ \t\r\n]+VARYING TOKEN(CHARACTER_VARYING)
```

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1176,7 +1236,24 @@ logic_expr : expr AND expr { $$ = Expr::makeOpBinary($1, kOpAnd, $3); }
 in_expr : operand IN '(' expr_list ')' { $$ = Expr::makeInOperator($1, $4); }
 | operand NOT IN '(' expr_list ')' { $$ = Expr::makeOpUnary(kOpNot, Expr::makeInOperator($1, $5)); }
 | operand IN '(' select_no_paren ')' { $$ = Expr::makeInOperator($1, $4); }
-| operand NOT IN '(' select_no_paren ')' { $$ = Expr::makeOpUnary(kOpNot, Expr::makeInOperator($1, $5)); };
+| operand NOT IN '(' select_no_paren ')' { $$ = Expr::makeOpUnary(kOpNot, Expr::makeInOperator($1, $5)); }
+// Row constructor on the left of IN, e.g. (a, b) IN (SELECT x, y FROM t). The
+// tuple is represented as an array expression - Expr has no dedicated row type
+// and this keeps every element in the tree instead of dropping any.
+| '(' row_expr_list ')' IN '(' select_no_paren ')' { $$ = Expr::makeInOperator(Expr::makeArray($2), $6); }
```
matejk, line 1286: Item 5 of #5441 is the literal-list form `(a, b) IN ((1, 2), (3, 4))`; that still fails (`unexpected '(', expecting SELECT`) - only the subquery form is added here. This is conflict-free (checked) and reuses the row shape:

```
in_row_list : '(' row_expr_list ')' {
  $$ = new std::vector<Expr*>();
  $$->push_back(Expr::makeArray($2));
}
| in_row_list ',' '(' row_expr_list ')' {
  $1->push_back(Expr::makeArray($4));
  $$ = $1;
};

| '(' row_expr_list ')' IN '(' in_row_list ')' { $$ = Expr::makeInOperator(Expr::makeArray($2), $6); }
| '(' row_expr_list ')' NOT IN '(' in_row_list ')' { $$ = Expr::makeOpUnary(kOpNot, Expr::makeInOperator(Expr::makeArray($2), $7)); }
```

(`%type <expr_vec> in_row_list`; rows have 2+ elements, so there is no clash with a parenthesised operand.) Also, `row_expr_list` is built from `expr_alias`, so `(a AS q, b) IN (...)` is accepted; `expr ',' expr | row_expr_list ',' expr` is conflict-free as well.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1273,7 +1382,69 @@ between_expr : operand BETWEEN operand AND operand { $$ = Expr::makeBetween($1,
 column_name : IDENTIFIER { $$ = Expr::makeColumnRef($1); }
 | IDENTIFIER '.' IDENTIFIER { $$ = Expr::makeColumnRef($1, $3); }
 | '*' { $$ = Expr::makeStar(); }
-| IDENTIFIER '.' '*' { $$ = Expr::makeStar($1); };
+| IDENTIFIER '.' '*' { $$ = Expr::makeStar($1); }
+| nonreserved_keyword { $$ = Expr::makeColumnRef($1); }
```
matejk, line 1441: Copilot's suppressed comment is right: the keywords only work unqualified. `SELECT t.YEAR FROM t;` -> `unexpected YEAR, expecting IDENTIFIER or '*'`; `dbo.t.MONTH`, `SELECT a AS year FROM t;`, `SELECT * FROM t AS next;` also fail.

Checked conflict-free with bison: add explicit alternatives `IDENTIFIER '.' nonreserved_keyword`, `IDENTIFIER '.' IDENTIFIER '.' nonreserved_keyword` (and 4-part) here, `alias : AS nonreserved_keyword`, and a `name_or_keyword : IDENTIFIER | nonreserved_keyword` used in `table_name`, `ident_commalist`, `update_clause`, `column_def`, `table_alias : AS name_or_keyword '(' ident_commalist ')'`, `with_description`, DROP INDEX/COLUMN and `opt_index_name`. Do not use `name_or_keyword` inside `column_name` itself (it forces early reduction against `IDENTIFIER '('`), and do not add a bare `alias : nonreserved_keyword` (19 conflicts with interval literals such as `SELECT 1 YEAR`).

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1202,12 +1279,35 @@ comp_expr : operand '=' operand { $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
 // `function_expr is used for window functions, aggregate expressions, and functions calls because we run into shift/
 // reduce conflicts when splitting them.
 function_expr : IDENTIFIER '(' ')' opt_window { $$ = Expr::makeFunctionRef($1, new std::vector<Expr*>(), false, $4); }
-| IDENTIFIER '(' opt_distinct expr_list ')' opt_window { $$ = Expr::makeFunctionRef($1, $4, $3, $6); }
+| IDENTIFIER '(' opt_distinct expr_list ')' opt_within_group opt_window {
```
matejk: `opt_within_group` is only on this alternative; `s.LISTAGG(a, ',') WITHIN GROUP (ORDER BY b)`, `f() WITHIN GROUP (...)` and `FORMAT(a) WITHIN GROUP (...)` are rejected. If `function_expr` is written as `table_function_expr opt_within_group opt_window { $$ = $1; $$->withinGroupOrder = $2; $$->windowDescription = $3; }` (and the `nonreserved_keyword` call forms are moved into `table_function_expr`), every call form gets WITHIN GROUP and OVER, and the four productions duplicated between `function_expr` and `table_function_expr` go away.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1015,14 +1049,24 @@ set_type : UNION {
 opt_all : ALL { $$ = true; }
 | /* empty */ { $$ = false; };
 
-select_clause : SELECT opt_top opt_distinct select_list opt_from_clause opt_where opt_group {
+// Oracle hierarchical query clauses. Written in the canonical order, START
+// WITH before CONNECT BY; either may be omitted.
+opt_start_with : START WITH expr { $$ = $3; }
```
matejk: Oracle also allows `CONNECT BY [NOCYCLE] cond START WITH cond` (either order): `SELECT * FROM t CONNECT BY PRIOR id = parent_id START WITH id = 1;` and `... CONNECT BY NOCYCLE PRIOR id = parent_id;` fail here. A single `opt_hierarchical_clause` returning both expressions is conflict-free:

```
opt_hierarchical_clause : START_WITH expr connect_by expr
| connect_by expr START_WITH expr
| START_WITH expr
| connect_by expr
| /* empty */ ;
connect_by : CONNECT_BY | CONNECT_BY_NOCYCLE;
```

(e.g. as `std::pair<Expr*, Expr*>*` in `%union` with a `%destructor`, unpacked in `select_clause`). Ties in with the multi-word token suggestion at line 286.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1377,13 +1548,45 @@ table_ref : table_ref_atomic | table_ref_commalist ',' table_ref_atomic {
...
   tbl->select = $2;
   tbl->alias = $4;
   $$ = tbl;
+}
+// Table-valued function, e.g. FROM STRING_SPLIT(s, ','). name mirrors the
+// function name so getName() behaves like it does for a plain table.
+| table_function_expr opt_table_alias {
+  auto tbl = new TableRef(kTableFunc);
+  tbl->func = $1;
+  if ($1->name) tbl->name = strdup($1->name);
```
matejk: `name` is mirrored so `getName()` works, but `schema` is not, so `FROM dbo.fn_split('a', ',') f` gives `hasSchema() == false` while `func->schema == "dbo"`. The guard is always true for a function ref.

```suggestion
  tbl->name = strdup($1->name);
  if ($1->schema) tbl->schema = strdup($1->schema);
```

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -27,6 +28,17 @@
     result->setErrorDetails(strdup(msg), llocp->first_line, llocp->first_column);
     return 0;
   }
+
+  // Flattens a possibly qualified name back into one string, taking ownership
+  // of both parts. Used where the statement has a single name field, e.g. the
+  // procedure of an ODBC call escape.
+  char* qualifiedName(TableName name) {
```
matejk, line 46: Nit: this is a non-static function in the global namespace inside the prologue, so it becomes an exported symbol (`nm -DC libsqlparser.so` shows `T qualifiedName(hsql::TableName)`). Wrapping it in an anonymous namespace (or `static`) avoids that. The same `std::string(a) + "." + b; free; strdup` join is also written out in the four multi-part `column_name` actions (1392-1417) and in the 3-part `table_name` action (1627); a small `static char* dotJoin(char* a, char* b)` used by all six would remove the duplication.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1202,12 +1279,35 @@ comp_expr : operand '=' operand { $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
...
+}
+// Dialect functions whose names collide with grammar keywords, e.g.
+// ISNULL(a, 0), CHAR(10), FORMAT(x, '00'). See nonreserved_keyword.
+| nonreserved_keyword '(' ')' opt_window { $$ = Expr::makeFunctionRef($1, new std::vector<Expr*>(), false, $4); }
+| nonreserved_keyword '(' opt_distinct expr_list ')' opt_window { $$ = Expr::makeFunctionRef($1, $4, $3, $6); };
+
+// A plain function call, without the OVER and WITHIN GROUP clauses
+// function_expr allows. Only this form can stand as a table reference: a
+// window function or an ordered-set aggregate is not a table.
+table_function_expr : IDENTIFIER '(' ')' {
```
matejk, line 1338: Nit: these four productions are a verbatim copy of the IDENTIFIER alternatives of `function_expr` minus the trailing clauses; see the comment at line 1282 for defining `function_expr` in terms of `table_function_expr` instead. Similarly, `odbc_call_statement` repeats `execute_statement`'s with-args / without-args pair - both build an `ExecuteStatement` and could share an `opt_call_args` nonterminal.

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1146,14 +1200,20 @@ expr_alias : expr opt_alias {
...
-scalar_expr : column_name | literal;
+scalar_expr : column_name | literal
+// Oracle outer join marker, e.g. WHERE a.id (+) = b.id. Only a column
+// reference can carry it, which is where Oracle allows it; the marked side
+// stays visible in the tree as a unary operator.
+| column_name OUTERJOIN { $$ = Expr::makeOpUnary(kOpOuterJoin, $1); };
 
 unary_expr : '-' operand { $$ = Expr::makeOpUnary(kOpUnaryMinus, $2); }
+// Oracle's PRIOR, referring to the parent row inside CONNECT BY.
+| PRIOR operand %prec PRIOR { $$ = Expr::makeOpUnary(kOpPrior, $2); }
```
matejk: Nit: `%prec PRIOR` is a no-op here (PRIOR is the rule's only terminal, so bison assigns that precedence anyway; the regenerated tables are identical without it).

```suggestion
| PRIOR operand { $$ = Expr::makeOpUnary(kOpPrior, $2); }
```

Data/SQLParser/src/parser/bison_parser.y
```diff
@@ -1087,10 +1137,14 @@ opt_null_ordering : /* empty */ { $$ = NullOrdering::Undefined; }
   $$ = null_ordering;
 };
 
-// TODO: TOP and LIMIT can take more than just int literals.
+// TODO: LIMIT can take more than just int literals.
```
matejk: Nit: `opt_limit` already takes `expr` (line 1150), so the edited TODO is now wrong rather than stale.

```suggestion
```

Data/SQLParser/src/parser/flex_lexer.l
```diff
@@ -254,6 +272,19 @@ CHARACTER[ \t\n]+VARYING TOKEN(CHARACTER_VARYING)
 
 [-+*/(){},.;<>=^%:?[\]|]    { return yytext[0]; }
```
matejk, line 295: Follow-up notes, not blocking: `@` is not in this set nor in any rule, so T-SQL variables hit the catch-all rule (which prints `[SQL-Lexer-Error] Unknown Character` to stderr and ends the token stream). That means the example in the `opt_top` comment, `TOP (@n)`, and item 6 of #5441, `STRING_SPLIT(@s, ',')`, do not parse. Other gaps found while probing: 3-part function names `mydb.dbo.fn(1)`, AS-less alias column lists `(VALUES (1, 2)) v(a, b)`, `NEXT VALUE FOR seq OVER (ORDER BY x)`, and `FROM t (NOLOCK)` now parsing as a table-valued function `t(NOLOCK)`. Also, the new `[oO][jJ]` / `[eE]` classes are redundant under `%option case-insensitive`, and `{oj` needs no delimiter (`{ojtest ...` is accepted).

Data/testsuite/src/SQLParserTest.cpp
```diff
@@ -192,6 +193,642 @@ void SQLParserTest::testArrayLiteralNotShadowedByBracketIdentifier()
 }
 
 
+void SQLParserTest::testNonReservedKeywords()
```
matejk: Two remarks on the test additions: (1) the 14 new CppUnit tests repeat the upstream select_tests.cpp / sql_tests.cpp tests almost statement-by-statement (and the same statements are in queries-good.sql); the upstream suite already runs in CI (`linux-sqlparser-upstream-tests`), so a compact table-driven "parses and classifies as SELECT" test plus the Poco-specific cases (CRLF, `parameters()`) would give the same coverage with a fraction of the maintenance cost. (2) Several new comments describe the pre-fix behaviour ("used to parse as ...", "before this the behaviour depended on ...", "counterpart of ... added in #5429"); comments should state the current rule only - the history is in git and in the PR.

matejk (changes requested):
Thanks - the AST additions, destructors, `%destructor` coverage and the regenerated bison/flex files all check out (a fresh bison 3.8.2 / flex 2.6.4 regeneration differs only in `#line` directives and the POCO-LOCAL `io.h` block; 0 grammar conflicts; upstream suite and valgrind clean; `Data-testrunner SQLParserTest` 28/28). I ran the PR grammar next to main with a probe binary and found the items below; the inline comments carry the details and suggested fixes.

**Must fix**
1. Use-after-free (valgrind): `SELECT TOP (?) * FROM t LIMIT 5;` - the TOP `LimitDescription` is deleted by `select_no_paren` while its placeholder `Expr` is still in `yyloc.param_list`; the `input` rule then reads/writes freed memory and `parameters()` returns a dangling pointer. New in this PR (TOP was literal-only before).
2. Regression: `START`, `CONNECT`, `PRIOR`, `WITHIN` are now reserved words. On main they were identifiers, so `SELECT * FROM start`, `UPDATE t SET start = 1`, `INSERT INTO t (start, connect) VALUES (1, 2)`, `CREATE TABLE t (start INT)`, `SELECT a AS within`, `SELECT prior FROM t` etc. parsed and now fail. In Poco this also means `Statement::checkBeginTransaction()` no longer starts a transaction for such non-SELECT statements. Suggest multi-word tokens (START WITH / CONNECT BY / WITHIN GROUP), see inline.
3. `{? = call p(?, ?)}`: the return marker is a bare `'?'` token, so `parameters()` reports 2 placeholders instead of 3 (`Utility::boundSQL` uses that count).
4. `\r` is now whitespace, but the multi-word tokens (new NEXT VALUE FOR, existing CURRENT ROW / CHARACTER VARYING / afterarray) still use `[ \t\n]`.
5. Pre-existing but on the edited whitespace line: `--` comments use an inclusive start condition, so a comment followed by an indented or blank line silently swallows the next line (`SELECT a FROM t\n-- c\n   WHERE b = 1;` parses without the WHERE). Same failure class as the CR fix; `%x COMMENT` fixes it.
6. Issue #5441 item 5 (`(a, b) IN ((1, 2), (3, 4))`) still fails - only the subquery form was added.
7. Copilot's suppressed comment is correct: `t.YEAR`, `dbo.t.MONTH`, `a AS year`, `FROM t AS next` still fail.

**Should fix**
8. WITHIN GROUP is only accepted on `IDENTIFIER '(' expr_list ')'`.
9. CONNECT BY before START WITH and NOCYCLE are rejected (Oracle allows both).
10. kTableFunc mirrors `func->name` but not `func->schema` into the TableRef.

**Cleanup / nits:** exported `qualifiedName` symbol and the dotted-name join duplicated six times, `table_function_expr` duplicating `function_expr`, `row_expr_list` built from `expr_alias`, `%prec PRIOR` no-op, redundant `[oO][jJ]` / `[eE]` under case-insensitive, wrong LIMIT TODO, comments that narrate the pre-fix behaviour, and the CppUnit tests copying the upstream tests 1:1 (upstream already runs in CI).

**Follow-ups** (not blocking, maybe worth a note in #5441): `@var` is not lexed (so the PR's own `TOP (@n)` and the issue's `STRING_SPLIT(@s, ',')` do not parse), 3-part function names `db.dbo.fn(...)`, AS-less `(VALUES ...) v(a, b)`, `NEXT VALUE FOR seq OVER (...)`, and `FROM t (NOLOCK)` now parsing as a table-valued function.

Every grammar change suggested inline was checked with bison 3.8.2 on a copy of this branch: 0 conflicts, upstream suite green.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Fixed in `select_no_paren` as suggested, and `!SELECT TOP (?) * FROM t LIMIT 5;` added to `queries-bad.sql`.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Done as suggested. `START`, `CONNECT` and `WITHIN` are now the four multi-word tokens with the trailing context you proposed; WS [ \t\r\n] already went in with the CR fix, so the phrases reuse it. `START/CONNECT` are dropped from `nonreserved_keyword` and `opt_within_group` takes `WITHIN_GROUP`.

It is checked the cases the trailing context is there for: CONNECT BY nocycleflag = 1, FROM start withinx, and names like connectionid / startdate / withinrange all lex as identifiers.

`PRIOR` stays reserved as you suggested, and it is updated the PR description to say so.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Done exactly as suggested: `opt_odbc_return` now goes through `param_expr`, with `%type <expr>` and a new `Expr* returnValue` on `ExecuteStatement`, nulled in the constructor and deleted in the destructor.

The with-args / without-args pair it shared with `execute_statement` has also been folded into a common `opt_call_args`, per your other note, so `odbc_call_statement` is now a single alternative and `returnValue` is set there.

This was verified through `Utility::boundSQL`, since `parameters()` isn't exposed on the `Statement` facade: `{? = call p(?, ?)}` now takes 3 bindings and rejects 2, and `{call p(?, ?)}` takes 2 and rejects 3.

Data/SQLParser/src/parser/flex_lexer.l
sreckovic1987: Fixed as suggested: `COMMENT` is now `%x`. Both of the cases quoted above keep the `WHERE` clause, and CRLF variants of them were added to the tests as well, since the corpus this is aimed at is entirely CRLF.

The diagnosis is confirmed by the regenerated output: the tables shrink by about a third, which is the keyword rule set no longer being duplicated into the comment state. `flex` 2.6.4 regenerates with no warnings.

Data/SQLParser/src/parser/flex_lexer.l
sreckovic1987: Fixed, using the `WS [ \t\r\n]` definition rather than the minimal version, since that keeps it in one place as you suggested. It is now used by every rule that skips or spans whitespace: the general whitespace rule, all three multi-word tokens, the `afterarray` state, the `{oj` and `(+)` patterns, and the four phrase tokens added for your other comment. Eleven sites, one definition.

Thanks for catching `<afterarray>`. It is an exclusive start condition, so the general whitespace rule never covered it and it would have been missed.

All four forms listed above were checked in their CRLF variants. Three of them are pinned in `testCarriageReturn`: `NEXT VALUE\r\nFOR`, `CURRENT\r\nROW` and `ARRAY\r\n[foo]`. `CHARACTER\r\nVARYING` parses as well but is not in the test.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Both parts are in. `in_row_list` was added as suggested, so item 5 of #5441 now parses in the literal-list form as well as the subquery one, and `row_expr_list` is built from `expr` rather than `expr_alias`, so `(a AS q, b) IN (...)` is no longer accepted. Regenerated with 0 conflicts, as you said.
The two literal-list forms are pinned in `SQLParserTest`.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Removed. `opt_limit` already takes an `expr`, so editing the TODO had turned a stale comment into a wrong one.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Removed. `PRIOR` is the rule's only terminal, so the precedence is assigned anyway.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Both done. `qualifiedName` and the new `dotJoin` are inside an anonymous namespace now, so nothing is exported.

All six sites you listed go through `dotJoin`, plus three more that came out of the qualified keyword forms added for your other comment, so there are nine call sites and no hand-written joins left.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Both done. `function_expr` is now `table_function_expr opt_within_group opt_window`, with the `nonreserved_keyword` call forms moved into `table_function_expr` as well, so the four duplicated productions are gone and every call form gets `OVER` and `WITHIN GROUP`. That covers your other comment on `opt_within_group` at the same time.

`opt_call_args` is shared by `execute_statement` and `odbc_call_statement`, which went in with the return marker fix.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Done, including both warnings: `name_or_keyword` is not used inside `column_name`, and there is no bare `alias : nonreserved_keyword`. The interval-literal clash was confirmed before trying it, since `SELECT 1 YEAR` is `interval_literal : INTVAL duration_field`.

One deviation in `table_name`, since your version was checked conflict-free and this one is written differently. Putting `name_or_keyword` on a leading part adds `'.'` to its follow set, which collides with `table_function_expr : IDENTIFIER '.' IDENTIFIER '('` and gives one shift/reduce on `'.'`. Restructuring `function_expr` as you suggested in the other comment does not clear it. The leading parts are therefore spelled out: `IDENTIFIER`-leading alternatives keep `IDENTIFIER`, which is a terminal and needs no reduce decision, and keyword-leading ones are separate alternatives. Coverage is the same as your version, `FROM year.t` and `FROM year.dbo.t` included, and it regenerates with 0 conflicts.

The one form still not accepted is a keyword as an alias written without `AS`, for the reason you gave.

Data/testsuite/src/SQLParserTest.cpp
sreckovic1987: Both done. The 14 dialect tests are now one table-driven "parses and classifies" test, plus the two Poco-specific cases the upstream suite cannot cover: CRLF input and the placeholder counts from `parameters()`. `SQLParserTest` is 17 tests instead of 28.

The comments that narrated pre-fix behaviour are rewritten to state the current rule only, in the grammar, the lexer and `select_tests.cpp`.

Data/SQLParser/src/parser/flex_lexer.l
sreckovic1987: The three in-scope points are applied: `{oj` now requires a delimiter via trailing context, the redundant `[oO][jJ]` / `[eE]` classes are gone under `%option case-insensitive`, and the `TOP (@n)` example in the `opt_top` comment is replaced with `TOP (?)` / `TOP (:limit)`, since `@` does not lex.

One extra data point on `@`, since it is more than a parse failure: the catch-all rule writes to stderr, so a statement containing a T-SQL variable produces lexer error output on every parse attempt, not just a failed one.

The rest, `@` variables, 3-part function names, AS-less `(VALUES ...) v(a, b)`, `NEXT VALUE FOR seq OVER (...)` and `FROM t (NOLOCK)` parsing as a table-valued function, are out of scope for this PR and will be filed separately.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Applied as suggested: `schema` is mirrored alongside `name`, and the guard on `name` is gone since `makeFunctionRef` always assigns it. `FROM dbo.fn_split('a', ',') f` now reports `hasSchema() == true`.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Done exactly as written: `function_expr` is now `table_function_expr opt_within_group opt_window`, with the `nonreserved_keyword` call forms moved into `table_function_expr`. All three of your examples parse now, `s.LISTAGG(a, ',') WITHIN GROUP (ORDER BY b)`, `f() WITHIN GROUP (...)` and `FORMAT(a) WITHIN GROUP (...)`, and `OVER` applies to every call form the same way.

The four duplicated productions are gone with it, which also covers your other comment on `table_function_expr`.

Data/SQLParser/src/parser/bison_parser.y
sreckovic1987: Done as suggested: `opt_hierarchical_clause` with the five alternatives and `connect_by : CONNECT_BY | CONNECT_BY_NOCYCLE`, unpacked in `select_clause`. Both of your examples parse now, and so does `CONNECT BY NOCYCLE ... START WITH ...` in that order.

One small deviation: the carrier is a two-pointer POD struct rather than `std::pair`, since it reads better in the actions. Same `%union` and `%destructor` story. It builds on the multi-word tokens from your other comment, as you expected.

Data/testsuite/src/SQLParserTest.cpp
```diff
@@ -192,6 +193,162 @@ void SQLParserTest::testArrayLiteralNotShadowedByBracketIdentifier()
...
+	}
+
+	// A line comment ends at its newline. If the following line were swallowed,
+	// the FROM clause would be lost and the statement would not parse.
+	static const char* const commented[] =
+	{
+		"SELECT A FROM\n-- comment\n  Test;",
+		"SELECT A FROM -- comment\n\nTest;",
+		"SELECT A FROM\r\n-- comment\r\n  Test;",
+		"SELECT A FROM Test -- comment   \n   WHERE B IN (1, 2);"
```
matejk: Nit: unlike the three cases above, this one still parses if the second line is swallowed (`SELECT A FROM Test` alone is valid), so the assertion below does not catch the comment bug. Either drop the table name to the second line as in the other cases, or assert `whereClause != nullptr` for this one.

Data/SQLParser/src/parser/flex_lexer.l
```diff
@@ -254,6 +294,18 @@ CHARACTER[ \t\n]+VARYING TOKEN(CHARACTER_VARYING)
...
 
+  /*
+   * Scientific notation, e.g. 1.5e3, 1.5E+3, .5e-3, 2e10. Has to come before the
```
matejk: Nit: the ordering claim is not what makes this work - flex picks the longest match, so `1.5e3` goes to this rule wherever it is placed relative to the float/integer rules (only equal-length matches are decided by order). Suggest: "Matched as one token; the plain float and integer rules would otherwise take the mantissa and leave the exponent to be scanned as an identifier."

Data/SQLParser/test/queries/queries-good.sql
```diff
@@ -135,3 +135,54 @@ SELECT * FROM [test];
...
+# String literal used as an alias
+SELECT a AS 'Coil Id', b AS 'X' FROM test;
+SELECT * FROM test AS 'my table';
+# Non-reserved keywords used as names
+SELECT * FROM test WHERE created >= DATEADD(MINUTE, -10, GETDATE());
+SELECT DATEDIFF(SECOND, a, b), DATEADD(year, -1, GETDATE()) FROM test;
+SELECT ISNULL(a, 0), CHAR(10), FORMAT(a, '00') FROM test;
+SELECT CONVERT(INT, a), CONVERT(DATETIME, b) FROM test;
+SELECT YEAR, MONTH FROM test ORDER BY YEAR;
+# Scientific notation numeric literals
```
matejk, line 185: Nit: the forms added in the follow-up commits (`SELECT * FROM start`, `UPDATE t SET start = 1`, `(a, b) IN ((1, 2), (3, 4))`, `CONNECT BY ... START WITH ...`, `CONNECT BY NOCYCLE ...`, `t.YEAR` / `a AS year`, `s.f(...) WITHIN GROUP (...)`) are only pinned in the CppUnit test now. Adding them here as well puts them under the upstream valgrind run in CI, which the CppUnit test does not get.

matejk (approved):
Re-reviewed the five new commits (up to b83bb48b3) against the previous review, with the same setup as before: fresh bison 3.8.2 / flex 2.6.4 regeneration matches the committed files (only `#line` and the POCO-LOCAL block differ), 0 conflicts, no flex warnings, upstream suite + valgrind clean, `Data-testrunner SQLParserTest` 17/17, and a PR-vs-main probe over the full regression list.

All points from the previous review are addressed:
- `SELECT TOP (?) * FROM t LIMIT 5;` / `TOP (:n) ... OFFSET 3` are rejected; valgrind clean on those and on `TOP (?) ... UNION ... LIMIT`; `select top 20 a from t1 limit 10` still parses.
- `START`, `CONNECT`, `WITHIN` are identifiers again: `SELECT * FROM start`, `UPDATE t SET start = 1`, `INSERT INTO t (start, connect) ...`, `CREATE TABLE t (start INT)`, `SELECT a AS within`, `SELECT a start FROM t connect` all parse; both hierarchical clause orders and `NOCYCLE` parse; the trailing context holds (`CONNECT BY nocycleflag = 1`, `FROM start withinx`). `PRIOR` stays reserved as agreed.
- `{? = call p(?, ?)}` reports 3 parameters (ival2 1, 15, 18); `{call p(?)}` 1.
- CRLF inside `NEXT VALUE FOR`, `CURRENT ROW`, `CHARACTER VARYING`, `ARRAY [`, `START WITH`, `CONNECT BY`, `WITHIN GROUP`, `(+)` and `{oj` all lex; `%x COMMENT` keeps the WHERE in every comment case, LF and CRLF.
- `(a, b) IN ((1, 2), (3, 4))` and the NOT IN form parse; `(a AS q, b) IN (...)` and `((1), (2))` are rejected.
- `t.YEAR`, `dbo.t.MONTH`, `a AS year`, `FROM t AS next`, `UPDATE t SET year = 1`, `INSERT INTO t (year, month)`, `CREATE TABLE t (year INT)`, `WITH year AS (...)` parse; `SELECT 1 YEAR` is still an interval.
- WITHIN GROUP / OVER work on every call form; `FROM COUNT(*) OVER (...)` and `FROM LISTAGG(...) WITHIN GROUP (...)` are still rejected; `FROM dbo.fn(...)` has `hasSchema() == true`; `{ojtest` is rejected, `{OJ ...}` accepted.
- `qualifiedName` / `dotJoin` are no longer exported; the duplicated productions, `%prec`, the bracket classes and the LIMIT TODO are gone; the trimmed CppUnit tests read well.

Three small nits inline, none blocking. Approving.

Data/testsuite/src/SQLParserTest.cpp
sreckovic1987: Both, in the end. The fourth case now carries the `FROM` target on the second line like the other three, and the `WHERE` form is kept as a separate check asserting `whereClause != nullptr`, since `isValid()` cannot see the difference there.

The assertion was confirmed to earn its place: against the pre-fix lexer that statement parses as valid with `whereClause == nullptr`.

Data/SQLParser/src/parser/flex_lexer.l
sreckovic1987: Taken verbatim. The ordering claim is gone: the comment now says the literal is matched as one token, and that the plain float and integer rules would otherwise take the mantissa and leave the exponent to be scanned as an identifier.

Data/SQLParser/test/queries/queries-good.sql
sreckovic1987: Added, 21 statements in six groups: non-reserved words as ordinary names (including `UPDATE test SET start = 1` and `CREATE TABLE test (start INT)`, which were not pinned anywhere), the literal-list `IN`, both hierarchical clause orders with and without `NOCYCLE`, `WITHIN GROUP` and `OVER` on qualified and keyword call forms, the table-valued function with a schema, and the ODBC call escape with a return marker.

`CONNECT BY nocycleflag = 1` went in as well, since that is the case the trailing context exists for. The upstream suite goes from 261 to 282 grammar tests.

PR #5453, opened 2026-08-19
chore: `constexpr`ify as many `static` constants as possible

mikomikotaishi:
This PR aims to make as many class constants `constexpr` as possible; primarily targeting primitives. In order to do so, I have moved some of the definitions for constants into the classes themselves rather than having them outside.


Foundation/include/Poco/Timestamp.h
```diff
@@ -20,6 +20,7 @@
 
 #include "Poco/Foundation.h"
 #include <ctime>
+#include <limits>
```
matejk, line 23: `POCO_CHECK_MINMAX_MACROS` here, so a translation unit with a `min`/`max` macro in scope gets a message naming the cause instead of "too few arguments provided to function-like macro invocation". See the review summary for the definition. Same one-liner is needed in the five other headers this PR touches.

```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Foundation/include/Poco/Clock.h
```diff
@@ -19,6 +19,7 @@
 
 
 #include "Poco/Foundation.h"
+#include <limits>
```
matejk, line 22:
```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Foundation/include/Poco/Dynamic/VarIterator.h
```diff
@@ -21,6 +21,7 @@
 #include "Poco/Exception.h"
 #include <iterator>
 #include <algorithm>
+#include <limits>
```
matejk, line 24:
```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Net/include/Poco/Net/SocketProactor.h
```diff
@@ -37,6 +37,7 @@
 #include <memory>
 #include <iostream>
 #include <system_error>
+#include <limits>
```
matejk, line 40:
```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Data/include/Poco/Data/RecordSet.h
```diff
@@ -28,6 +28,7 @@
 #include "Poco/Exception.h"
 #include "Poco/AutoPtr.h"
 #include <ostream>
+#include <limits>
```
matejk, line 31:
```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Data/include/Poco/Data/RowIterator.h
```diff
@@ -23,6 +23,7 @@
 #include "Poco/Dynamic/Var.h"
 #include <iterator>
 #include <algorithm>
+#include <limits>
```
matejk, line 26:
```suggestion
#include <limits>


POCO_CHECK_MINMAX_MACROS
```

Foundation/src/Clock.cpp
```diff
@@ -29,7 +29,6 @@
 #include <algorithm>
 #undef min
 #undef max
```
matejk: `#include <limits>` moved to the header, so this guard is now orphaned -- it protects a line that is no longer here. Please drop both lines.

```suggestion
```

Foundation/src/Timestamp.cpp
```diff
@@ -18,7 +18,6 @@
 #include <algorithm>
 #undef min
 #undef max
```
matejk: Same as `Clock.cpp` -- orphaned guard, please drop.

```suggestion
```

Foundation/src/VarIterator.cpp
```diff
@@ -16,15 +16,11 @@
 #include "Poco/Dynamic/Var.h"
 #undef min
 #undef max
```
matejk: Same as `Clock.cpp` -- orphaned guard, please drop.

```suggestion
```

Data/src/RowIterator.cpp
```diff
@@ -16,15 +16,11 @@
 #include "Poco/Data/RecordSet.h"
 #undef min
 #undef max
```
matejk: Same as `Clock.cpp` -- orphaned guard, please drop.

```suggestion
```

Net/src/SocketProactor.cpp
```diff
@@ -22,7 +22,6 @@
 #undef max
 #endif // max
 #endif // POCO_OS_FAMILY_WINDOWS
```
matejk: Same here -- this guarded the `#include <limits>` that moved into the header. The whole block goes, including the two lines above that GitHub will not let me include in a suggestion: delete lines 20-24 (`#ifdef POCO_OS_FAMILY_WINDOWS` through `#endif // POCO_OS_FAMILY_WINDOWS`). No suggestion block on this one, since applying a partial deletion would leave the `#ifdef`s dangling.

Prometheus/include/Poco/Prometheus/MetricsServer.h
```diff
@@ -35,7 +35,7 @@ class Prometheus_API MetricsServer
 	/// requests, based on Poco::Net::HTTPServer.
 {
 public:
-	static const Poco::UInt16 DEFAULT_PORT; /// 9100
+	static constexpr Poco::UInt16 DEFAULT_PORT {9100};
```
matejk: Nit: brace init here, `=` everywhere else in the PR. Carried over from the old .cpp definition -- worth aligning.

```suggestion
	static constexpr Poco::UInt16 DEFAULT_PORT = 9100;
```

Net/include/Poco/Net/ICMPv4PacketImpl.h
```diff
@@ -157,13 +157,12 @@ class Net_API ICMPv4PacketImpl : public ICMPPacketImpl
 	virtual std::string typeDescription(int typeId);
 		/// Returns the description of the packet type.
 
-	static const Poco::UInt16 MAX_PACKET_SIZE;
 	static const std::string MESSAGE_TYPE[MESSAGE_TYPE_LENGTH];
```
matejk, line 160: Good catch removing the shadowing `MAX_PACKET_SIZE` declaration -- it was never defined anywhere, so it was a latent link error waiting to happen. The name now resolves to `ICMPPacketImpl::MAX_PACKET_SIZE`, same value (65535).

matejk (changes requested):
Nice cleanup -- the conversion itself is right, and I checked every moved value
against the definition it replaced (`DigestBuf::BUFFER_SIZE`,
`TaskManager::MIN_PROGRESS_NOTIFICATION_INTERVAL`, the Zip `HEADER[]` byte
sequences, `HelpFormatter::TAB_WIDTH`/`LINE_WIDTH`, `WinService::STARTUP_TIMEOUT`,
all `SSLManager::VAL_*`, `SQLExecutor::EPSILON_*`). They all match. Leaving the
`static const std::string` members alone was the correct call -- `constexpr
std::string` is C++20-only and cannot have static storage duration anyway.

Two things I especially liked:

- Dropping `ICMPv4PacketImpl::MAX_PACKET_SIZE`. That declaration shadowed
  `ICMPPacketImpl::MAX_PACKET_SIZE` and was never defined anywhere -- I confirmed
  it is absent from `libPocoNet`, so any odr-use of it was a latent link error.
  Now the name resolves to the inherited member, same value.
- Removing the `#if !defined(_MSC_VER) || defined(__STDC__)` blocks in
  `IPAddress.cpp` and `SocketAddress.cpp`. `constexpr` makes those members
  implicitly inline, so the workaround is genuinely obsolete.

### One thing to sort out before merge: `min`/`max` macro exposure

The PR moves `std::numeric_limits<T>::min()/max()` out of .cpp files and into
`Timestamp.h`, `Clock.h`, `Dynamic/VarIterator.h`, `Net/SocketProactor.h`,
`Data/RecordSet.h` and `Data/RowIterator.h`. If a consumer's translation unit
has a function-like `min` or `max` macro in scope, those headers no longer
compile:

```
error: too few arguments provided to function-like macro invocation
   static constexpr TimeVal TIMEVAL_MAX = std::numeric_limits<TimeVal>::max();
```

Windows is the usual source -- `<windows.h>` defines both unless `NOMINMAX` is
set -- but nothing restricts this to Windows; any header, on any platform, can
define such a macro, so the check below is deliberately platform-independent.

Our CI cannot catch the Windows case, because POCO's own .cpp files always
include a POCO header first, and `Poco/Foundation.h:106` ->
`Platform_WIN32.h:22` -> `Poco/UnWindows.h:29` defines `NOMINMAX` before pulling
`<windows.h>`. That is also why the change looked safe.

We do not want to work around this at every call site. The requirement is
reasonable -- POCO users have dealt with `NOMINMAX` for years -- but the failure
should say so instead of producing the message above.

#### Please add a `POCO_CHECK_MINMAX_MACROS` macro

One macro, placed once near the top of each header that uses
`numeric_limits<>::min()` / `max()`. Suggested definition, in
`Poco/Foundation.h` immediately after the `Platform.h` / `Platform_WIN32.h`
include block (around line 116):

```cpp
//
// Guard for headers that use std::numeric_limits<>::min()/max().
//
// A function-like min() or max() macro breaks those headers, and the compiler
// diagnostic does not say why. Most often the macro comes from <windows.h>
// included without NOMINMAX -- Poco/UnWindows.h sets it, but a consumer that
// reaches <windows.h> before any POCO header never goes through it. Any other
// header, on any platform, can define the same macros. Affected headers invoke
// POCO_CHECK_MINMAX_MACROS once, below their includes, to report the cause.
//
#if defined(min) || defined(max)
	#define POCO_CHECK_MINMAX_MACROS \
		static_assert(false, \
			"A min() or max() macro is defined, which breaks this header. " \
			"Undefine it; on Windows, define NOMINMAX or include a POCO " \
			"header before <windows.h>.");
#else
	#define POCO_CHECK_MINMAX_MACROS
#endif
```

Verified on clang++ and g++: with the macros defined the static assertion is the
first diagnostic emitted, and with them absent the macro expands to nothing and
the header compiles unchanged.

It has to live somewhere every affected header reaches, which rules out
`UnWindows.h` on its own -- that is only included from `Platform_WIN32.h`, so it
never runs on POSIX and the macro would be undefined there. `Foundation.h` is
the first common ancestor of all the headers involved.

Inline suggestions on each of the six place the invocation after the
`#include <limits>` you added.

### Cleanup to fold into this PR

With the check in place, the ad-hoc `#undef min` / `#undef max` guards scattered
around the tree can all go -- `Poco/UnWindows.h` is already the single entry
point for `<windows.h>` and guarantees `NOMINMAX`, so inside POCO they were
never doing anything. While you are in here, could you also:

1. **Drop the orphaned guards in the five .cpp files this PR already touches.**
   You moved `#include <limits>` into the headers but left the `#undef min` /
   `#undef max` that guarded it, so the guard now sits in the file that no
   longer needs it. Inline suggestions on these.
   - `Foundation/src/Clock.cpp:30-31`, `Foundation/src/Timestamp.cpp:19-20`,
     `Foundation/src/VarIterator.cpp:17-18`, `Data/src/RowIterator.cpp:17-18`,
     `Net/src/SocketProactor.cpp:21-24` (the whole `#ifdef POCO_OS_FAMILY_WINDOWS`
     block).

2. **Drop the same redundant guards elsewhere.** I checked each of these -- they
   all include a POCO header before the `#undef`, so `NOMINMAX` is already set:
   - `Crypto/src/CryptoStream.cpp:22-23`, `JSON/src/Parser.cpp:16-17`,
     `JSON/src/ParserImpl.cpp:18-19`, `PDF/src/Page.cpp:19-20`
   - `Util/testsuite/src/AbstractConfigurationTest.cpp:20-21`,
     `Util/testsuite/src/WinConfigurationTest.cpp:19-20`,
     `Util/testsuite/src/WinRegistryTest.cpp:17-18`,
     `Zip/testsuite/src/CompressTest.cpp:21`, `Zip/testsuite/src/ZipTest.cpp:29`
     (the last two are `#undef min` only)

   `PDF/src/Page.cpp` includes `<hpdf.h>` and the Zip tests pull in zlib, so
   those three are the ones to watch on the Windows CI run -- the others are
   POCO-only translation units.

3. **`Foundation/include/Poco/NumberParser.h:23-24`** -- this guard is simply
   vestigial: the header has no `min`, `max` or `numeric_limits` use at all.
   Drop the two `#undef` lines. I would leave the `<limits>` include on line 25
   alone even though it is unused, in case anyone depends on it transitively.

4. **Replace the remaining header guards with the check**, so the whole tree
   ends up on one mechanism. These three headers use
   `numeric_limits<>::min()` / `max()` and currently `#undef` their way around
   it -- swap each `#undef` block for a single `POCO_CHECK_MINMAX_MACROS`. The
   call sites themselves need no change.
   - `Foundation/include/Poco/Dynamic/VarHolder.h` -- guard on lines 45-46
     (10 affected call sites, all left as they are)
   - `Foundation/include/Poco/NumericString.h` -- guard on lines 24-29
     (8 call sites)
   - `Data/include/Poco/Data/Constants.h` -- `#undef max` on line 21 (1 call
     site). This header currently includes no POCO header at all, so it also
     needs `#include "Poco/Foundation.h"` for the macro to be available.

   Only `::min()` and `::max()` are macro-sensitive, so the many
   `::is_specialized`, `::digits`, `::lowest()` and `::epsilon()` uses in these
   files are irrelevant here.

5. **`OrderedSet.h:24-29` and `OrderedMap.h:24-29`** -- these guard the bundled
   `dependencies/tessil/include/Poco/ordered_hash.h`, which uses `std::min`,
   `std::max` and `std::numeric_limits<>::max()` in 15 places. Same treatment
   for consistency: replace the `#undef` blocks with `POCO_CHECK_MINMAX_MACROS`.
   Flagging it explicitly because it is the one spot where a configuration that
   works today starts failing -- the guard currently lets a `<windows.h>`-first
   TU use `OrderedSet`/`OrderedMap`. Happy to leave these two alone instead if
   you would rather not touch vendored-adjacent code.

6. **Route every direct `#include <windows.h>` through `Poco/UnWindows.h`** so
   there is exactly one way to reach it. 23 sites:
   - `Foundation/include/Poco/FileStreamRWLock_WIN32.h:24`,
     `Foundation/src/AsyncChannel.cpp:29`, `Net/src/SocketImpl.cpp:47`,
     `Net/src/SSPINTLMCredentials.cpp:24-25` (its own
     `#define WIN32_LEAN_AND_MEAN` becomes redundant and can go)
   - `NetSSL_Win/include/Poco/Net/{SecureSocketImpl,Context,AutoSecBufferDesc}.h`,
     `NetSSL_Win/src/Utility.cpp`
   - the 15 `Data/ODBC/include/Poco/Data/ODBC/*.h` headers, which need
     `<windows.h>` before `<sqlext.h>` -- `UnWindows.h` preserves that ordering

   These are all already no-ops at runtime, since `Poco/Foundation.h` reaches
   `<windows.h>` first in every case, so this is a readability change with no
   behavioural risk. It removes the last places where someone could copy the
   wrong pattern. NetSSL_Win and ODBC are both covered by Windows CI.

7. **The function-local constants**, to finish what the title promises. These
   are the last `const`s in the tree with literal initialisers that could be
   `constexpr`. Being function-local they change no codegen -- every compiler
   already folds them -- so the gain is intent plus a compile-time guarantee
   that the initialiser stays constant. Zero risk, and it means the sweep is
   actually complete:
   - `Foundation/include/Poco/Hash.h:143-144` -- `m` and `r` in `hashCombine()`
   - `Foundation/src/Thread_POSIX.cpp:223`, `Process_UNIX.cpp:247`,
     `ThreadPool.cpp:157`, `Logger.cpp:256`, `FastLogger.cpp:1767`,
     `RegularExpression.cpp:74`
   - `Foundation/src/Mutex_POSIX.cpp:136`, `Mutex_WIN32.cpp:38`,
     `Mutex_STD.cpp:43` and `:88` -- all `const int sleepMillis = 5;`
   - `Net/src/TCPReactorServerConnection.cpp:10` (namespace scope in the .cpp),
     `Net/src/NTLMCredentials.cpp:61`, `Net/src/NTPPacket.cpp:150`
   - `NetSSL_OpenSSL/src/SSLManager.cpp:273`, `Crypto/src/ECKeyImpl.cpp:431`
     and `:456`, `Data/SQLite/src/MemoryDB.cpp:755`,
     `Data/PostgreSQL/src/BinaryExtractor.cpp:125` and `:152`,
     `Data/DataTest/src/SQLExecutor.cpp:3087`

   One to leave alone: `Net/src/MultipartReader.cpp:363`,
   `const std::size_t minMatch = 3 + bLen;` -- `bLen` is `_boundary.size()`, so
   it is not a constant expression and `constexpr` there will not compile.

After all of that there should be no `#undef min` / `#undef max` left in the
tree, and every header that touches `numeric_limits<>::min()` / `max()` declares
the requirement in one line at the top. One mechanism, one place per file.

For what it is worth, I checked the in-class conversions for gaps and did not
find any -- of the 109 in-class `static const` initialisers in the tree, the
only two you left are `Data/ODBC/Diagnostics.h` and `Data/ODBC/Utility.h`, and
both are `std::string`, so they are correctly excluded.

### Please add the `breaking` label

This removes 37+ exported symbols -- Foundation 12, Net 17, plus Util, Data,
DataSQLite, MongoDB, XML, CppUnit, Zip, NetSSL and Prometheus -- verified with
`nm` against the shipped 1.15.3 libraries (`Timespan::SECONDS`,
`Timestamp::TIMEVAL_MAX`, `IPAddress::IPv4`, `HTTPMessage::UNKNOWN_CONTENT_LENGTH`
and so on). Source-compatible, binary-incompatible.

That is fine for 2.0, which breaks ABI anyway, but the PR needs the `breaking`
label so `gh-cli-for-release-notes.sh` picks it up, and an entry in the 2.0
Breaking Changes section of `CHANGELOG` and `doc/99100-ReleaseNotes.page`. The
requirement introduced by the check macro -- that no `min`/`max` macro is in
scope, which on Windows means `NOMINMAX` -- belongs in the same release-notes
entry, since it is a source-level change for consumers.

### One more small thing

Could you add a sentence to the PR description about why the TextEncoding
`_names[]` arrays and `Ascii::CHARACTER_PROPERTIES[128]` are out of scope? They
are the same shape as the Zip `HEADER[]` arrays you did convert, so the boundary
looks arbitrary without it. Leaving them alone is the right call.

PR #5456, opened 2026-08-19
chore: replace `for (;;)` with `while (true)`

mikomikotaishi:
This PR replaces `for (;;)`, which seems to be an archaism and vestige of 90s-era C (when compilers would skip evaluating `for (;;)`'s conditions but always evaluate `while (1)`'s condition), with `while (true)`, which is clearer in intent and reads much more smoothly in human language, and also compiles to the same thing.

### DIFF
--- a/Data/SQLParser/src/parser/bison_parser.cpp
+++ b/Data/SQLParser/src/parser/bison_parser.cpp
@@ -2318,7 +2318,7 @@ yytnamerr (char *yyres, const char *yystr)
     {
       YYPTRDIFF_T yyn = 0;
       char const *yyp = yystr;
-      for (;;)
+      while (true)
         switch (*++yyp)
           {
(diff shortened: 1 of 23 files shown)


Data/SQLParser/src/parser/bison_parser.cpp
```diff
@@ -2318,7 +2318,7 @@ yytnamerr (char *yyres, const char *yystr)
     {
       YYPTRDIFF_T yyn = 0;
       char const *yyp = yystr;
-      for (;;)
+      while (true)
```
matejk: This file is generated with `bison`. Every re-generation will create the `for` pattern again. 

No need to do changes in this file.

mikomikotaishi: Fixed

PR #5463, opened 2026-08-24
Mark all methods which necessarily throw, as `[[noreturn]]` (addresses #5152)

mikomikotaishi:
This PR adds the `[[noreturn]]` attribute to all methods which necessarily throw, or do not return.

One possible regression is that, because `Poco::Exception::rethrow()` is marked `[[noreturn]]`, any client code which overrides it to not throw and instead return, could potentially invoke undefined behaviour. However, I do not forsee this being a problem as code which does this would be violating the intended use of the method anyway.

Virtual overrides whose whole body is a throw (`WebSocketImpl::bind()`, `SecureStreamSocketImpl::listen()`, `DTDMap::setNamedItem()`, `Document::copyNode()` and etc.) are intentionally not annotated, because the attribute only aids direct calls on the derived type and none of those classes are `final`.

Continues to address #5152.

### DIFF
--- a/Foundation/include/Poco/Exception.h
+++ b/Foundation/include/Poco/Exception.h
@@ -81,7 +81,7 @@ public:
 		/// The copy can later be thrown again by
 		/// invoking rethrow() on it.
 
-	virtual void rethrow() const;
+	[[noreturn]] virtual void rethrow() const;
 		/// (Re)Throws the exception.
 		///
 		/// This is useful for temporarily storing a
@@ -161,6 +161,7 @@ inline int Exception::code() const
 		const char* className() const noexcept;										\
 		[[nodiscard]]                                                               \
 		Poco::Exception* clone() const;												\
+		[[noreturn]]                                                                \
 		void rethrow() const;														\
 	};
 
--- a/Foundation/testsuite/src/AsyncNotificationCenterTest.h
+++ b/Foundation/testsuite/src/AsyncNotificationCenterTest.h
@@ -95,7 +95,7 @@ protected:
 	void handleAuto(const Poco::AutoPtr<Poco::Notification>& pNf);
 	void handleCount(const Poco::AutoPtr<Poco::Notification>& pNf);
 	Poco::NotificationResult handleSync(const Poco::AutoPtr<Poco::Notification>& pNf);
-	void handleThrow(const Poco::AutoPtr<Poco::Notification>& pNf);
+	[[noreturn]] void handleThrow(const Poco::AutoPtr<Poco::Notification>& pNf);
 	void handleThreadSafe(const Poco::AutoPtr<Poco::Notification>& pNf);
 	bool matchAsync(const std::string& name) const;
 
--- a/Net/include/Poco/Net/DNS.h
+++ b/Net/include/Poco/Net/DNS.h
@@ -173,10 +173,10 @@ protected:
 	[[nodiscard]] static int lastError();
 		/// Returns the code of the last error.
 
-	static void error(int code, const std::string& arg);
+	[[noreturn]] static void error(int code, const std::string& arg);
 		/// Throws an exception according to the error code.
 
-	static void aierror(int code, const std::string& arg);
+	[[noreturn]] static void aierror(int code, const std::string& arg);
 		/// Throws an exception according to the getaddrinfo() error code.
 
 	[[nodiscard]] static std::string encodeIDNLabel(const std::string& idn);
--- a/CppParser/include/Poco/CppParser/CppToken.h
+++ b/CppParser/include/Poco/CppParser/CppToken.h
@@ -34,7 +34,7 @@ public:
 	~CppToken();
 
 protected:
-	void syntaxError(const std::string& expected, const std::string& actual);
+	[[noreturn]] void syntaxError(const std::string& expected, const std::string& actual);
 };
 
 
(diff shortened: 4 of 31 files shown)


mikomikotaishi:
cc @matejk

Net/include/Poco/Net/DNS.h
```diff
@@ -173,10 +173,10 @@ class Net_API DNS
 	[[nodiscard]] static int lastError();
 		/// Returns the code of the last error.
 
-	static void error(int code, const std::string& arg);
+	[[noreturn]] static void error(int code, const std::string& arg);
 		/// Throws an exception according to the error code.
 
-	static void aierror(int code, const std::string& arg);
+	[[noreturn]] static void aierror(int code, const std::string& arg);
```
matejk, line 179: `aierror()` is not noreturn in every configuration. In `Net/src/DNS.cpp` the whole body is inside `#if defined(POCO_HAVE_IPv6) || defined(POCO_HAVE_ADDRINFO)`. `POCO_HAVE_ADDRINFO` is not defined on VxWorks or when `POCO_NET_NO_ADDRINFO` is set (`SocketDefs.h`), so without IPv6 the function body is empty and returns. The call sites are guarded the same way, so nothing calls it in that configuration, but the definition still contradicts the attribute: GCC warns `'noreturn' function does return`, clang warns under `-Winvalid-noreturn`, and `-Werror` turns that into a build failure.

Suggested change in `Net/src/DNS.cpp`, so that every configuration ends in a throw:

```cpp
	default:
		throw DNSException("EAI", NumberFormatter::format(code));
	}
#else
	throw DNSException("EAI", NumberFormatter::format(code));
#endif // POCO_HAVE_IPv6 || defined(POCO_HAVE_ADDRINFO)
}
```

Nit in the same function: the `break;` after `error(lastError(), arg);` in `case EAI_SYSTEM:` is unreachable now that `error()` is `[[noreturn]]` and can be removed.

Foundation/include/Poco/Exception.h
```diff
@@ -81,7 +81,7 @@ class Foundation_API Exception: public std::exception
 		/// The copy can later be thrown again by
 		/// invoking rethrow() on it.
 
-	virtual void rethrow() const;
+	[[noreturn]] virtual void rethrow() const;
 		/// (Re)Throws the exception.
 		///
```
matejk, line 86: Since `rethrow()` is virtual, the attribute is now part of the contract for every override, including overrides outside this repository: an override that returns normally is undefined behaviour when called through `Poco::Exception&`. All in-tree overrides throw `*this`, so the change itself is fine, but the doc comment should state the requirement, and the 2.0 release notes should mention it.

```suggestion
		/// (Re)Throws the exception.
		///
		/// Declared [[noreturn]]: an overriding implementation
		/// must always throw and must never return normally.
		///
```

Foundation/testsuite/src/AsyncNotificationCenterTest.h
```diff
@@ -95,7 +95,7 @@ class AsyncNotificationCenterTest: public CppUnit::TestCase
 	void handleAuto(const Poco::AutoPtr<Poco::Notification>& pNf);
 	void handleCount(const Poco::AutoPtr<Poco::Notification>& pNf);
 	Poco::NotificationResult handleSync(const Poco::AutoPtr<Poco::Notification>& pNf);
-	void handleThrow(const Poco::AutoPtr<Poco::Notification>& pNf);
+	[[noreturn]] void handleThrow(const Poco::AutoPtr<Poco::Notification>& pNf);
```
matejk: Annotating test code is inconsistent here: the PR marks three test helpers (`handleThrow()` and the two `TestRollbackTransactor::operator()`), but the test suites contain about thirty more functions whose body is an unconditional throw, for example the stubs in `Data/ODBC/testsuite/src/ODBCTest.h` and the request handlers in `Net/testsuite/src/HTTPReactorServerTest.cpp`. The attribute has no practical effect in tests. I would drop the three test annotations and keep the PR to library code.

matejk (changes requested):
I checked every annotated function against its body in all platform variants and preprocessor configurations, and ran an independent scan (`clang -Wmissing-noreturn` over all library translation units, plus a read of the Windows-only and VxWorks-only sources that a compiler on Linux/macOS never sees). The set of annotated functions is correct except for `DNS::aierror()` (see the inline comment). The PR also correctly leaves alone the functions that look noreturn but are not (`SocketImpl::error()` returns on `POCO_ENOERR`, `Data::SQLite::Utility::throwException()` returns on `SQLITE_OK`/`SQLITE_ROW`/`SQLITE_DONE`) and the virtual base defaults that have returning overrides (`VarHolder::convert()`, `Channel::setProperty()`, `AbstractConfiguration::removeRaw()`, `AbstractExtractor::extract()` and similar).

Three Windows-only functions are missing; a compiler scan on Linux/macOS cannot see them:

- `Util/include/Poco/Util/WinRegistryKey.h`: `[[noreturn]] void handleSetError(const std::string& name);` -- the body is a single `throw SystemException(...)`.
- `NetSSL_Win/include/Poco/Net/SecureSocketImpl.h`: `[[noreturn]] void stateIllegal();` and `[[noreturn]] void stateError();` -- both end in an unconditional throw. The attribute does not change the member-function-pointer type used in the `_states` table.

One request for the PR description: please state that virtual overrides whose whole body is a throw (`WebSocketImpl::bind()`, `SecureStreamSocketImpl::listen()`, `DTDMap::setNamedItem()`, `Document::copyNode()` and similar) are intentionally not annotated. That is the right choice, because the attribute only helps direct calls on the derived type and none of those classes is `final`, but a reader of the diff will otherwise wonder whether they were missed.

Issue #4850, opened 2025-01-15
WebSocket: non-blocking receiveFrame()/receiveBytes() with TLS connection may get stuck receiving header 


obiltschnig:
The reason is that calling `receiveBytes()` with `MSG_PEEK` does not look behind the current TLS record. If the current record contains an incomplete WebSocket header, `peekHeader()` will not be able to ever return a full header.

The solution is to not use `MSG_PEEK`, and instead do a regular `receiveBytes()` and keep the bytes in an internal buffer.

PR #5473, opened 2026-09-07
5472 websocket loop forever

### LINKED ISSUE #5472: Net: WebSocket::receiveFrame() busy-loops forever when the peer closes after a partial frame header
### Version

`main` at a58826aee (also present in 1.15.x). Linux aarch64, clang 18.1.8.

### Symptom

A blocking `WebSocket::receiveFrame()` never returns and spins at 100% CPU when the peer sends part of a frame header and then closes the connection. A close with no partial header is handled correctly.

Measured with the reproducer below over 3 seconds of wall clock:

| header bytes sent before close | receiveFrame returns | process CPU burned |
|---|---|---|
| 0 | yes, returns 0 | 0.002 s |
| 1 | no | 3.00 s |
| 2 | no | 3.00 s |

### Cause

`WebSocketImpl::peekSomeBytes()` (Net/src/WebSocketImpl.cpp:481) returns the number of bytes already held in `_buffer` when the underlying `receiveBytes()` returns 0, so the EOF never becomes visible to its caller. `peekHeader()` keeps returning -1 because the header stays incomplete, and the retry loop in `WebSocketImpl::receiveBytes()` (Net/src/WebSocketImpl.cpp:296)

```cpp
int payloadLength = -1;
while (payloadLength < 0)
{
    payloadLength = peekHeader(_receiveState);
}
```

never terminates. After EOF nothing in that loop blocks, hence the spin. With an empty buffer the same EOF is reported as 0 and `receiveBytes()` returns cleanly, which is why only the partial header case hangs.

The internal buffer was introduced with #4850; the EOF case appears not to be covered.

### Impact

A peer can occupy a server thread permanently at 100% CPU by writing two bytes and closing.

### Suggested fix

Let the EOF reach the caller: `peekSomeBytes()` should report the end of the connection when the underlying `receiveBytes()` returned 0 and nothing new was buffered, and an incomplete header at EOF should be treated as a closed connection, returning 0 as the clean close path already does.

### Reproducer

Build: `c++ -std=c++17 -o wsrepro wsrepro.cpp -lPocoNet -lPocoFoundation`, run: `./wsrepro [header-bytes]` (default 2).

<details>
<summary>wsrepro.cpp</summary>

```cpp
#include "Poco/Net/HTTPServer.h"
#include "Poco/Net/HTTPRequestHandler.h"
#include "Poco/Net/HTTPRequestHandlerFactory.h"
#include "Poco/Net/HTTPServerRequest.h"
#include "Poco/Net/HTTPServerResponse.h"
#include "Poco/Net/HTTPServerParams.h"
#include "Poco/Net/HTTPClientSession.h"
#include "Poco/Net/HTTPRequest.h"
#include "Poco/Net/HTTPResponse.h"
#include "Poco/Net/HTTPMessage.h"
#include "Poco/Net/ServerSocket.h"
#include "Poco/Net/SocketAddress.h"
#include "Poco/Net/StreamSocket.h"
#include "Poco/Net/WebSocket.h"
#include "Poco/Thread.h"
#include "Poco/Event.h"
#include <atomic>
#include <cstdlib>
#include <iostream>
#include <sys/resource.h>
#include <unistd.h>

namespace
{
	std::atomic<bool> g_returned{false};
	Poco::Event g_reading;

	double cpuSeconds()
	{
		struct rusage ru;
		getrusage(RUSAGE_SELF, &ru);
		return ru.ru_utime.tv_sec + ru.ru_utime.tv_usec/1e6 + ru.ru_stime.tv_sec + ru.ru_stime.tv_usec/1e6;
	}

	class Handler: public Poco::Net::HTTPRequestHandler
	{
	public:
		void handleRequest(Poco::Net::HTTPServerRequest& request, Poco::Net::HTTPServerResponse& response)
		{
			Poco::Net::WebSocket ws(request, response);
			char buffer[1024];
			int flags = 0;
			g_reading.set();
			try
			{
				int n = ws.receiveFrame(buffer, sizeof(buffer), flags);
				std::cout << "receiveFrame returned " << n << ", flags " << flags << std::endl;
			}
			catch (Poco::Exception& exc)
			{
				std::cout << "receiveFrame threw: " << exc.displayText() << std::endl;
			}
			g_returned = true;
		}
	};

	class Factory: public Poco::Net::HTTPRequestHandlerFactory
	{
	public:
		Poco::Net::HTTPRequestHandler* createRequestHandler(const Poco::Net::HTTPServerRequest&)
		{
			return new Handler;
		}
	};
}

int main(int argc, char** argv)
{
	Poco::Net::ServerSocket svs(Poco::Net::SocketAddress("127.0.0.1", 0));
	Poco::UInt16 port = svs.address().port();
	Poco::Net::HTTPServer server(new Factory, svs, new Poco::Net::HTTPServerParams);
	server.start();

	// Handshake by hand so the frame can be sent in pieces.
	Poco::Net::HTTPClientSession session("127.0.0.1", port);
	Poco::Net::HTTPRequest request(Poco::Net::HTTPRequest::HTTP_GET, "/ws", Poco::Net::HTTPMessage::HTTP_1_1);
	request.set("Connection", "Upgrade");
	request.set("Upgrade", "websocket");
	request.set("Sec-WebSocket-Version", "13");
	request.set("Sec-WebSocket-Key", "dGhlIHNhbXBsZSBub25jZQ==");
	session.setKeepAlive(true);
	session.sendRequest(request);
	Poco::Net::HTTPResponse response;
	session.receiveResponse(response);
	if (response.getStatus() != Poco::Net::HTTPResponse::HTTP_SWITCHING_PROTOCOLS)
	{
		std::cout << "handshake failed: " << response.getReason() << std::endl;
		return 2;
	}
	Poco::Net::StreamSocket raw = session.detachSocket();

	g_reading.wait();

	// Part of a masked text frame: FIN|TEXT, MASK|length 4. The header needs
	// four more bytes (the mask) that never arrive.
	const char partialHeader[2] = {static_cast<char>(0x81), static_cast<char>(0x84)};
	int headerBytes = argc > 1 ? std::atoi(argv[1]) : 2;
	if (headerBytes > 0) raw.sendBytes(partialHeader, headerBytes);
	raw.close();

	double before = cpuSeconds();
	Poco::Thread::sleep(3000);
	double burned = cpuSeconds() - before;

	std::cout << "after peer close: receiveFrame returned = " << (g_returned ? "yes" : "no")
		<< ", process CPU burned in 3 s of wall clock = " << burned << " s" << std::endl;

	_exit(g_returned ? 0 : 1);
}
```

</details>

aleks-f:


Net/src/WebSocketImpl.cpp
```diff
@@ -178,6 +179,17 @@ int WebSocketImpl::sendBytes(const void* buffer, int length, int flags)
...
 
 
+int WebSocketImpl::incompleteHeader(ReceiveState& receiveState)
+{
+	receiveState.frameFlags = 0;
+	// The bytes seen so far do not make a complete header. Once the peer has
+	// closed the connection the rest of that header can never arrive, so the
+	// close is reported instead of asking the caller to peek again, which it
+	// would otherwise do forever.
+	return _peerClosed ? 0 : -1;
```
matejk, line 195: This makes a connection closed mid-header indistinguishable from a clean close, while a connection closed mid-payload still throws `WS_ERR_INCOMPLETE_FRAME`. It also contradicts the `receiveFrame()` doc comment in `WebSocket.h` (lines 279-284), which says that an incomplete frame throws and that 0 with flags 0 means the peer shut down or closed the connection.

Returning 0 is defensible: RFC 6455 section 7.1.5 treats any transport close without a Close frame as abnormal closure 1006, whether or not a partial frame preceded it. I would keep the 0 and update the `receiveFrame()` doc comment to say that a close in the middle of a header is reported as a close.

Net/src/WebSocketImpl.cpp
```diff
@@ -178,6 +179,17 @@ int WebSocketImpl::sendBytes(const void* buffer, int length, int flags)
 }
 
 
+int WebSocketImpl::incompleteHeader(ReceiveState& receiveState)
+{
+	receiveState.frameFlags = 0;
```
matejk: Nit: `peekHeader()` already resets `frameFlags` at entry (line 197), so this is redundant. Harmless; keeping it makes the helper self-contained.

Net/src/WebSocketImpl.cpp
```diff
@@ -507,6 +520,10 @@ int WebSocketImpl::peekSomeBytes(char* buffer, int length)
 			std::memcpy(_buffer.begin(), buffer, rc);
 			_bufferOffset = 0;
 		}
+		else if (rc == 0)
+		{
+			_peerClosed = true;
```
matejk, line 536: `receiveSomeBytes()` calls the same underlying `receiveBytes()` (line 485) and ignores a 0 return. No live bug, since every path that reaches it throws `WS_ERR_INCOMPLETE_FRAME` on a short read, but `_peerClosed` is now a class invariant that only one of the two readers maintains. Setting it there as well keeps a future reordering from missing the close.

Net/src/WebSocketImpl.cpp
```diff
@@ -468,7 +477,7 @@ int WebSocketImpl::receiveSomeBytes(char* buffer, int length)
 	{
 		if (length < n) n = length;
 		std::memcpy(buffer, _buffer.begin() + _bufferOffset, n);
-		_bufferOffset += length;
+		_bufferOffset += n;
```
matejk: Verified: the old over-advance was reached on every frame whose payload exceeded the leftover peeked bytes, and `receiveSomeBytes()`, `peekSomeBytes()` and `available()` all treat a negative remainder the same as zero, so the no-change-in-behaviour claim holds.

Net/include/Poco/Net/WebSocketImpl.h
```diff
@@ -147,6 +152,9 @@ class Net_API WebSocketImpl: public StreamSocketImpl
 	Poco::Buffer<char> _buffer;
 	int _bufferOffset;
 	bool _mustMaskPayload;
+	bool _peerClosed;
```
matejk: Informational: this adds a data member to an exported class, so `sizeof(WebSocketImpl)` changes. Fine for a release, since the SONAME is bumped every time, but worth a line in the backport note: a distro cherry-pick onto an already shipped 1.15.x would affect code that derives from `WebSocketImpl`.

Net/testsuite/src/WebSocketTest.cpp
```diff
@@ -287,6 +357,55 @@ void WebSocketTest::testWebSocketLargeInOneFrame()
 }
 
 
+void WebSocketTest::testPeerCloseAfterPartialHeader()
```
matejk, line 384: The commit message says the fix covers both blocking modes and both `receiveBytes()` overloads, but only blocking `receiveBytes(void*, int, int)` is exercised here. I checked non-blocking by hand (`setBlocking(false)` in the handler plus a retry loop): with the fix, `receiveFrame()` returns 0 at once; without it, the loop spins at 100% CPU and `received` stays -1. A `bool blocking` parameter on this test would lock that in.

Net/testsuite/src/WebSocketTest.cpp
```diff
@@ -287,6 +357,55 @@ void WebSocketTest::testWebSocketLargeInOneFrame()
...
+	assertTrue (response.getStatus() == HTTPResponse::HTTP_SWITCHING_PROTOCOLS);
+	Poco::Net::StreamSocket ws = cs.detachSocket();
+	assertTrue (state.upgraded.tryWait(10000));
+
+	// The first two bytes of a masked text frame carrying four bytes of
+	// payload. The four mask bytes that would complete its header never
+	// arrive, because the peer closes instead.
+	const char partialHeader[2] = {'\x81', '\x84'};
+	ws.sendBytes(partialHeader, sizeof(partialHeader));
+	ws.close();
```
matejk: `ws.shutdownSend()` would guarantee a FIN. `close()` sends an RST instead when the socket still has unread data; the server's `recv()` then fails with `ECONNRESET`, the handler catches it, and `state.received` stays -1. There is no unread data at this point, so this is robustness only; the test did not flake in 10 local runs.

Net/testsuite/src/WebSocketTest.cpp
```diff
@@ -287,6 +357,55 @@ void WebSocketTest::testWebSocketLargeInOneFrame()
...
+
+	// The peer is gone, so the header can never be completed: the read must
+	// report the closed connection rather than wait for the rest of it.
+	bool returned = state.done.tryWait(10000);
+	if (!returned)
+	{
+		// Release the reader, so that neither this server's shutdown nor
+		// the rest of the suite waits for a thread that never returns.
+		Poco::FastMutex::ScopedLock lock(state.mutex);
+		if (state.pWebSocket) state.pWebSocket->close();
```
matejk: Failure path only, so not blocking: this closes the socket from another thread while the handler may be inside `recv()` on it, which TSan reports as a race if it ever runs. Also, `~TCPServer()` does not join in-flight connection threads (`TCPServerDispatcher::stop()` only enqueues stop notifications), so if the second `tryWait()` below expires as well, `state` is destroyed while the handler thread may still write to it. In practice the close does release the reader: with the fix reverted, this branch ended the test after the 10 s wait.

matejk (approved):
Approved. I built `Net-testrunner` (Debug, macOS arm64) at the head and ran `WebSocketTest` in four configurations:

| Configuration | Result | CPU |
|---|---|---|
| PR head, blocking (as written), 10 runs | OK every run | ~0.02 s |
| Implementation reverted to a58826aee, test kept, blocking | FAILURE `"returned"` | 10.0 s in 12.7 s wall |
| PR head, test patched to non-blocking | OK, `receiveFrame()` returns 0 | 0.03 s |
| Implementation reverted, non-blocking | FAILURE `"state.received == 0"` | 5.0 s in 7.7 s wall |

So the test reproduces the defect, the fix resolves it in both modes, and CI is green.

`_peerClosed` can only be set by a real close: `SocketImpl::receiveBytes()` returns 0 only for `recv() == 0` (non-blocking starvation is -1), and `SecureSocketImpl::handleError()` returns 0 only for `SSL_ERROR_ZERO_RETURN` and the server-side unexpected EOF; `ERR_SSL_WOULD_BLOCK`, `ERR_SSL_WANT_READ` and `ERR_SSL_WANT_WRITE` are all negative.

The inline comments cover the documented contract for the 0 return (the one thing I would change before merging), the `receiveSomeBytes()` asymmetry, and the test.

Not inline, since the lines are outside the diff:

- The `while (payloadLength < 0) payloadLength = peekHeader(...)` loops in both blocking `receiveBytes()` overloads still have no progress check and no yield. The known trigger is gone; the only other way to a -1 there that I could find is the OpenSSL 3 `SSL_ERROR_SSL` fallthrough in `handleError()` with a stale errno, so this is hardening, not a bug. Given that this loop is why the PR exists, a bail-out when `peekSomeBytes()` returns without having grown the buffer would be cheap insurance.

Process:

- Milestone and project are not set, the title does not follow the `fix(Net): ...` convention (the first commit's subject fits), and the body is empty.
- This is a pre-authentication remote DoS: two bytes and a disconnect pin a server thread at 100% CPU permanently, repeatable until the pool is exhausted. It should carry the `security` label and go into 1.15.4.
- Affected range is 1.14.0 through 1.15.3. The swallowed-EOF mechanism came with #4850 (1.15.0), as the issue says. 1.14.x spins as well, through `recv(MSG_PEEK)` returning the pending bytes forever; I confirmed that by running the reproducer from #5472 against 1.14.2 (1 or 2 header bytes: no return, 3.00 s CPU in 3 s of wall clock). This fix does not apply to that code, and 1.14 is not maintained, so it only matters for the affected-versions line of an advisory.

PR #5486, opened 2026-09-20
Add Windows to installing dependencies heading

mikomikotaishi:
This heading only mentions Linux and macOS, but there are instructions for Windows as well; it should be added.

Or, alternatively, the (Linux, macOS, Windows) could just be removed since these are essentially the major platforms of interest.


mikomikotaishi:
cc @matejk

PR #5488, opened 2026-09-21
fix(Crypto): check the results of the OpenSSL calls in DigestEngine

matejk:
`Poco::Crypto::DigestEngine` ignored the results of `EVP_DigestInit_ex()`, `EVP_DigestUpdate()` and
`EVP_DigestFinal_ex()`. With OpenSSL 3.0 and newer, `EVP_get_digestbyname()` succeeds for an algorithm that no
loaded provider offers under the current property query, and the failure only shows up in
`EVP_DigestInit_ex()`. This happens, for example, for MD5 when the default property query is `fips=yes`, or for
a digest of the legacy provider when that provider is not available. `digest()` then returned a zero-filled
buffer of the digest length instead of reporting an error, and the OpenSSL error stayed on the error queue of
the thread.

### Changes

- Every OpenSSL call in `DigestEngine` is checked. A failure throws `Poco::Crypto::OpenSSLException`; the
  message contains the algorithm name and the reasons reported by OpenSSL, and the error queue is left empty.
  An unknown algorithm name still throws `Poco::NotFoundException`.
- The context is created and initialized in one step (`newContext()`). `reset()` replaces the old context only
  after the new one was initialized, so a failing `reset()` leaves the engine unchanged. The constructor no
  longer leaks the `EVP_MD_CTX` when it throws.
- `digestLength()` and `digest()` reject a non-positive size from `EVP_MD_CTX_size()` (SHAKE digests return -1,
  which was converted to a 4 GiB `resize()`).
- No change to the class layout; the header change is documentation only.
- `RSADigestEngine`, `ECDSADigestEngine` and the JWT HMAC signers use this class and now get the exception
  instead of computing over a zero digest.

### Tests

`DigestEngineTest`: `testUnknownAlgorithm`; with OpenSSL 3.0+ `testInitFailure` (default property query
`provider=nonexistent` makes every fetch fail; checks the exception, the empty error queue and that an existing
engine stays usable after a failed `reset()`) and `testFIPSRefusesMD5` (runs when a FIPS provider is active,
prints a note otherwise).

Tested with OpenSSL 1.1.1w, 3.5.5, 3.6.4 and 4.0.2 (AddressSanitizer, without and with an active FIPS
provider): Crypto and JWT test suites pass.

The same change will be proposed for `main`.


matejk:
Already in poco-1.15.4 as 58e1b6eac (same change, the branch was rebuilt).

